PR #1, opened 2015-01-20
rewrote monomer construction to pull directly from Statements

jmuhlich:
@johnbachman let's review this.


jmuhlich:
Accepting as is, but Ben and I will hack on it some more to address the issues we identified in review.

PR #8, opened 2015-02-09
Models, RAS demo

bgyori:
Added a models folder and a RAS demo model. Also added BEL/RDF files for the demo in the data folder.


johnbachman:
Looks like the files data/RAS_combined.bel/rdf are missing.

bgyori:
I added the combined rdf and cleaned up the model folder.

PR #11, opened 2015-06-05
Restructuring BelPy around PySB Assembler

bgyori:


johnbachman:
Merged these directly into sorgerlab/master after rebasing to remove redundant commits.

PR #12, opened 2015-06-08
TripsProcessor development and database clients

bgyori:
Basic functionalities implemented in TripsProcessor, and basic versions of clients for accessing information from Uniprot and HGNC


johnbachman:
Looks great. We might want to start writing some tests, especially to cover basic use of the database clients (something like what you're doing in the __main__ sections to demonstrate basic usage). At some point we should also set up a travis build process to make sure we're all settled on the same set of dependencies.

PR #22, opened 2016-02-01
Assembler refactoring

johnbachman:
In this pull request we refactored pysb_assembler and statements.py so that the PySB assembly specific code would be decoupled from the Statement classes. The Statement classes now only serve as data structures to record the arguments associated with each interaction type; policies and procedures for assembly will now be implemented in specific assembler code.

Rather than create a registry-type data structure in pysb_assembler to define the monomers and assembly function for specific statements and policies, we used a naming convention in which assembly functions followed the pattern classname_assemble_policyname. If a function with the name corresponding to a specific statement/policy does not exist, then an exception is raised.

In addition, this pull request also implements a slightly more sophisticated approach to specification of policies: policies can now be specific globally using a string (e.g., 'interactions_only') or on a statement-specific basis using a dict (e.g. {'Phosphorylation': 'two_step', 'Dephosphorylation': 'interactions_only'}).

### DIFF
--- a/indra/pysb_assembler.py
+++ b/indra/pysb_assembler.py
@@ -49,6 +49,7 @@ class BaseAgentSet(object):
     def __getitem__(self, name):
         return self.agents[name]
 
+
 class BaseAgent(object):
     def __init__(self, name):
         self.name = name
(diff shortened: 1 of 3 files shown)


bgyori:
Looks good, merged.

johnbachman:
Thanks--any insight into why the tests are failing?

bgyori:
Yes, it is a problem with pyjnius on Travis and it is unrelated to the pull request. All the relevant tests are passing.

PR #48, opened 2016-03-29
Modifications refactoring

bgyori:
This pull request changes the way modifications are represented in INDRA. The modifications in ActivityModification statements and the modification conditions of Agents are now a list of ModCondition objects. A ModCondition object has a mod_type (e.g. 'phosphorylation'), a residue (e.g. 'serine'), a position (e.g. '222') and an is_modified flag (True/False) to be able to represent the lack of a modification. 

Modification and SelfModification statements have also been refactored. They now have a residue (e.g. 'serine') and a position (e.g. '222') associated with them. This replaces the old 'PhosphorylationSerine' and similar modification strings. 

All processors and assemblers as well as the Statements and Preassembler of INDRA are affected.

### DIFF
--- a/indra/bel/processor.py
+++ b/indra/bel/processor.py
@@ -168,20 +168,25 @@ class BelProcessor(object):
 
             if act_type == 'Kinase' and mod.startswith('Phosphorylation'):
                 self.statements.append(
-                        Phosphorylation(enz, sub, mc, evidence))
+                        Phosphorylation(enz, sub, mc.residue, mc.position,
+                                        evidence))
             elif act_type == 'Catalytic':
                 if mod == 'Hydroxylation':
                     self.statements.append(
-                            Hydroxylation(enz, sub, mc, evidence))
+                            Hydroxylation(enz, sub, mc.residue, mc.position,
+                                          evidence))
                 elif mod == 'Sumoylation':
                     self.statements.append(
-                            Sumoylation(enz, sub, mc, evidence))
+                            Sumoylation(enz, sub, mc.residue, mc.position,
+                                        evidence))
                 elif mod == 'Acetylation':
                     self.statements.append(
-                            Acetylation(enz, sub, mc, evidence))
+                            Acetylation(enz, sub, mc.residue, mc.position,
+                                        evidence))
                 elif mod == 'Ubiquitination':
                     self.statements.append(
-                            Ubiquitination(enz, sub, mc, evidence))
+                            Ubiquitination(enz, sub, mc.residue, mc.position,
+                                           evidence))
                 else:
                     print "Warning: Unknown modification type!"
                     print("Activity: %s, Mod: %s, Mod_Pos: %s" %
--- a/indra/bel/processor.py
+++ b/indra/bel/processor.py
@@ -160,32 +160,32 @@ class BelProcessor(object):
             sub_name = gene_name_from_uri(stmt[2])
             sub = Agent(sub_name)
             mod = term_from_uri(stmt[3])
+            residue = self._get_residue(mod)
             mod_pos = term_from_uri(stmt[4])
-            mc = self._get_mod_condition(mod, mod_pos)
             stmt_str = strip_statement(stmt[5])
             # Mark this as a converted statement
             self.converted_stmts.append(stmt_str)
 
             if act_type == 'Kinase' and mod.startswith('Phosphorylation'):
                 self.statements.append(
-                        Phosphorylation(enz, sub, mc.residue, mc.position,
+                        Phosphorylation(enz, sub, residue, mod_pos,
                                         evidence))
             elif act_type == 'Catalytic':
                 if mod == 'Hydroxylation':
                     self.statements.append(
-                            Hydroxylation(enz, sub, mc.residue, mc.position,
+                            Hydroxylation(enz, sub, residue, mod_pos,
                                           evidence))
                 elif mod == 'Sumoylation':
                     self.statements.append(
-                            Sumoylation(enz, sub, mc.residue, mc.position,
+                            Sumoylation(enz, sub, residue, mod_pos,
                                         evidence))
                 elif mod == 'Acetylation':
                     self.statements.append(
-                            Acetylation(enz, sub, mc.residue, mc.position,
+                            Acetylation(enz, sub, residue, mod_pos,
                                         evidence))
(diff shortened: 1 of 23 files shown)


indra/bel/processor.py
```diff
@@ -191,6 +196,20 @@ def get_modifications(self):
...
+            if mod == 'Phosphorylation':
+                mc.residue = None
+            else:
+                mc.residue = mod[15:].lower()
+            mc.position = mod_pos
+        else:
+            mc = ModCondition(mod)
+            mc.position = mod_pos
+        return mc
+
```
johnbachman: Perhaps we could separate the lookup of residue/position from the construction of  ModCondition here (since it's not a condition after all).

indra/biopax/processor.py
```diff
@@ -197,25 +185,26 @@ def _get_complex_members(cplx):
...
+                    # Skip activity as a modification state for now
+                    continue
+                mods.append(mc)
+        return mods
 
     def _get_generic_modification(self, mod_filter=None, mod_gain=True, 
                                   force_contains=None):
+        '''
+        Get all modification reactions given a filter
+        '''
```
johnbachman: Make docstring format consistent

indra/biopax/processor.py
```diff
@@ -334,6 +319,9 @@ def _get_evidence(bpe):
 
     @staticmethod
     def _construct_modification_pattern():
+        '''
+        Constructs the BioPAX pattern to extract modification reactions
+        '''
         pb = bpp('PatternBox')
```
johnbachman: Docstring format

indra/statements.py
```diff
@@ -3,24 +3,61 @@
...
+    def matches(self, other, mod_hierarchy):
+        type_match = (self.mod_type == other.mod_type or \
+            mod_hierarchy.isa(self.mod_type, other.mod_type))
+        residue_match = (self.residue == other.residue or \
+            (self.residue is not None and other.residue is None))
+        pos_match = (self.position == other.position or \
+            (self.position is not None and other.position is None))
+        return (type_match and residue_match and pos_match)
+
+    def matches_key(self):
```
johnbachman: Matches key should return a string.

indra/statements.py
```diff
@@ -3,24 +3,61 @@
...
+        if not isinstance(position, basestring):
+            if position is None:
+                self.position = None
+            else:
+                self.position = str(position)
+        else:
+            self.position = position
+        self.is_modified = is_modified
+
+    def matches(self, other, mod_hierarchy):
```
johnbachman: Matches should call matches_key for the equality test.

johnbachman: The current implementation of matches_key should be redefined as refinement_of.

indra/statements.py
johnbachman: Matches_key should return a string.

indra/statements.py
johnbachman: Matches should be redefined as refinement_of.

indra/statements.py
```diff
@@ -36,8 +73,7 @@ def matches(self, other):
 
     def matches_key(self):
         key = (self.name,
-               set(self.mods),
-               set(self.mod_sites),
+               set([m.matches_key() for m in self.mods]),
```
johnbachman: Maybe add a note that this may break if you get an agent with two phosphorylations at serine with unknown sites (as we see for MEK, for example).

indra/statements.py
```diff
@@ -95,23 +131,13 @@ def refinement_of(self, other, entity_hierarchy, mod_hierarchy):
         # MODIFICATIONS
         # Similar to the above, we check that self has all of the modifications
         # of other.
-        assert len(self.mods) == len(self.mod_sites), \
-               "Mods and mod_sites must match."
-        assert len(other.mods) == len(other.mod_sites), \
-               "Mods and mod_sites must match."
         # Make sure they have the same modifications
         for other_mod_ix in range(len(other.mods)):
```
johnbachman: We don't need indexing for these anymore because we don't have two parallel lists.

indra/statements.py
```diff
@@ -95,23 +131,13 @@ def refinement_of(self, other, entity_hierarchy, mod_hierarchy):
...
-            other_mod_site = other.mod_sites[other_mod_ix]
             for self_mod_ix in range(len(self.mods)):
                 self_mod = self.mods[self_mod_ix]
-                self_mod_site = self.mod_sites[self_mod_ix]
-                # Or has an isa relationship...
-                if (self_mod == other_mod or \
-                        mod_hierarchy.isa(self_mod, other_mod)) and \
-                   (self_mod_site == other_mod_site or \
-                        (self_mod_site is not None and other_mod_site is None)):
+                if self_mod.matches(other_mod, mod_hierarchy):
```
johnbachman: This should be a call to a ModCondition refinement_of.

indra/statements.py
```diff
@@ -437,44 +467,21 @@ class ActivityModification(Statement):
...
-            # any entries in the mod_pos list that might be ints
-            self.mod = mod
-            self.mod_pos = [str(mp) for mp in mod_pos]
-        elif isinstance(mod, basestring) and \
-                (isinstance(mod_pos, int) or isinstance(mod_pos, basestring)):
-            self.mod = [mod]
-            self.mod_pos = [str(mod_pos)]
-        elif isinstance(mod, basestring) and mod_pos is None:
+        # Turn mod conditions into a list of a single one is given
+        if isinstance(mod, ModCondition):
```
johnbachman: Maybe file an issue for ActivityModifications where the modcondition is None, implying that the agent is constitutively active. This could be useful for indicating that mutants are active while the WT proteins require modifications to be active.

indra/statements.py
```diff
@@ -506,15 +507,9 @@ def refinement_of(self, other, entity_hierarchy, mod_hierarchy):
         for other_mod_ix in range(len(other.mod)):
             mod_found = False
             other_mod = other.mod[other_mod_ix]
-            other_mod_pos = other.mod_pos[other_mod_ix]
             for self_mod_ix in range(len(self.mod)):
```
johnbachman: Get rid of indexing here.

bgyori: Fixes in statements.py in https://github.com/sorgerlab/indra/pull/48/commits/68a7babb258755a90b6bd03bc339fd8f509e7264

PR #57, opened 2016-04-21
Module reorganize

bgyori:
This pull request reorganizes INDRA modules in two important ways:
## 1. Assemblers are now in indra.assembler
Asemblers are now in their own folder under indra. All assembler classes (PysbAssembler, CxAssembler, etc.) are imported by indra/assembler/\__init\__.py so from now on the preferred way of importing an assembler is:
`from indra.assembler import PysbAssembler`
which replaces the old
`from indra.pysb_assembler import PysbAssembler`
indra.assembler.\__init__.py catches ImportErrors so one can still use PysbAssembler without and error, just because, for instance pygraphviz is missing, which is a dependency of GraphAssembler. 

## 2. Input APIs are exposed by their respective modules
For instance indra/bel/\__init\__py now imports the API functions from indra/bel/bel_api.py. Therefore from now on the preferred way to use input APIs is:
```
from indra import bel
bel.process_belrdf(rdf_str)
```
or
```
from indra import trips
trips.process_text('MEK phosphorylates ERK.')
```
The old way will still work, so 
```
from indra.trips import trips_api
trips_api.process_text('MEK phosphorylates ERK.')
```
is still valid but less intuitive. 

------
Any thoughts?


bgyori:
The test failure is unrelated, it's because of a doctest in preassembler.

johnbachman:
Shouldn't it be indra.assemblers? (plural)

Also, I thought we had a deal! No commits!!! :)

bgyori:
Unable to resist the urge...

bgyori:
I renamed assembler to **assemblers** and dded one more thing:
## 3. Literature module is reorganized 
So that you can now instead of having to 
```
from indra.literature import literature_client
literature_client.get_full_text('PMID123')
```
you can use it as
```
import indra.literature
indra.literature.get_full_text('PMID123')
```

johnbachman:
:+1:

PR #61, opened 2016-05-03
Activity conditions

bgyori:
This pull request replaces ActivityModification and ActivatingSubstitution by a more general and simpler ActiveForm statement. The ActiveForm statement declares that a given Agent with arbitrary ModConditions, MutConditions and BoundConditions is active or inactive with respect to a given activity type. This allows us to express more general and more complex activity conditions than what was previously possible.

### DIFF
--- a/indra/assemblers/pysb_assembler.py
+++ b/indra/assemblers/pysb_assembler.py
@@ -16,7 +16,7 @@ SelfExporter.do_export = False
 # in this list then we require that there is at least one default
 # policy implemented to assemble that type of statement.
 statement_whitelist = [ist.Modification, ist.SelfModification, ist.Complex,
-                       ist.ActivityActivity, ist.ActivityModification,
+                       ist.ActivityActivity, ist.ActiveForm,
                        ist.RasGef, ist.RasGap]
 
 def is_whitelisted(stmt):
@@ -86,7 +86,8 @@ class BaseAgent(object):
         # The list of site/state configurations that lead to this agent
         # being active (where the agent is currently assumed to have only
         # one type of activity)
-        self.activating_mods = []
+        self.active_forms = []
+        self.inactive_forms = []
         self.db_refs = {}
 
     def create_site(self, site, states=None):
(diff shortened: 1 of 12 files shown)


indra/assemblers/pysb_assembler.py
```diff
@@ -107,8 +108,11 @@ def add_site_states(self, site, states):
             if state not in self.site_states[site]:
                 self.site_states[site].append(state)
 
-    def add_activating_modification(self, activity_pattern):
-        self.activating_mods.append(activity_pattern)
+    def add_active_form(self, activity_pattern, is_active):
```
jmuhlich: This method name is confusing as it can also be used to add _inactive_ forms via `is_active=False`. Perhaps `add_activity_form` or something like that?

jmuhlich: Or just a separate `add_inactive_form` method.

indra/assemblers/pysb_assembler.py
```diff
@@ -154,10 +158,10 @@ def add_activating_modification(self, activity_pattern):
 # ActivityActivity rules, where the modification site mediating the activation
 # is not specified.
 default_mod_site_names = {
-    'Kinase': 'phospho',
-    'GtpBound': 'RBD',
-    'Phosphatase': 'phospho',
-    'Activity': 'act'
+    'kinase': 'phospho',
+    'gtpbound': 'RBD',
```
jmuhlich: For consistency should "rbd" be lower case too?

indra/assemblers/pysb_assembler.py
```diff
@@ -177,11 +181,11 @@ def get_mod_site_name(mod_type, residue, position):
     name = ('%s%s' % (mod_str, mod_pos))
     return name
 
-def get_activating_mods(agent, agent_set):
-    act_mods = agent_set[agent.name].activating_mods
-    if not act_mods:
-        act_mods = [{}]
-    return act_mods
+def get_active_forms(agent, agent_set):
```
jmuhlich: Is the intent to eventually implement `get_inactive_forms` too? If so, might be good to implement it as a stub that just throws `NotImplementedError`.

indra/assemblers/pysb_assembler.py
```diff
@@ -245,10 +249,8 @@ def get_create_parameter(model, name, value, unique=True):
     return parameter
 
 
-def get_complex_pattern(model, agent, agent_set, extra_fields=None):
+def get_complex_pattern(model, agent, extra_fields=None):
```
jmuhlich: A bit nitpicky, but doesn't this actually return a MonomerPattern?

jmuhlich:
That's it from me, looks good. (except for Travis failing :)

bgyori:
Thanks, Jeremy, I think those were all valid comments and I made updates accordingly.

PR #104, opened 2016-09-25
Reimplementation of the Preassembler

bgyori:
This is a complete, conceptual reimplementation of building the Statement hierarchy (combine_related) in the Preassembler.

Some results on RAS Machine subsets: 
**10k Statements** from the RAS Machine now preassemble in **10 seconds**. 
**20k Statements** from the RAS Machine now preassemble in **1 minute**. 
**45k Statements** from the RAS Machine now preassemble in **3 minutes**. 
And no memory issues either :)

First, we start with an extension of the HierarchyManager. Namely, this pull request extends the transitive closure construction with a connected-component labeling algorithm. This allows one to look up an entity such as "ERK" (namespace not shown here) and in O(1) time get the ID of the connected component that it is in within the entity hierarchy, for instance, 123. Logically, "MAPK1" and "MAPK" would be in the same connected component and would therefore also be associated with the same component ID: 123. 

The Preassembler first groups Statements by their type. Then for each type of Statement it constructs groups of Statements that are potentially related based on their Agents' component IDs.   The trick of course is that this grouping is done in O(n) time since we just need to do O(1) lookups for each Statement. Within each compatible component ID group, it does pairwise comparisons to find the supports-supported_by relationships.

All relevant tests are passing but we should inspect real use cases carefully to make sure everything still works.

### DIFF
--- a/indra/preassembler/__init__.py
+++ b/indra/preassembler/__init__.py
@@ -142,15 +142,16 @@ class Preassembler(object):
         The procedure for combining statements in this way involves a series
         of steps:
 
-        1. The statements are grouped by whether they are of the same type
-           (e.g., Phosphorylation) and involve the same entities (e.g.,
-           BRAF and MAP2K1).
-        2. The groups of statements are then further compared to see if one
-           group involves superfamily entities of another group. If so, the
-           statements involving the superfamily are added into the group of
-           statements involving the more concrete entities to create an
-           "extended group."
-        3. The statements within each extended group are then compared; if one
+        1. The statements are grouped by type (e.g., Phosphorylation) and
+           each type is iterated over independently.
+        2. Statements of the same type are then grouped according to their
+           Agents' entity hierarchy component identifiers. For instance,
+           ERK, MAPK1 and MAPK3 are all in the same connected component in the
+           entity hierarchy and therefore all Statements of the same type
+           referencing these entities will be grouped. This grouping assures
+           that relations are only possible within Statement groups and
+           not among groups.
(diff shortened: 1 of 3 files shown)


bgyori:
Travis errors are related to an unresponsive NDEx web service.

bgyori:
I added a few more simple optimizations and got another 6x speedup...

20k Statements from the RAS Machine now preassemble in 10 seconds. 
45k Statements from the RAS Machine now preassemble in 35 seconds.

johnbachman:
Looks great! One thought I had is that given the efficiency of this approach, we should think carefully about adding highly generic families like "Kinase" and "Protein" into the hierarchy. Generics at that level might be useful for expressing certain types of prior knowledge, but they will produce a large connected component that will lead to a large number of comparisons within the component. But nevertheless it should always be faster than the previous O(n^2) approach.

bgyori:
Yes, you're right - this algorithm scales k^2 with k being the size of the largest connected component. So it works well only if you don't have giant connected components in your entity hierarchy.
There are some trivial optimizations we can still do:
- look at the number of members in complexes to skip some cases
- look at the position of an entity argument (e.g. ezyme or substrate)  to filter out some cases

By the way, are all the large scale applications still working? I hope nothing serious is broken.

johnbachman:
Regarding the applications/tools, I've added tests for the complexes/bioprocesses benchmarks and will add additional tests for the other applications (models, tools) as part of the Python 3 refactoring. So far so good after merging this PR.

PR #141, opened 2016-11-30
Bioentities map

bgyori:
This pull request integrates Bioentities mapping into the BEL, REACH and TRIPS processors. It also completely changes the way grounding is extracted and prioritized in the TRIPS processor - a much needed improvement. New resources are added which allow NCIT entries to be mapped to BE, HGNC, CHEBI and GO. The mapping to BE is enabled partly by REACH, which now includes BE in grounding, and partly by new resource files establishing mappings to BE from NCIT, XFAM and NextProt.


bgyori:
Looks like errors are all due to API keys.

PR #143, opened 2016-12-13
Model checker and updates to PySB assembler

johnbachman:
This pull request supports the ability to check statements about influences or mechanisms (in the form of INDRA statements) against executable models (in the form of PySB rule-based models). The functionality is implemented in a new tool, the ModelChecker (in indra/tools/model_checker.py) that uses the Kappa influence map to determine if there is a *possible* causal path linking a subject (e.g., an enzyme or perturbation) to a downstream object or target. Due to the non-transitivity of causality in the influence map (due to unmentioned context that doesn't carry over from the output of one rule to the input of another) the existence of a path does not guarantee that the model satisfies the statement. However, the failure of the model to satisfy the statement definitively indicates that the model cannot reproduce the statement if it were to be simulated. This therefore because a way to diagnose gaps and limitations in the scope of the model.

In the current implementation, the model checker returns only a boolean response indicating whether a statement is or is not satisfied; it doesn't return the path itself. For efficiency purposes, the breadth-first search algorithm (implemented in ModelChecker._find_sources) keeps track of path polarities and lengths but not the exact paths. This could be updated in the future.

Mapping the elements of the statement against the model requires that the model be annotated with grounding information. This is implemented via the "global" list of Annotations associated with the model, as well as Monomer-specific annotations added (monkeypatched) into the monomer field the site_annotations field, which consists of a list of Annotation objects. Site annotations currently used include the following:

* Site states representing modifications: ```Annotation(('T185', 'p'), 'phosphorylation', 'is_modification')```
* Residues: ```Annotation('T185', 'T', 'is_residue')```
* Residue positions: ```Annotation('T185', '185', 'is_position')```

The grounded_monomer_pattern method in the pysb_assembler is used to return a monomer pattern matching the elements of a statement while taking grounding and site annotations into account. Because there may be more than one monomer pattern associated with a "generic" state of an Agent, this function is a generator, returning matching monomer patterns with each call.

This pull request also adds support for all Modification types in the PySB assembler by creating generic modification assembly functions; and adds several new Modification types related to lipid modifications of proteins.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: python
 python:
   - "2.7"
   - "3.5"
+addons:
+    apt:
+        packages:
+            - ocaml-nox
+            - opam
 before_install:
   - sudo add-apt-repository --yes ppa:webupd8team/java
   - sudo apt-get update
@@ -29,6 +34,17 @@ install:
   - tar xzf BioNetGen-2.2.6-stable.tar.gz
   - export BNGPATH=`pwd`/BioNetGen-2.2.6-stable
   - pip install git+https://github.com/pysb/pysb.git
+  # Kappa
+  # First install ocamlfind via opam (needed to build KaSim/KaSa)
+  - opam init -a git://github.com/ocaml/opam-repository && eval $(opam config env)
+  - opam install ocamlfind --yes
+  # Install KaSim/KaSa
+  - git clone https://github.com/Kappa-Dev/KaSim.git
+  - cd KaSim
+  - git checkout f87eada
+  - make all
+  - export KAPPAPATH=`pwd`
+  - cd ../
   # Biopax/Paxtools dependencies
   - pip install jnius-indra
   # Download a number of useful resource files for testing purposes
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ python:
   - "3.5"
 addons:
     apt:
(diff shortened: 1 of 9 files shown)


jmuhlich:
Are we diverging from Walter's concept that a model is ungrounded? I thought the grounding would remain at the KR level, i.e. in the PysbAssembler and not the generated model. Wouldn't it be just as easy to keep the site annotations (and even the model annotations) in the assembler object instead? The site annotations lists could be kept in a dict keyed on monomer name.

bgyori:
My view is that the PySB model is still not the final executable model and it's fine (and very useful) to have annotations on it. I think "ungrounded" starts when you generate Kappa or ODEs which is still the case.

johnbachman:
Indeed--in fact, this is precisely why we added the Annotation class to PySB back in the day. An advantage with keeping the annotations in the PySB model is that the Annotations can be linked to actual model objects, not just to string names, etc.

jmuhlich:
Well the site annotations are just linked to string names! I started 
thinking about adding a richer representation of sites to pysb to 
support site annotations better, then realized we might want to pull 
back and keep the grounding in the KR. The pysb model is definitely more 
model-like than KR-like.

bgyori:
I've tested and everything seems to work fine within Bob/MRA.

PR #166, opened 2017-02-03
Increase efficiency of preassembly combine_related by more restrictive statement grouping

johnbachman:
This pull request changes the way statements are grouped during the combine_related assembly process. Previously, statements having corresponding agents at the same position were grouped together (where agent correspondence was determined by having the same hierarchy component ID or the same entity_matches_key). This made it straightforward to group together statements where one or more of the arguments were None (since they would be grouped together by the not-None arguments at other positions), but had the cost of including many statements that could never be refinements of each other because of non-correspondence at other positions. For example, the statements Phosphorylation(MEK, ERK) and Phosphorylation(SRC, ERK) would be in the same group due to the correspondence of the ERK argument; even though the non-correspondence of MEK/SRC implies that neither statement could be a refinement of the other.

In the new implementation, statements are grouped together if they have corresponding agents at *each* of their positions, and None statements are handled separately. The procedure goes in two phases. In the first phase, statements are grouped by a tuple of keys, where the keys consist of a component ID or an entity_matches_key. This way, Phosphorylation(MEK1, ERK1) and Phosphorylation(MEK2, ERK2) would end up in the same group, but Phosphorylation(SRC, ERK1) would not be included. After assembling the list of keys for a statement, the statement is added in to the corresponding entry in the dict, *and* the key is added to two index dicts: one which keeps tracks of keys by their first argument and the other which keeps track of keys by their second argument. Statements having None as the first or second argument are tracked separately in two dicts, none_first and none_second, which are indexed by the not-None argument.

In the second phase, the statements with None arguments that have been collected are iterated over and added to *all* of the statement groups where they could potentially be found to have a refinement relationship. For example, the statement Phosphorylation(None, ERK) would need to be added to the list of statements including [Phosphorylation(MEK1, ERK), Phosphorylation(MEK2, ERK)] *and* the list of Statements including Phosphorylation(SRC, ERK). This is done by pulling out the not-None argument from the none_first/second list, using it to look up the keys for the statement groups having a corresponding argument *at that position* using the stmt_by_first and stmt_by_second dicts, then iterating over those groups and adding the None-argument statements to each of those lists.

This approach dramatically reduces the size of the largest component group during preassembly, with a corresponding increase in performance. Note that these comparisons are to the previous preassembly approach, but after the bugfix in ddc99d2b9eaf, which made it perhaps 10-20% slower than it used to be.

Here are some stats for randomly selected subsets of the Phase3 statement set (917k total) of different sizes. Note that these are for the time taken to run the assemble_corpus.run_preassembly_related function, which includes not only the time taken to run combine_related, but also to propagate belief scores in the BeliefEngine. So the speedup in the time taken just for combine_related is greater than what is shown here.

```
50,000 stmts:
    - 37,726 unique statements
    - Before: 41.2 sec
        35,993 top-level statements (No MP)
        Total comparisons: 3,660,417
        Max group size: 816
        (18% of all comparisons)
    - After: 14.2 sec
        Total comparisons: 799613
        Max group size: 56
        Largest group: (('128', '14'), [Phosphorylation(MAP2K1(), ERK())]
        (0.4 % of all comparisons)
        INFO: indra/assemble_corpus - 35993 statements after filter...

200,000 stmts:
    - 124,631 unique statements
    - Before: 332.6 sec
        118,128 top-level statements
        Total comparisons: 33,043,477
        Max group size: 2108
        (13.4 pct of all comparisons)
    - After: 114.3 sec
        118128 top-level statements
        Total comparisons: 11,221,034
        Max group size: 136
        Largest group: (('131', '131'), [Phosphorylation(ERK1_2(), ERK1_2(), S)
        (0.2 % of all comparisons)

500,000 stmts:
    - Before: 1188.4 sec 
        Total comparisons: 124,151,960
        Max group size: 3626
        (10.6 pct of all comparisons)
        INFO: indra/assemble_corpus - Filtering 258935 statements for top-level
        INFO: indra/assemble_corpus - 243929 top-level statements
    - After: 521.8 sec
        Total comparisons: 66,764,637
        Max group size: 256
        Largest group: (('104', '104'), [Phosphorylation(ERK1_2(mods: (phosphorylation)), ERK1_2())
        (0.1 % of all comparisons)
        INFO: indra/assemble_corpus - Filtering 258935 statements for top-level
        INFO: indra/assemble_corpus - 243929 top-level statements

### DIFF
--- a/indra/preassembler/__init__.py
+++ b/indra/preassembler/__init__.py
@@ -231,6 +231,8 @@ class Preassembler(object):
             stmts_by_type[type(stmt)].append(stmt)
 
         group_sizes = []
+        largest_group = None
+        largest_group_size = 0
         num_stmts = len(unique_stmts)
         related_stmts = []
         # Each Statement type can be preassembled independently
@@ -248,7 +250,7 @@ class Preassembler(object):
                 entities = []
                 for i, a in enumerate(stmt.agent_list()):
                     # Entity is None
-                    if a is None:
+                    if a is None and stmt_type != Complex:
                         entities.append(a)
                         continue
                     # Entity is not None, but could be ungrounded or not
@@ -270,19 +272,10 @@ class Preassembler(object):
                             entities.append(a.entity_matches_key())
                         # Component ID, so this is in a family
                         else:
-                            # For Complexes we cannot optimize by argument
-                            # position because all permutations need to be
-                            # considered but we can use the number of members
-                            # to statify into groups
-                            if stmt_type == Complex:
-                                key = (len(stmt.members), component)
-                            # For all other statements, we separate groups by
-                            # the argument position of the Agent
-                            else:
-                                entities.append(component)
-                            # Don't add the same Statement (same object) twice
-                            #if stmt not in stmt_by_group[key]:
-                            #    stmt_by_group[key].append(stmt)
+                            # We turn the component ID into a string so that
+                            # we can sort it along with entity_matches_keys
+                            # for Complexes
+                            entities.append(str(component))
                 # At this point we have an entity list
                 # If we're dealing with Complexes, sort the entities and use
                 # as dict key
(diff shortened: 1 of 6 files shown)


johnbachman:
FYI the tests now pass when I run them.

PR #171, opened 2017-02-13
PysbPreassembler

bgyori:
This pull request changes the way the PysbAssembler works in two important ways. 

Until now, each time an Agent was in an position of activity (Modification.enz, RegulateActivity.subj, etc.) it was replaced in the assembly function itself, locally, by its active forms irrespective of whether its activity was explicitly asserted. This is a functionality which is useful for large-scale assembly from unstructured sources. But I believe it should not be implemented inside the PysbAssembler, especially not inside each assembly function, and should not be applied by default. This pull request removes this logic which will later have to be reimplemented in the context of the MechanismLinker.

Further, in cases when an Agent is asserted to be active, it makes sense to substitute this active state with the Agent's explicitly given ActiveForms. Consider for instance: "BRAF that is phosphorylated and not bound to Vemurafenib is active. Active BRAF phosphorylates MEK" which is a clear case in which "Active BRAF" should be replaced by "BRAF that is phosphorylated and not bound to Vemurafenib". 
The application of these rules is best done globally, before statement-wise assembly, and can be achieved by rewriting the original INDRA Statements to generate a new set of Statements. This is done by a class called PysbPreassembler. The PysbAssembler internally calls the PysbPreassembler to update its original Statement set and then calls the statement-wise assembly functions on the newly obtained Statements. This is the major addition of this pull request.


johnbachman:
Superseded by #183 .

PR #184, opened 2017-03-03
Parallelization of Preassembler.combine_related using multiprocessing

johnbachman:
This pull request parallelizes the combine_related function using Python’s built-in multiprocessing library. Once statements are partitioned into subsets the comparisons of statements within the subsets can be handled by worker processes in a parallel fashion. The key issue is that Python doesn’t support shared memory between processes, instead relying on communication between processes based on serialization and deserialization. As a result, any parallelization comes with an upfront cost based on the amount of communication required between processes.  The greatest efficiency gains therefore come from remotely distributing the work of comparisons within large groups (having large numbers of potentially related statements), while performing comparisons among the much more numerous small groups in the parent process. This is because for a group of k statements, the performance benefit of performing k^2 comparisons in a worker process outweighs the communication cost of serializing and deserializing the k statements when k is large. Empirical tests showed that the greatest performance benefits came from sending groups of more than 30-100 statements to worker processes while performing comparisons on all remaining smaller groups in the parent process. This makes it possible to preassemble the family-expanded Phase 3 evaluation set of over 1 million statements (which had previously failed to assemble due to excessive memory consumption) within 40 minutes on a 32 virtual CPU Amazon EC2 instance.

To reduce memory overhead, the implementation relies on the "spawn" mode of creating worker processes, which is only available on Python 3.4 and later. If a user attempts to run multiprocessing for combine_related in Python 2, a logging info message is produced and all comparisons are run in the parent process.


bgyori:
Looks good!

PR #252, opened 2017-07-06
Orthology and isoform mapping for site mappering, paper 2 sitemapper figs

johnbachman:
This pull request aims to automate many of the heuristics used to curate sites for the SiteMapper. Sites of post-translational modifications that are inconsistent with the reference sequence but are nevertheless meaningful (i.e., that should be mapped to the reference sequence in order to integrate information) seem to result from the following sources of error:

1. Site position drawn from a non-canonical protein isoform (i.e., not the isoform used for the reference sequence)
2. Site position drawn from the orthologous mouse protein
3. Site position off by one from the reference sequence (most likely due to cleavage of the initial methionine, a common protein modification)

This PR automates the process of validating and mapping inconsistent site positions due to these sources of error. To prevent the occurrence of false positive mappings, it only maps from, and to, sites that are listed in the PhosphoSitePlus database as known to be modified.

For example, we frequently encounter references in both databases and reading to phosphorylation on serine 112 of the protein BAD. However, there is no corresponding serine at position 112 of the human reference sequence. To identify a mapping to the human reference residue, the SiteMapper now does the following:

- Identifies the Uniprot ID for the mouse protein orthologous to human BAD. This uses the HGNC (human genome) to MGI (mouse genome) mapping table incorporated in 62e00039de53a007.
- Searches the PhosphoSitePlus dataset for any known post-translational modification sites on mouse BAD. (This requires that the Phosphosite site dataset be requested from Cell Signaling Technology and downloaded into the indra/resources folder.)
- If there is a serine at position 112 of the mouse sequence that is annotated as modified in PhosphoSitePlus, it is linked to the corresponding human site (serine 75).
- A mapping is created, identified as an inferred site from the mouse sequence, and a new, mapped INDRA Statement is generated.

The mapping procedure checks for matches to an inconsistent site in the following order:

1. another human isoform
2. a mouse protein
3. an off-by-one position (cleaved methionine)

If no match is found, then the manually curated site map table is checked. If the PhosphoSite dataset cannot be loaded, then all of these procedures are skipped and only the site map table is checked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ before_install:
   - sudo add-apt-repository --yes ppa:webupd8team/java
   - sudo apt-get update
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
-      wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh --nv;
+      wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh -nv;
     else
-      wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh --nv;
+      wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh -nv;
     fi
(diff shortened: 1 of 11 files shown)


johnbachman:
All tests passing.

PR #263, opened 2017-07-27
Improve docs

cthoyt:
Fixed a typo where a single item in parentheses is not turned to a tuple without a trailing comma. Replaced with brackets as a list instead to keep it pretty.


bgyori:
Thanks, this is a good suggestion! The case where you pass a single BoundCondition is handled in https://github.com/sorgerlab/indra/blob/master/indra/statements.py#L513-L514 so it's not really a bug but I also prefer lists. There are a few other places I can change this in the docstrings so I'll add a commit on top of yours and then merge.

cthoyt:
Great! I love overloading arguments in python (take that, static typed languages!)

PR #264, opened 2017-07-27
Master

cthoyt:
Starting to build test cases for PyBEL export. Not ready to merge for a while :)

### DIFF
--- a/indra/tests/test_statements_serialization.py
+++ b/indra/tests/test_statements_serialization.py
@@ -1,6 +1,9 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
+from pybel.constants import *
+
 from indra.statements import *
+from indra.util.export_pybel import convert_statement
 
 ev = Evidence(source_api='bel', pmid='12345', epistemics={'direct': True},
               text='This is the evidence.')
@@ -28,6 +31,20 @@ def test_modification():
     g = stmt.to_graph()
     jd2 = Statement._from_json(jd).to_json()
     assert (jd == jd2)
+
+
+def test_pybel_modification():
+    stmt = Phosphorylation(Agent('a'), Agent('b'), 'S', evidence=[ev])
+
+    u, v, d = convert_statement(stmt)
+
+    source = PROTEIN, 'HGNC', 'a'
+    target = PROTEIN, 'HGNC', 'b'
+    target_phosphorylated = PROTEIN, 'HGNC', 'b', (PMOD, (BEL_DEFAULT_NAMESPACE, 'Ph'), 'Ser')
+    target_reaction = REACTION, (target,), (target_phosphorylated,)
+
+    assert source == u
+    assert target_reaction == v
 
 
 def test_selfmodification():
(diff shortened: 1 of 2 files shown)


bgyori:
Hey, @johnbachman has actually implemented a lot of a PyBEL assembler already so consult with him first on what's left to do.

bgyori:
The other issue is that I think the way you started is not quite the standard structure in terms of file placement, etc. We should discuss before you proceed.

cthoyt:
I talked to him last weekend in Prague and it looks like we should all work on this together :)

Let's plan on having a skype call sometime next week.

johnbachman:
Hi, yes--I just pushed my pybel_assembler branch to sorgerlab so we can work off of the same feature branch. Will submit as separate pull request so we can discuss there.

indra/tests/test_statements_serialization.py
```diff
@@ -1,88 +1,99 @@
 from __future__ import absolute_import, print_function, unicode_literals
-from builtins import dict, str
```
johnbachman: I tend to prefer to keep these as boilerplate imports because this way if the code is ever extended to call str() in Python 2 (e.g., to convert an int to a string) it will return a future.newstr, which is unicode compatible, rather than a Python 2 bytestring.

johnbachman:
With respect to the organization of the code--we put different "export" modules in the assemblers submodule and call them "xxx_assembler.py", with an a class "XxxxAssembler" that takes INDRA Statements as a constructor; has an add_statements() method for adding more statements; and a "make_model()" method for generating the output format.

indra/tests/test_statements_serialization.py
cthoyt: aha! good call. I thought you were over python 2 thought 😉

johnbachman: I am, but we continue to drag it along behind us like an anchor :) I look forward to the day when we can jettison all of the extra code I wrote to keep everything 2/3 compatible.

cthoyt: I can't remember if I asked why you weren't using the `six` module to take care of these things automatically. Was there a problem with unicode/bytes?

johnbachman: The issue was that we wanted to enforce a pure Unicode sandwich where everything inside the application was unicode, with bytes only at the edges. Didn't look much into six--would probably have helped make the code cleaner but ultimately wouldn't have been sufficient by itself. See this blog post I put together on a 2/3 compatible unicode sandwich: http://johnbachman.net/building-a-python-23-compatible-unicode-sandwich.html#building-a-python-23-compatible-unicode-sandwich

bgyori:
Hey, @cthoyt, I'll close this PR and we can continue work on #265. Another thing I wanted to add is that submitting a PR from your master is not ideal since now it's out of sync with sorgerlab/master and you'll have to do a hard reset to get it back on track. So please use feature branches off of master for PRs - thanks!

cthoyt:
Good plan! John's is much further along. I've never done a PR in a project that had a lot going on, and I'm not much of a git ninja, so this is valuable advice!

PR #265, opened 2017-07-28
INDRA to PyBEL Assembler

johnbachman:
This pull request is aimed at creating an assembler from INDRA Statements into PyBEL graphs. PyBEL itself can then be used to create BEL Script files or JGF (JSON Graph Format) files for use with other repositories and analysis tools.


indra/assemblers/pybel_assembler.py
```diff
@@ -0,0 +1,155 @@
...
+    hgnc_name = hgnc_client.get_hgnc_name(hgnc_id)
+    node = (pc.PROTEIN, 'HGNC', hgnc_name)
+    node_attr = {pc.FUNCTION: pc.PROTEIN, pc.NAMESPACE: 'HGNC',
+                 pc.NAME: hgnc_name}
+    return (node, node_attr)
+
+def _get_citation(evidence):
+    pass
+
+"""
```
johnbachman: @cthoyt As you can see here, I was trying to reverse-engineer the structure of the PyBEL graph and associated metadata--had included some notes in a comment below.

indra/assemblers/pybel_assembler.py
```diff
@@ -0,0 +1,155 @@
...
+
+        self.model = None
+
+
+    def make_model(self, **kwargs):
+        self.model = pybel.BELGraph(**kwargs)
+        for stmt in self.statements:
+            # Convert phosphorylation statements into
+            # kin(p(agent)) => pmod(p(agent))
+            if isinstance(stmt, Phosphorylation):
```
johnbachman: @cthoyt This block deals only with Phosphorylations--the idea would be to specify a canonical approach to representing each of the INDRA Statements as BEL/PyBEL, preserving all evidence text, context information, etc.

indra/tests/test_pybel_assembler.py
```diff
@@ -0,0 +1,33 @@
...
+def test_simple_phosphorylation_with_evidence():
+    braf = Agent('BRAF', db_refs={'HGNC': '1097', 'UP': 'P15056'})
+    mek = Agent('MAP2K1', db_refs={'HGNC': '6840', 'UP': 'Q02750'})
+    ev1 = Evidence('reach', '1234', '12345678', 'Some example text.')
+    ev2 = Evidence('reach', '2345', '23456789', 'Some more example text.')
+    stmt = Phosphorylation(braf, mek, 'S', '218', evidence=[ev1, ev2])
+    pba = pa.PybelAssembler([stmt])
+    belgraph = pba.make_model()
+    pybel.to_bel(belgraph)
+    assert len(belgraph.nodes()) == 2
```
cthoyt: PyBEL stores a different node for each variant of the original agent - there should also be another node for `(pc.PROTEIN, 'HGNC', 'MAPK1', (pc.PMOD, (pc.BEL_DEFAULT_NAMESPACE, 'Ph')))`.

We also have to consider if we want to store this information in BEL as `act(p(HGNC:BRAF)) => p(HGNC:MAP2K1, pmod(Ph, 'Ser', 218)` or `act(p(HGNC:BRAF)) => rxn(reactants(p(HGNC:MAP2K1)), products(p(HGNC:MAP2K1, pmod(Ph, 'Ser', 218))`. I personally prefer the latter but for different problems it might better to represent it with the former.

In the latter case, we also have to add a node for `(pc.REACTION, ((pc.PROTEIN, 'HGNC', 'MAPK1'),), ((pc.PROTEIN, 'HGNC', 'MAPK1',  (pc.PMOD, (pc.BEL_DEFAULT_NAMESPACE, 'Ph'), )), ))` however it looks like we're getting to an edge case in re-usability in the PyBEL data structure, since the node tuples are a bit redundant of the associated data dictionaries. We'll have to investigate further

indra/assemblers/pybel_assembler.py
cthoyt: Great start! I can either provide you examples on what to put where, or you could dive into reading the [unit tests](https://github.com/pybel/pybel/blob/develop/tests/test_parse_bel.py) in PyBEL. They're all structured the same so it would be easy to look up what tuple/data dictionary go with which BEL nodes.

indra/assemblers/pybel_assembler.py
```diff
@@ -0,0 +1,155 @@
...
+                # Define the enzyme node
+                (enz_node, enz_attr) = _get_agent_node(stmt.enz)
+                (sub_node, sub_attr) = _get_agent_node(stmt.sub)
+                sub_attr[pc.VARIANTS] = [
+                                    {pc.KIND: pc.PMOD,
+                                     pc.IDENTIFIER: {
+                                       pc.NAMESPACE: pc.BEL_DEFAULT_NAMESPACE,
+                                       pc.NAME: 'Ph'},
+                                     PmodParser.CODE: stmt.residue,
```
cthoyt: Btw, are you using a *really* old version of PyBEL? I think I moved all of the constants out of the parser classes for easier reuse a long time ago.

johnbachman:
OK, so some key todos at this point:

- Assemble ActiveForm statements
- Assemble Translocation statements
- Assemble Gef/Gap statements
- Assemble Complex statements
- Assemble SelfModification statements (if possible)
- Determine best practices for encoding evidence
- Determine best practices for adding/encoding BELGraph metadata

INDRA functions to interface with PyBEL tools (e.g., upload/download networks from PyBEL web) are probably best put in a separate feature/PR for e.g., a databases/pybel_client.py module.

Anything I'm missing for this feature to be considered complete?

cthoyt:
I've added the GEF/GAP activities to the language dictionary (see http://pybel.readthedocs.io/en/latest/constants.html#pybel.parser.language.activity_labels) to immediately enable you to write activity statements the same as before with the `BEL_DEFAULT_NAMESPACE` tag

Agreed about splitting the PR. Like NDEx, I'd like to develop a nice client that also handles authentication the same way as NDEx. This would be a good use case to guide that development.

With regards to metadata:

I've added new keyword arguments to BELGraph (https://github.com/pybel/pybel/commit/fdfb2a87f301bcc02aa909c4921a6d37233f4268) and additional getters and setters (https://github.com/pybel/pybel/commit/cd1d7652a1125e02f66fed5f8262190d2e3fb58d) so we can make the INDRA code more beautiful and idiomatic. I will release to PyPI soon after a couple more bug fixes with citation caching (unrelated).

PyBEL also enforces the constraint that the groundings used in the networks are backed by a BEL namespace file. I've already created a small function for looking them up by keyword: http://pybel-tools.readthedocs.io/en/latest/resourceutils.html#pybel_tools.resources.get_latest_arty_namespace

The list of all modules available are at: https://arty.scai.fraunhofer.de/artifactory/bel/namespace/

johnbachman:
[indra to pybel integration](https://trello.com/c/vZlGnDL9/5-indra-to-pybel-integration)

cthoyt:
I'm cthoyt on trello

johnbachman:
Hi @cthoyt, what's the best way to represent complexes in PyBEL, e.g. ```complex(p(HGNC:CCND1), p(HGNC:CDK4))```?

Similarly what are the best way to represent reactions, with and without an enzymatic controller?

johnbachman:
Note to self: we're going to need to handle BoundConditions when constructing Agents, which presumably take the form of Complex nodes on the BEL/PyBEL side.

Gef/Gap and ActiveForms seem to be working. RasGtpActivation is a little funky because it is handled by `_assemble_regulate_activity` already through inheritance but this method doesn't add in the `gtpbound` activity state to the subject that we probably want here.

So at this point this leaves:
* Assemble Translocation statements
* Assemble Complex statements
* Assemble SelfModification statements. I think we can just encode these as, e.g. `kin(p(EGFR)) => pmod(p(EGFR, etc.))`, which won't distinguish between *cis* and *trans* but that's OK from a static causal analysis perspective.
* Assemble Conversion statements
* Refactoring out common copy-pasted elements of various assembly funcs
* Updates to evidence and grounding schemes as necessary.
* Function docs

cthoyt:
Here are some specific examples:

``complex(p(HGNC:CCND1), p(HGNC:CDK4))`` can be represented as
```python
my_complex_data = {
  pc.FUNCTION: pc.COMPLEX,
  pc.MEMBERS: [
    {
      pc.FUNCTION: pc.PROTEIN,
      pc.NAMESPACE: 'HGNC',
      pc.NAME: 'CCND1'      
    }, {
      pc.FUNCTION: pc.PROTEIN,
      pc.NAMESPACE: 'HGNC',
      pc.NAME: 'CDK4'   
    }
}
```

Note the ordering is in canonical form here, but `BELGraph.add_node_from_data` will fix it for you in case it isn't. All you have to do now is:

```python
import pybel
graph = pybel.BELGraph(name='test', version='0.0.0')
my_complex_tuple = graph.add_node_from_data(my_complex_data)
```

Similarly, a reaction can be represented as:
```
my_reaction_data = {
    FUNCTION: REACTION
    REACTANTS: [
        {
            FUNCTION: ABUNDANCE,
            NAMESPACE: 'CHEBI',
            NAME: '(3S)-3-hydroxy-3-methylglutaryl-CoA'
        }, {
            FUNCTION: ABUNDANCE,
            NAMESPACE: 'CHEBI',
            NAME: 'NADPH'
        }, {
            FUNCTION: ABUNDANCE,
            NAMESPACE: 'CHEBI',
            NAME: 'hydron'
        }
    ],
    PRODUCTS: [
        {
            FUNCTION: ABUNDANCE,
            NAMESPACE: 'CHEBI',
            NAME: 'mevalonate'
        }, {
            FUNCTION: ABUNDANCE,
            NAMESPACE: 'CHEBI',
            NAME: 'NADP(+)'
        }
    ]
}
``` 
then similarly added to a graph with
```python
my_reaction_tuple = graph.add_node_from_data(my_reaction_data)
```

When it comes to enzymes, I would represent it in BEL as `kin(p(...)) => rxn(reactants(...), products(...))`. This would require no additional work after implementing reactions and could be done with something like this:

```python
graph.add_edge(
  my_complex_tuple,
  my_reaction_tuple,
  attr_dict={
     pc.RELATION: pc.INCREASES,
     pc.SUBJECT: {
         pc.MODIFIER: pc.ACTIVITY,
         EFFECT: {
            NAME: 'kin'
            NAMESPACE: BEL_DEFAULT_NAMESPACE
        }
    },
    pc.CITATION: { ... }
    pc.EVIDENCE: '...',
    pc.ANNOTATIONS: { ... }
  }
)
```

Disregard the fact that this doesn't make biological sense, but at least it outlines the right data structure :)

indra/assemblers/pybel_assembler.py
```diff
@@ -161,6 +163,19 @@ def _assemble_active_form(self, stmt):
         for edge_data in edge_data_list:
             self.model.add_edge(subj_node, obj_node, attr_dict=edge_data)
 
+    def _assemble_complex(self, stmt):
+        members_list = []
+        for member in stmt.agent_list():
+            func, namespace, name = _get_agent_grounding(member)
+            members_list.append({
```
cthoyt: I would make this recursive for getting the JSON of any given member

johnbachman: So I was wondering about that--can members of a complex be anything? e.g., can they be modified proteins? Other complexes?

johnbachman:
OK! I implemented assembly for all Statement types except Translocations, which are non-essential for the moment. I also implemented the creation of `complex` nodes for Complex statements and for Agents with BoundConditions. This works recursively as @cthoyt suggested so that, for example, it is possible to create complexes containing complexes, etc. I also refactored the `_assemble_foobar` methods considerably so they're much more streamlined now.

In addition to the tests in `test_pybel_assembler`, there is also a model script in `models/simple_pybel` that creates a set of Statements, one of each type (except Translocation) and uploads to PyBEL web. I also ran the PyBEL assembler as is against the statements from the Phase 3 evaluation and it doesn't error, and I was able to upload to PyBEL web.

Here are a few minor todos that could be pushed off to the future:
* Add tests for evidence
* Add tests for different groundings
* Should negative bound conditions create inhibition edges between proteins and complexes?
* Activities for complexes could be drawn from the activity condition from a top-level agent with bound conditions
* Handling of mutations lacking complete information (e.g., `MutCondition(None, None, None)`)

cthoyt:
Awesome, nice work @johnbachman and @bgyori! 

I'm pushing the new version of PyBEL to PyPI right now to make dependency management easier for your next INDRA release.
 
I would need a bit more explanation about negative bound conditions. Could you give an example?

In regards to `MutCondition(None, None, None)`, I think there's already a solution from the [BEL specification](http://openbel.org/language/version_2.0/bel_specification_version_2.0.html#_variants_2), which is referencing HGVS:

>2.2.2. Variants
>variant(""), var("")
>
> The variant("<expression>") or var("<expression>") function can be used as an argument within a geneAbundance(), rnaAbundance(), microRNAAbundance(), or proteinAbundance() to indicate a sequence variant of the specified abundance. The var("") function takes HGVS variant description expression, e.g., for a substitution, insertion, or deletion variant. Multiple var("") arguments may be applied to an abundance term.
>
> Protein examples
>
>reference allele
>
>p(HGNC:CFTR, var("="))
>This is different than p(HGNC:CFTR), the root protein abundance, which includes all variants.
>
>unspecified variant
>
>p(HGNC:CFTR, var("?"))

johnbachman:
A negative bound condition states that something happens specifically when the protein is *not* bound to a certain other protein. For example, BCL2 binding to BAX prevents it from triggering apoptosis, so in principle you could end up with a statement saying effectively that `Bax(not bound to Bcl2) => apoptosis`. There are perhaps a couple of ways of rendering these in BEL/PyBEL:

* `complex(Bax, Bcl2) =| apoptosis`
* `Bcl2 =| (Bax => apoptosis)`
* `Bax => apoptosis; Bcl2 =| apoptosis` (don't like this one because the information that Bcl2 acts via Bax is lost)
* `Bcl2 =| act(Bax); act(Bax) => apoptosis`

PR #266, opened 2017-07-29
Move bel, biopax, index_cards, reach, sparser and trips modules to sources submodule

johnbachman:
Fairly straightforward--this PR moves these modules to a submodule, called processors. Up for discussion is whether we want to stick with the name "processors" for this submodule, or potentially something more obvious like "input"--but then to be symmetric we might want the assemblers submodule to be called "output".


bgyori:
I like the idea of moving these into a submodule. The name `processors` is not totally a good fit since what you actually import when doing, for instance, `from indra import biopax`, is import an API that gives you functions to access outside resources. These APIs are otherwise coupled to processors and ultimately extract INDRA Statements - but one isn't importing the processor per se. Alternatives could be `inputs`, `sources`, `knowledge`, `reading` but none of these is perfect.

Before merging this, a simultaneous update will need to be made in the Bioagents, and even in code that is being developed by others who use INDRA (for instance the TFTA).

bgyori:
It looks like the REST API is missing from the updated.

bgyori:
I actually like `sources` best out of all the options that came up so far but not 100% sure.

bgyori:
Actually a bunch of places are left unchanged, sometimes within the same file in which some changes were made, I'm going over things now.

johnbachman:
Great, thanks. I agree that "sources" is better than "input" because the database modules also provide "input" in a generic sense, whereas "sources" seems to imply "knowledge sources".

bgyori:
I renamed `processors` to `sources`, and looking over things again, I prefer `sources`. But we can revert it or find another alternative.

johnbachman:
Created PR for bioagents at sorgerlab/bioagents#30. Now the TFTA is the only one left?

johnbachman:
Rebased against master.

bgyori:
Ha, I already fixed Bioagents in a branch but I'll take your PR :) You can go ahead with this one if the tests are passing.

PR #278, opened 2017-08-30
cBioPortal client

pvtodorov:
New Python2/3 compatible client for cBioPortal for INDRA.
Current functionality focus on CCLE dataset for getting mutations.


bgyori:
Fixed and updated a bunch of things, and migrated the existing client from bioagents.

PR #279, opened 2017-08-31
Cbio client expanded functionality

pvtodorov:
- added function to get case lists for any study
- added function to get genetic profile data for any study
- added specialized function to get CCLE CNA data
- added tests for getting profile data and CCLE CNA data
- corrected minor PEP8 violations caught by linter in cbio_client and its tests

### DIFF
--- a/indra/databases/cbio_client.py
+++ b/indra/databases/cbio_client.py
@@ -122,7 +122,8 @@ def get_case_lists(study_id):
     return case_set_ids
 
 
-def get_profile_data(study_id, gene_list, profile_filter, case_set_id=None):
+def get_profile_data(study_id, gene_list,
+                     profile_filter, case_set_filter=None):
     """Return dict of cases and genes and their respective values.
 
     Parameters
@@ -140,7 +141,7 @@ def get_profile_data(study_id, gene_list, profile_filter, case_set_id=None):
         - COPY_NUMBER_ALTERATION
         - MRNA_EXPRESSION
         - METHYLATION
-    case_set_id : Optional[str]
+    case_set_filter : Optional[str]
         A string that specifices which case_set_id to use, based on a complete
         or partial match. If not provided, will look for study_id + '_all'
 
@@ -153,8 +154,8 @@ def get_profile_data(study_id, gene_list, profile_filter, case_set_id=None):
     genetic_profile = get_genetic_profiles(study_id, profile_filter)[0]
     gene_list_str = ','.join(gene_list)
     case_set_ids = get_case_lists(study_id)
-    if case_set_id:
-        case_set_id = [x for x in case_set_ids if case_set_id in x][0]
+    if case_set_filter:
+        case_set_id = [x for x in case_set_ids if case_set_filter in x][0]
     else:
         case_set_id = study_id + '_all'
         # based on looking at the cBioPortal, this is a common case_set_id
--- a/indra/databases/cbio_client.py
+++ b/indra/databases/cbio_client.py
@@ -357,7 +357,7 @@ def get_ccle_lines_for_mutation(gene, amino_acid_change):
     return cell_lines
 
 
-def get_ccle_cna(gene_list):
+def get_ccle_cna(gene_list, cell_lines):
     """Return a dict of CNAs in given genes and cell lines from CCLE.
 
     This is a specialized call to get_mutations tailored to CCLE cell lines.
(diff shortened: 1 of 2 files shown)


bgyori:
This looks great! Let me try it and then I'll merge.

indra/databases/cbio_client.py
```diff
@@ -265,6 +356,34 @@ def get_ccle_lines_for_mutation(gene, amino_acid_change):
...
+                                     frame_shift_del, splice_site
+
+    Returns
+    -------
+    profile_data : dict[dict[int]]
+        A dict keyed to cases containing a dict keyed to genes
+        containing int
+    """
+    profile_data = get_profile_data(ccle_study, ['BRAF', 'PTEN'],
+                                    'COPY_NUMBER_ALTERATION', 'all')
```
bgyori: - Looks like the parameters are hard coded here instead of the actual input arguments
- I think this function should take a cell_lines argument so that not every cell line in CCLE is returned every time, but only the ones specified in cell_lines
- The documentation Parameters should match the actual input arguments

indra/tests/test_cbio_client.py
```diff
@@ -51,3 +56,20 @@ def test_get_mutations_ccle():
...
+    assert profile_data['BT20_BREAST']['PTEN'] == -2
+    assert profile_data['BT20_BREAST']['BRAF'] == 1
+    assert profile_data['LOXIMVI_SKIN']['PTEN'] == 0
+    assert profile_data['LOXIMVI_SKIN']['BRAF'] == 0
+
+
+def test_get_ccle_cna():
+    profile_data = cbio_client.get_ccle_cna(['BRAF'])
+    assert profile_data['BT20_BREAST']['BRAF'] == 1
+    assert profile_data['LOXIMVI_SKIN']['BRAF'] == 0
```
bgyori: This test only works because of the hard-coded parameters in get_ccle_cna

indra/databases/cbio_client.py
```diff
@@ -88,12 +89,92 @@ def get_mutations(study_id, gene_list, mutation_type=None,
...
+        A list of genes with their HGNC symbols.
+        Example: ['BRAF', 'KRAS']
+    profile_filter : str
+        A string used to filter the profiles to return. Will be one of:
+        - MUTATION
+        - MUTATION_EXTENDED
+        - COPY_NUMBER_ALTERATION
+        - MRNA_EXPRESSION
+        - METHYLATION
+    case_set_id : Optional[str]
```
bgyori: I'd rename this argument to case_set_filter because it acts as a filter. Then lines 156-157 would be
```
if case_set_filter:
    case_set_id = [x for x in case_set_ids if case_set_filter in x][0]
```

indra/databases/cbio_client.py
```diff
@@ -88,12 +89,92 @@ def get_mutations(study_id, gene_list, mutation_type=None,
...
+            'genetic_profile_id': genetic_profile,
+            'gene_list': gene_list_str,
+            'skiprows': 2}
+    df = send_request(**data)
+    case_list_df = [x for x in df.columns.tolist()
+                    if x not in ['GENE_ID', 'COMMON']]
+    gene_list_df = df['COMMON'].tolist()
+    profile_data = _recursive_dict()
+    for case in case_list_df:
+        profile_values = df[case].tolist()
```
bgyori: This produces `nan` for missing values, I think it would be better to have `None` added to the result for missing values.

indra/databases/cbio_client.py
pvtodorov: Thanks for catching this!
- Fixed hardcoded list in a610f4c
- Updated tests in  5860b83
- parameters matched to documentation

indra/tests/test_cbio_client.py
pvtodorov: - Fixed hardcoded list in function with commit a610f4c
- Updated tests in commit 5860b83

indra/databases/cbio_client.py
pvtodorov: Renamed parameter in code and tests with commit c22f95b

indra/databases/cbio_client.py
pvtodorov: Refactored code to use dict comprehension instead of `_recursive_dict()` function. In addition, it now produces `None` instead of `nan` for missing values. Commit b52febe

PR #281, opened 2017-09-01
Processor for CX networks from NDEX

johnbachman:
This PR adds a new processor to indra.sources for getting content from NDEx-based networks in the CX format. It converts a subset of the edges into INDRA Statements, currently the following:

```
_stmt_map = {
    'phosphorylates': Phosphorylation,
    'controls-phosphorylation-of': Phosphorylation,
    'in-complex-with': Complex,
    'NEGATIVE_INFLUENCE': Inhibition,
    'POSITIVE_INFLUENCE': Activation
}
```

Support for interacts-with, controls-state-change, etc. could be added but would most properly require new INDRA Statement types. Evidence information is added to the extracted Statements using the ndex:citation field and the ndex network ID as the source_id.


bgyori:
I really like this! I started adding this to the documentation when I realized that for consistency it might be better to have the content of __init__.py in ndex_cx_api.py with __inti__.py just importing the API functions from there.

johnbachman:
Funny you mention that. As I was doing this, I was looking at the other source APIs and thinking, "Hmm, maybe in all these other modules we should just move everything out of foobar_api.py and into __init__.py..." 😄

johnbachman:
Also, adding to the documentation would be great--though I was thinking might be worth doing an iteration with @dexterpratt first on how they're going to normalize edge types, etc.

Issue #282, opened 2017-09-05
cbio_client: max numbers of genes that can be requested is limited for CCLE CNA


pvtodorov:
cBioPortal client cannot request info about more than 15 genes at once. As per the following code, the request returns empty dataframes after 15 genes. Appears to be isolated to `get_profile_data()` and `get_ccle_cna()`. This does not seem to be an issue in the case of `get_mutations_ccle()`.

Input
```
from indra.databases import cbio_client

[...]

PR #283, opened 2017-09-06
Cbio client

pvtodorov:
Fixes https://github.com/sorgerlab/indra/issues/282


indra/databases/cbio_client.py
```diff
@@ -172,7 +172,7 @@ def get_profile_data(study_id, gene_list,
     for case in case_list_df:
         profile_values = df[case].tolist()
         for g, cv in zip(gene_list, profile_values):
-            if isinstance(cv, int):
+            if cv == cv:
```
bgyori: Umm... there's got to be a better way to do this.

PR #293, opened 2017-09-11
Add default config loading to RAS Machine

### LINKED ISSUE #292: Default configuration loading for RAS Machine
Rather than writing the redundant command `python rasmachine.py testmodel testmodel/config.yaml`, the script should check for the standard config file, since it's already made by the `new_machine.sh` shell script. If it's omitted and the default config file doesn't exist, it will send the same error that the argument is missing.

cthoyt:
Closes #292

1. Checks if argument is explicitly given and uses that
2. Checks if default location exists and uses that
3. Fails


johnbachman:
Looks good to me.

PR #300, opened 2017-09-15
Refactor NDEx network creation/update from CXAssembler and Ras Machine into ndex_client

### LINKED ISSUE #236: CxAssembler upload model should return UUID and print human-friendly URL
Currently it returns a link of the form
http://public.ndexbio.org/v2/network/<network-id> which goes to a machine-readable view of the network.

johnbachman:
Moves code to update networks from rasmachine.py and puts in ndex_client.update_network. Moves upload_model method from CXAssembler to a create_network function. Also fixes and closes #236.

### DIFF
--- a/indra/databases/ndex_client.py
+++ b/indra/databases/ndex_client.py
@@ -4,8 +4,9 @@ import requests
 import json
 import time
 import logging
+import ndex
 
-logger = logging.getLogger('ndex')
+logger = logging.getLogger('ndex_client')
 
 ndex_base_url = 'http://52.37.175.128'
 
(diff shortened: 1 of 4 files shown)


indra/databases/ndex_client.py
```diff
@@ -67,3 +70,129 @@ def send_request(ndex_service_url, params, is_json=True, use_get=False):
...
+            break
+        except Exception as e:
+            logger.error('Could not update NDEx network profile.')
+            logger.error(e)
+
+    # Update network style
+    import ndex.beta.toolbox as toolbox
+    template_uuid = "ea4ea3b7-6903-11e7-961c-0ac135e8bacf"
+
+    d_edge_types = ["Activation", "Inhibition",
```
bgyori: We can actually remove this d_edge_types thing, it's left over from something else.

PR #314, opened 2017-09-19
Update API, tweak cBio client in order to use it with Interactive Pathway Map

pvtodorov:
- added API endpoints for `cbio_client` functions:
    - `get_mutations_ccle`
    - `get_ccle_mrna`
    - `get_ccle_cna`
- minor change to `cbio_client.get_profile_data` function
    - now uses adaptive line skipping which was previously implemented but unused in this function

### DIFF
--- a/indra/databases/cbio_client.py
+++ b/indra/databases/cbio_client.py
@@ -424,6 +424,8 @@ def get_ccle_mrna(gene_list, cell_lines):
                 value_cell = df[cell_line][df['COMMON'] == gene]
                 if value_cell.empty:
                     mrna_amounts[cell_line][gene] = None
+                elif pandas.isnull(value_cell.values[0]):
+                    mrna_amounts[cell_line][gene] = None
                 else:
                     value = value_cell.values[0]
                     mrna_amounts[cell_line][gene] = value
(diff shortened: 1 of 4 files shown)


bgyori:
Could you also update rest_api/docs/api_spec.js? Thanks!

pvtodorov:
- Documented new functions in API using Swagger format (d302b55, 80cc72f, fca7918) and the objects that the API consumes and produces as a result of these changes (8d36568)
- Changed name of function `get_mutations_ccle` to `get_ccle_mutations` in `cbio_client` as well as the API in order to be consistent with other function names
  - change to function name in `cbio_client` cb3b586
  - change to test 06d5e93
  - change to function name in api 937320f
- Other modifications
  - `get_ccle_mrna`  returns `None` instead of `NaN`. This is more compatible with JSON format and consistent with the way `get_ccle_cna` which also returns `None` in case of `NaN`. In commit e1b7ca9.
  - API sends JSON directly instead of encoding as string for CCLE functions. This is consistent with the way other API endpoints respond and the JSON doesn't have to be parsed from string in the pathway map. In commit 1d021b6.
  - Updated `cbio_client.get_ccle_cna` docstring to indicate what the numbers returned by the funtion actually mean. In commit 61a3643.

Issue #315, opened 2017-09-19
Incorporate RAS machine to main repository and add CLI


cthoyt:

cthoyt:
I think this issue was also closed with merge #316

PR #316, opened 2017-09-19
Incorporate RAS Machine to main repository

cthoyt:
Automatically add a CLI so you can run commands from the command line with `indra machine make ~/Desktop/new_machine` or `indra machine run_with_search ~/Desktop/new_machine`

Later it will be easy to add other ones

References #315


cthoyt:
There's a problem when putting the rest of the machine commands into `indra.cli` though with importing `jnius_config` that I'll need a consult on @bgyori

cthoyt:
@bgyori Okay I think this PR has gotten a bit ridiculous (my bad)

Let me know which bits you want to keep (if any - I won't be offended) and I'll make a different one with just those parts

bgyori:
Thanks a lot! I can review this on Tuesday and get back to you or merge if it's good to go.

cthoyt:
Looks good - It would still be possible to have python install an entry point like `indra-machine` that points directly to that CLI if you'dlike

bgyori:
Sure, I'm not sure how to add it now that I moved everything inside indra.tools.machine - so if that would be useful feature for you, could you add that? After that I'll merge!

cthoyt:
All set! To get this CLI element working, you'll have to run `pip install ...` again even if you've already installed in editable mode with `-e`

Thanks for checking this code over and making use of it!

bgyori:
Fantastic, thanks a lot!

Issue #149, opened 2017-01-18
Assembly steps need to extend to bound conditions


bgyori:
Agents in bound conditions are not name standardized, not grounding mapped, etc. These then cause consistency problems in assembled models.

bgyori:
Fixed in #461

PR #324, opened 2017-09-24
PybelAssembler Fixes

cthoyt:
Is there a reason for using += instead of .extend? I would prefer .extend because it doesn't require a new list to be made (and this would be good for bigger lists)

### DIFF
--- a/indra/assemblers/pybel_assembler.py
+++ b/indra/assemblers/pybel_assembler.py
@@ -64,12 +64,12 @@ class PybelAssembler(object):
             self.statements = stmts
 
         # Create the model and assign metadata
-        self.model = pybel.BELGraph(**kwargs)
-        self.model.graph[pc.GRAPH_METADATA] = {
-                pc.METADATA_NAME: name,
-                pc.METADATA_DESCRIPTION: description,
-                pc.METADATA_VERSION: version
-            }
+        self.model = pybel.BELGraph(
+            name=name,
+            version=version,
+            description=description,
+            **kwargs
+        )
 
     def add_statements(self, stmts_to_add):
         self.statements += stmts_to_add
--- a/indra/assemblers/pybel_assembler.py
+++ b/indra/assemblers/pybel_assembler.py
@@ -72,7 +72,7 @@ class PybelAssembler(object):
             }
 
     def add_statements(self, stmts_to_add):
-        self.stmts += stmts_to_add
+        self.statements += stmts_to_add
 
     def make_model(self):
         for stmt in self.statements:
--- a/indra/assemblers/pybel_assembler.py
+++ b/indra/assemblers/pybel_assembler.py
@@ -104,7 +104,26 @@ class PybelAssembler(object):
             else:
                 logger.info('Unhandled statement: %s' % stmt)
         return self.model
+    
+    def save_model(self, path, output_format=None):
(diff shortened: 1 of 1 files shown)


cthoyt:
It seems like there's still a problem with grounding for `PFAM` that makes output to BEL Script impossible (because the dictionary lookup in PFAM returns None)

indra/assemblers/pybel_assembler.py
```diff
@@ -104,7 +104,26 @@ def make_model(self):
...
-
+    
+    def save_model(self, path, output_format=None):
+        """Saves the :class:`pybel.BELGraph` using one of the outputs from 
+        :py:mod:`pybel`
+
+        :param str path: The path to output to
+        :param Optional[str] output_format: Output format as ``cx``, 
+                            ``pickle``, ``json`` or defaults to ``bel``
+        """
```
johnbachman: Can switch these docstrings to the numpy doc format used elsewhere in INDRA, for consistency?

indra/assemblers/pybel_assembler.py
```diff
@@ -171,6 +190,9 @@ def _assemble_active_form(self, stmt):
     def _assemble_complex(self, stmt):
         """Example: complex(p(HGNC:MAPK14), p(HGNC:TAB1))"""
         complex_data, _ = _get_complex_node(stmt.members)
+        if complex_data is None:
+            logger.info('skip adding complex with no members: %s', stmt.members)
+            return
         self.model.add_node_from_data(complex_data)
```
johnbachman: Interesting--do you have any examples of the complexes where this has occurred? I assume this is because of the failure to generate one or more nodes.

cthoyt: These all had problems:

```
Activation(CD106(bound: [CD29, True]), adhesion())
Inhibition(GAPs__GTPase_activating_proteins(bound: [Rac1_GTP, True]), RAC1())
Activation(rhHSP60(bound: [OPCs, True]), apoptosis())
Activation(XBP1s(bound: [its, True]), transcription())
```

johnbachman: Aha! This must be because we haven't yet implemented grounding mapping or filtering for ungrounded entities in BoundConditions--see issue #149.

cthoyt: Thanks for looking into that. Where do you think the best place to check if statements like these should be skipped? For some reason, placing this if statement here doesn't sit well with me since it doesn't really document what the problem is

johnbachman: If the issue is what I think it is, you would have seen the warning "Agent foo has no grounding" from _get_agent_node_no_bcs, below (line 312). Did that show up in the logs?

cthoyt: Indeed. Here's a link to the pickle of the offending IncrementalModel I made with the RAS machine: https://owncloud.scai.fraunhofer.de/index.php/s/5jJHSwIQ2Ci1Uf5

cthoyt:
@johnbachman shall we keep these checks, or will this be blocked by #149?

johnbachman:
Keep that code block checking for empty complexes--it's a good safeguard. But the underlying issue to resolve is going to be #149.

johnbachman:
Great, looks like all relevant tests pass. Will rebase and merge.

Issue #321, opened 2017-09-23
Processor for Signor dataset


johnbachman:
Signor contains ~18k curated mechanisms, often incorporating both the mechanism (e.g. phosphorylation) and effect (increase/decrease activity).

PR #326, opened 2017-09-27
Signor Processor

johnbachman:
Adds a processor for the interaction data in Signor (see http://signor.uniroma2.it). The processor is implemented as a single module, `indra/sources/signor.py`. The data can either be downloaded from the Signor site at the point of use (which is the default behavior) or cached locally, in which case the SignorProcessor is initialized with a path to the CSV file. The SignorProcessor can be instantiated very simply as follows:

```
from indra.sources.signor import SignorProcessor
sp = SignorProcessor()
```

The extracted statements are then contained in `sp.statements`.

Each row in the Signor dataset contains an "EFFECT" relation, which includes up/down regulation and complex formation. This information is used to create `Activation`, `Inhibition`, `IncreaseAmount`, and `DecreaseAmount` Statements. Most rows (though not all) also contain MECHANISM relations, which can be quite detailed. Not all of these can be mapped to INDRA Statements at present; the ones that are mapped are created and others are stored for inspection in the `SignorProcessor.no_mech_rows` and `SignorProcessor.no_mech_ctr` attributes. Further, some effect/mechanism pairs describe only a single mechanism (for example, the EFFECT 'up-regulation' coupled to the MECHANISM 'transcriptional regulation'). In this case only a single statement is generated. Finally, modification and binding events coupled to an up or down regulation are also used to generate `ActiveForm` Statements with `ModConditions` or `BoundConditions` set on the target agent, respectively.

As of today the Signor dataset, with 18,562 rows, produces 44,087 INDRA Statements, with 388 mechanism relations not giving rise to INDRA Statements and distributed as follows:

```
 ('relocalization', 184),
 ('cleavage', 93),
 ('small molecule catalysis', 65),
 ('transcriptional regulation', 18), # Note: these are due to unknown polarity
 ('neddylation', 11),
 ('tyrosination', 7),
 ('post transcriptional regulation', 4),
 ('translation regulation', 3),
 ('s-nitrosylation', 1),
 ('oxidation', 1),
 ('lipidation', 1)]
```

Closes issue #321.

Some known issues:
* SignorProcessor should be included in the documentation
* The generic "up-regulates" effect type should ultimately be mapped to a generic up regulation rather than Activation/Inhibition, as it is currently.
* Mappings for SIGNOR families, complexes, etc. (see Bioentities issues)
* Need a Statement similar to ActiveForms but representing effects on amounts ("StateEffect")
* Implement translocation statements

### DIFF
--- a/indra/sources/signor.py
+++ b/indra/sources/signor.py
@@ -1,0 +1,25 @@
+from __future__ import absolute_import, print_function, unicode_literals
+from builtins import dict, str
+from indra.util import read_unicode_csv
+
+
+class SignorProcessor(object):
+    """Processor for Signor dataset available at http://signor.uniroma2.it.
+
+    See publication:
+
+    Perfetto et al., "SIGNOR: a database of causal relationships between
+    biological entities," Nucleic Acids Research, Volume 44, Issue D1, 4
+    January 2016, Pages D548–D554. https://doi.org/10.1093/nar/gkv1048
+
+    Parameters
+    ----------
+    signor_csv : str
+        Path to SIGNOR CSV file.
+    delimiter : str
+        Field delimiter for CSV file. Defaults to semicolon ';'.
+    """
+    def __init__(self, signor_csv, delimiter=';'):
+        self._data = list(read_unicode_csv(signor_csv, delimiter=';'))
+        
+
--- a/indra/sources/signor.py
+++ b/indra/sources/signor.py
@@ -237,7 +237,7 @@ class SignorProcessor(object):
             stmts.append(IncreaseAmount(agent_a, agent_b, evidence=evidence))
         elif row.MECHANISM == 'destabilization' and \
              row.EFFECT in ('down-regulates', 'down-regulates quantity',
-                            'down-regulates quantity by stabilization'):
+                            'down-regulates quantity by destabilization'):
             stmts.append(DecreaseAmount(agent_a, agent_b, evidence=evidence))
         elif row.MECHANISM == 'chemical activation' and \
              row.EFFECT in ('up-regulates', 'up-regulates activity'):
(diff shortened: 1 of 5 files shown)


indra/sources/signor.py
```diff
@@ -0,0 +1,379 @@
...
+                'TISSUE_DATA': _n(row.TISSUE_DATA),
+                'MODULATOR_COMPLEX': _n(row.MODULATOR_COMPLEX),
+                'TARGET_COMPLEX': _n(row.TARGET_COMPLEX),
+                'MODIFICATIONA': _n(row.MODIFICATIONA),
+                'MODASEQ': _n(row.MODASEQ),
+                'MODIFICATIONB': _n(row.MODIFICATIONB),
+                'MODBSEQ': _n(row.MODBSEQ),
+                'NOTES': _n(row.NOTES),
+                'ANNOTATOR': _n(row.ANNOTATOR)}
+        return Evidence(source_api='SIGNOR', source_id=row.SIGNOR_ID,
```
bgyori: Not a big deal but we use lowercase for all other sources, so 'signor' might be better. Could you add some parameters to `indra.belief.__init__.py` for 'signor' too?

indra/sources/signor.py
```diff
@@ -0,0 +1,389 @@
...
+            # Process into a list of SignorRow namedtuples
+            # Strip off any funky \xa0 whitespace characters
+            self._data = [SignorRow(*[f.strip() for f in r]) for r in data_iter]
+        # If no CSV given, download directly from web
+        else:
+            res = requests.post('http://signor.uniroma2.it/download_entity.php',
+                                data={'organism':'human', 'format':'csv',
+                                      'submit':'Download'})
+            if res.status_code == 200:
+                csv_reader = csv.reader(StringIO(res.text), delimiter=delimiter,
```
bgyori: The delimiter here needs to be bytes in Python 2. After trying to change that, line 168 gives a Unicode error.

johnbachman: Fixed in d9d3ab32d.

PR #341, opened 2017-11-03
Chembl

pvtodorov:
- Pull request for integrating ChEMBL client in INDRA
- ability to query ChEMBL API endpoints using `send_query`
    - specialized query to get targets
- ability to generate Inhibition statement from ChEMBL activity lookup given a drug and protein target as Agents
- ability to generate Inhibition statements from ChEMBL activity lookup for all `SINGLE PROTEIN` targets of a drug, given that drug as an Agent


bgyori:
Great, thanks!

PR #351, opened 2017-12-14
Clean up and PEP8-ify Statements

pagreene:
Tidy the statements file, and fix some bugs.

### DIFF
--- a/indra/statements.py
+++ b/indra/statements.py
@@ -133,9 +133,7 @@ import sys
 import uuid
 import rdflib
 import logging
-import textwrap
 import networkx
-from collections import namedtuple
 from collections import OrderedDict as _o
 from indra.util import unicode_strs
 import indra.databases.hgnc_client as hgc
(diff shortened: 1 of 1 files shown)


bgyori:
This doesn't seem right, evidence is set in the parent class, and doesn't need to be set in child classes. Note also that these are not API methods, they are underscored  on purpose. The extensive cosmetic work makes it very hard to review. If something is/was broken, please resubmit with a corresponding test. Thanks!

pagreene:
The cosmetic work is the reason I suggested it would be easier to review by
commit rather than by file. There is a TODO that you might want to take a
look at regarding another variable, 'uid' that is set and, as far as I can
tell, not used. You can just search for TODO, and you should find the place
easy enough.

I do see now how the evidence is set. I wouldn't have tidied the code if
there hadn't apparently been something to fix, but it seems silly to not
apply the changes now that they're done. It is a central file, and I think
it be nice to have it look as professional as possible. Again review by
commit would be easier, I think.

Cheers,
Patrick

bgyori:
Alright, there turns out not to have been an issue with evidences but this PR PEP8-ifies statements.py, and removes some unnecessary code.

PR #353, opened 2017-12-21
Improve Travis configuration and other fixes

bgyori:
This PR eliminates Anaconda from the Travis setup and instead uses pip to install dependencies, which is expected to be much more stable. Since packages like scipy and pandas are installed from wheels, this doesn't slow down the build. It adds aspcud to improve the Opam install. The PR also fixes some other minor bugs that caused tests to fail. Finally, some tests that depend on third party services that are known to be unavailable currently are skipped.


bgyori:
Issue #355 is the only remaining test failure after these fixes.

PR #357, opened 2018-01-04
Update eidos_api.py to be compatible with Python 3

adarshp:


bgyori:
Thanks for catching that!

adarshp:
No problem!

PR #360, opened 2018-01-04
Fix model checker sampling tests

pagreene:
Introduces fixes to random selections in the model checker to allow for reliable use of random seeds.

### DIFF
--- a/indra/explanation/cycle_free_paths.py
+++ b/indra/explanation/cycle_free_paths.py
@@ -12,7 +12,7 @@ cycle-free paths. However, we are able to "easily" sample cycle-free paths from
 The algorithm is described further in :py:func:`cycle_free_paths_graph`, below.
 """
 
-import random
+import numpy as np
 import itertools
 from copy import copy, deepcopy
 import networkx as nx
@@ -314,7 +314,7 @@ def _cf_succ(H, t, path, v):
     for u in H.successors(v):
         if set(path) <= set(t[u]):
             succ.append(u)
-    w = random.choice(succ)
+    w = np.random.choice(succ)
     return w
 
 
--- a/indra/explanation/cycle_free_paths.py
+++ b/indra/explanation/cycle_free_paths.py
@@ -314,6 +314,8 @@ def _cf_succ(H, t, path, v):
     for u in H.successors(v):
         if set(path) <= set(t[u]):
             succ.append(u)
+    # Note that the circuitous way of choosing from this list is the result of
+    # the odd way numpy.random handles lists of lists (it excepts).
     idx_list = list(range(len(succ)))
     w_idx = np.random.choice(idx_list)
     w = succ[w_idx]
--- a/indra/explanation/cycle_free_paths.py
+++ b/indra/explanation/cycle_free_paths.py
@@ -314,7 +314,9 @@ def _cf_succ(H, t, path, v):
     for u in H.successors(v):
         if set(path) <= set(t[u]):
             succ.append(u)
-    w = np.random.choice(succ)
+    idx_list = list(range(len(succ)))
+    w_idx = np.random.choice(idx_list)
+    w = succ[w_idx]
     return w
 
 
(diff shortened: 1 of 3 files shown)


indra/explanation/cycle_free_paths.py
```diff
@@ -314,7 +314,9 @@ def _cf_succ(H, t, path, v):
     for u in H.successors(v):
         if set(path) <= set(t[u]):
             succ.append(u)
-    w = random.choice(succ)
+    idx_list = list(range(len(succ)))
+    w_idx = np.random.choice(idx_list)
+    w = succ[w_idx]
```
bgyori: Are these 3 lines necessary? Wouldn't just using ```w = np.random.choice(succ)``` be the same?

indra/explanation/paths_graph.py
```diff
@@ -3,6 +3,7 @@
 import numpy as np
 import networkx as nx
 from indra import logging
+import sys
```
bgyori: This import is probably not needed, right?

indra/tests/test_model_checker.py
```diff
@@ -1152,15 +1166,27 @@ def test_weighted_sampling1():
...
+    import os
+    import pickle
+    i = 0
+    fname = 'case_py%d_%d.pkl' % (sys.version_info.major, i)
+    while fname in os.listdir('.'):
+        i += 1
+        fname = 'case_py%d_%d.pkl' % (sys.version_info.major, i)
+    with open(fname, 'wb') as f:
+        pickle.dump((pa.model, results, path_ctr), f, protocol=2)
+    # ---------------------------------------------------------------
```
bgyori: This block should be removed, I think.

pagreene:
You'd think, but that's to avoid the weird way np.random handles lists of
lists (it raises an exception). If you're just trying to pick a random
list, you have to specify that by creating an index and choosing from that.
You can actually see basically the same technique used in the path graph, I
think. Or perhaps it was a different instance of random.

indra/explanation/cycle_free_paths.py
pagreene: You'd think, but that's to avoid the weird way np.random handles lists of
lists (it raises an exception). If you're just trying to pick a random
list, you have to specify that by creating an index and choosing from that.
You can actually see basically the same technique used in the path graph, I
think. Or perhaps it was a different instance of random.

indra/tests/test_model_checker.py
pagreene: Yup, I forgot about that. Thanks!

indra/explanation/paths_graph.py
```diff
@@ -287,14 +287,15 @@ def sample_single_path(pg, source, target, signed=False, target_polarity=0,
                 out_edges = pg.out_edges(current_node, data=True)
             else:
                 out_edges = pg.out_edges(current_node)
+            out_edges.sort()
```
johnbachman: Is there any way to avoid this sort call? Sampling single paths will get called over and over and having to run a sort in the innermost loop here may have a high cost.

pagreene: Well it has to be regularized some how. Perhaps we can set a parameter to make it optional...? Let me take a look...

pagreene: That would be tough. Perhaps if we had some kind of global setting to see if we're in a test?

bgyori: Not pretty but we could do ```
if 'TRAVIS' in os.environ:
    out_edges.sort
```

pagreene:
That wouldn't help when testing locally though.

indra/explanation/paths_graph.py
pagreene: That wouldn't help when testing locally though.

pagreene: We might be able to introspect into the stack and see if nose is in it.

bgyori:
Wow, tests pass!

Issue #354, opened 2017-12-21
Tests with special requirements should be tagged to enable turning them on or off


bgyori:
We should define a set of tags for tests, some basic categories being:
- `slow` for tests that are slow, like reading and processing a full paper
- `webservice` for tests that use outside web services
- `nonpublic` for tests that rely on non-public environment variables to run
with possibly other ones that seem useful.

There are several ways to implement this, and the one that to me seems like the best fit is using `nose.plugins.attrib` (http://nose.readthedocs.io/en/latest/plugins/attrib.html), which lets you add a decorator to a test like
```
from nose.plugins.attrib import attr
@attr(speed='slow')
def test_something_slow():
    ...
```
and then run tests skipping a specific tag as
```
nosetests -a '!slow'
```
Given a set of tags, we should review all tests in INDRA and apply tags as needed.

A useful further step will be setting up the Travis configuration to take arguments to `nosetests -a` as environment variables. This would let us define that, for instance, PRs to sorgerlab should not run `nonpublic` tests since hidden environment variables are not enabled in PRs, and these tests always fail.

pagreene:
Would 'nonpublic' environment variables include database urls?

bgyori:
Yes, if the database is not publicly accessible. The point would be to tag everything with 'nonpublic' that would fail in a PR due to missing environmental variables.

cthoyt:
When I write tests that require environment variables, I have nose skip them if they are unavailable
```python
import unittest
import os

class TestStuff(unittest.TestCase):

    @unitttest.skipUnless('CONNECTION' in os.environ, reason='missing connection')
    def test_thing(self):
        ...
```

pagreene:
Issue resolved with #362.

PR #362, opened 2018-01-05
Tag tests with attributes

pagreene:
Implement some attributes to the nosetests. Use these attributes to skip tests that require nonpublic environment variables during a pull request in travis tests. This pull request should test that feature.
 
Addresses: https://github.com/sorgerlab/indra/issues/354

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,9 +71,9 @@ script:
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
       export NOSE_IGNORE_FILES="manage_content.py,build_sample_set.py,_require_python3.py";
     fi
-  - export NOSEATTR=""
+  - export NOSEATTR="";
   - if [[ $TRAVIS_PULL_REQUEST ]]; then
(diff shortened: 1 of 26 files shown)


bgyori:
This looks good! The only issue is that a number of tests (for instance in the HGNC and UniProt clients) that are currently tagged as webservice actually use off-line resource files. I will take a look at these and remove the webservice tag where applicable.

pagreene:
Fair enough. I was in many cases guessing. And there is the opposite problem in some cases, such as with the trips tests, which I think go out to a web service. Feel free to add more tags, especially nonpublic tags, where fitting. I suspect I missed some of those, and thus the travis test will not be green, but at least won't have not failed for some of the wrong reasons.

pagreene:
In my first "Make s3 nonpublic" commit, I had accidentally edited the pmc file, not the s3 file. Ben's "Fix wrong webservice/nonpublic tags" commit fixed those bugs.

pagreene:
Well, the change to travis definitely seems to be working. Now it just a matter of actually classifying things correctly.

PR #368, opened 2018-01-15
Improving high-throughput reading and utilities

bgyori:
This PR adds a memory constraint to REACH and raises an Exception if there is an error while running REACH. It also improves some of the utilities to monitor reading jobs.


pagreene:
I pushed the proper fix to the list issue to the sorgerlab version of this branch. That should probably be applied before merging.

bgyori:
Alright, the failing test is unrelated.

PR #370, opened 2018-01-16
Move reach upload

pagreene:
Make reach upload it's results prior to processing into statements for better recovery.


bgyori:
Look good!

PR #371, opened 2018-01-16
Reorganize the database manager 

pagreene:
Sort out the different tools of the database interface system to separate out utilities from core interface tools, and provide a more pythonic interface.


bgyori:
Oops, I didn't realize this wasn't rebased, let me do some surgery.

PR #373, opened 2018-01-17
Add stalled log timeout option to wait_for_complete

pagreene:
Makes it possible to set a timeout on non-updating logs for aws batch jobs, which helps catch jobs that have stalled. There is also an option to automatically kill such jobs, for ease of mind when starting a large job before a vacation.

The command line wait_for_complete tool was also generally improved, using argparse to access all of its features.


bgyori:
Looks good!

PR #376, opened 2018-01-18
Fix sparser cache

pagreene:
Make sparser process cached results in old reading pipeline.

### DIFF
--- a/indra/tools/reading/read_pmids.py
+++ b/indra/tools/reading/read_pmids.py
@@ -475,7 +475,9 @@ def run_sparser(pmid_list, tmp_dir, num_cores, start_index, end_index,
             )
         logger.info("Mapping get_stmts onto pool.")
         unread_res = pool.map(get_stmts_func, batches)
+        logger.info('len(unread_res)=%d' % len(unread_res))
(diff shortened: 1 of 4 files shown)


indra/tools/reading/read_pmids.py
```diff
@@ -445,11 +445,19 @@ def get_stmts(pmids_unread, cleanup=True, sparser_version=None):
     return stmts
 
 
+def get_stmts_from_cache(pmid):
+    json_str = s3_client.get_reader_json_str('sparser', pmid)
+    stmts = []
+    if json_str is not None:
+        stmts = sparser.process_json_dict(json.loads(json_str))
```
bgyori: I haven't tested this but this doesn't look right - wouldn't sparser.process_json_dict return a SparserProcessor, not a list of Statements? Has this worked in practice?

pagreene: Ah, yes, it does. I misread it when I saw `sp.get_statements()`.

pagreene: Ok...this got much harder to push the fix because of the rebase...but I have the fix ready.

bgyori:
Great, this is working on some examples I ran

PR #379, opened 2018-01-19
Restructuring the reading submodule

bgyori:
This PR restructures the reading submodule, and introduces additional structure to it by separating files according to use cases:
- pmid_reading: reading a list of PMIDs according to best available (cached) content and producing pickle files on S3
- db_reading: reading a list of files and adding it to the database
- starcluster_reading: reading utilities that assume that scripts are run from within a cluster instance (mostly deprecated)
- util: various helper functions and scripts

The PR specifically heavily restructures the pmid_reading submodule (previosuly read_pmids.py) by breaking it up into multiple files, simplifying the naming of functions, and documenting their purpose. 

Tests are passing, and the PMID reading use case is working robustly (tested with multiple machines, and 3 other large reading jobs). Note, however, that the database reading script (now in a new submodule called db_reading) has not been tested on real jobs yet.

Due to the extensive changes, this PR is probably best reviewed by checking out the branch and browsing the files offline.


pagreene:
A couple of thoughts:
- Why not use the reader objects for the pmid reading? All they need is a list of file locations, so it should be fairly strait-forward to integrate them. That would then also reduce redundant code. It is something I have planned to do for a while but haven't gotten around to doing.
- We really need more tests for the old reading system. With all this refactoring, there's a good chance something broke, and the tests won't catch it, as they stand. And honestly, we keep testing by running large readings, and finding months later that something hadn't actually worked right. Reducing redundant code would also help this cause.
- The read_db use case needs to be fixed as well, as the tests are currently erroring. (this won't show up on Travis because we still haven't gotten travis to play nice with any databases.)

I'm working on the read_db bug(s?) as we speak.

bgyori:
Since the github diff highlight is confusing, here is a little bit more detail about what changed:
- The new core reading functionalities readers.py and read_files.py are unchanged (except for some trivial updates in imports)
- The database reading scripts read_db.py and read_db_aws.py are unchanged but are now in their own db_reading submodule
- Each file that uses bsub or qsub is unchanged but now resides in a separate starcluster_reading submodule
- read_pmid.py contained 4 visually separated sections which are now broken out into the pmid_reading submodule as separate files (reach_reading, sparser_reading, get_content and util), while the top-level script which used to be read_pmid_aws.py is now simply in pmid_reading/__init__.py. Many of the functions' _names_ have changed (i.e. been simplified) but their internal implementation is almost completely unchanged.
- The central submission script was renamed from submit_reading_pipeline_aws.py to submit_reading_pipeline.py (there was a deprecated submit_reading_pipeline.py which is now in starcluster_reading).

bgyori:
To @pagreene's comments:
- Yes, going forward, a next restructuring could involve creating a pipeline that replaces the core of pmid_reading/read_reach and reach_sparser with the implementation in readers.py
- I'm satisfied with the test runs I did to the extent that I don't believe new problems were introduced to the pmid_reading machinery in this PR, but of course more tests would be good.
- I think that can be resolved by correctly updating import and execution paths since the actual content of the db reading files hasn't changed

bgyori:
The conflicts are too complicated to sort out at this point; instead, the submodule will be reorganized along similar lines but in smaller steps.

PR #380, opened 2018-01-19
Geneways processor

djmilstein:
Processor for converting Geneways extraction data into INDRA statements


bgyori:
This looks great! In the commits I just pushed, I made a number of changes for various reasons:
- I changed the API call to assume a single folder to look for the files, and exposed it in __init__.py.
- I added some handling for missing dependencies (note that nltk and stemming are not universtal INDRA dependencies so they need to be installed specifically for this input source) and added an underscore to the full text finding file so that Travis doesn't error when it tries to run doctests.
- I made many small changes in all the files to change the style e.g. remove semicolons and extra spaces from the ends of lines, change camel case variable and function names to all lower case with underscores, etc.
- I merged the statement generator into the processor for simplicity, it's easier to have these colocated.
I hope I didn't break anything.

Before merging this, I think we should make two more additions: the entrez-to-hgnc lookup should be made more efficient in indra.databases.hgnc_client since this slows down processing significantly (I will work on this). Also, we should have at least a few tests for the API, processor and the various other units. Individual functions can be tested with synthetic arguments, while the API could be tested with a set of mock input files that match the format of the real ones (@djmilstein could you work on these?).

djmilstein:
Thank you very much for the polishing! And will do, I'll write some tests.

On Mon, Jan 22, 2018 at 3:45 PM, Benjamin M. Gyori <notifications@github.com

bgyori:
Alright, the trivial change in the HGNC client I just pushed speeds up the processor by a lot, and it's not slow in Python 2 anymore either so I removed that warning.

bgyori:
And in this last commit, I added name standardization for getting Agents (based on the HGNC ID, the standard gene name is set as the Agent's name), which is how the other input processors work (even though in principle it could be done at a later stage of assembly).

djmilstein:
Huzzah! I've now added tests for the Geneways processor.

On Mon, Jan 22, 2018 at 4:26 PM, Benjamin M. Gyori <notifications@github.com

bgyori:
Test are passing on my pushed branch so I'll merge this now. Great job @djmilstein !

PR #381, opened 2018-01-19
Enhance pipeline logs

pagreene:
This PR primarily does two things:
- `wait_for_complete` can now stash logs from batch jobs both locally and on s3.
- the logs produced by sparser processes are now preserved and placed on s3.
These apply to both the database reading system and the non-db reading system.


pagreene:
I think this should be all set now.

Cheers!

indra/tools/reading/read_db_aws.py
```diff
@@ -91,7 +98,25 @@
...
     # Read everything ========================================
     outputs = produce_readings(id_dict, readers, verbose=True,
-                               read_mode=args.mode)
+                               read_mode=args.mode,
+                               force_fulltext=args.force_fulltext)
+
+    # Preserve the sparser logs
+    contents = os.listdir('.')
+    sparser_logs = [fname for fname in contents
+                    if fname.startswith('sparser') and fname.endswith('log')]
```
bgyori: You could replace this with just glob.glob('sparser*.log') for simplicity.

indra/tools/reading/submit_reading_pipeline_aws.py
```diff
@@ -86,6 +94,17 @@ def check_logs(job_defs):
...
 
+    # Don't start watching jobs added after this command was initialized.
+    observed_job_def_set = set()
+
+    if stash_log_method is not None:
+        def update_observed_jobs(job_defs):
+            for job_def in job_defs:
+                observed_job_def_set.add(
+                    tuple([(k, v) for k, v in job_def.items()
+                           if k in ['jobName', 'jobId']])
```
bgyori: I'm trying to wrap my head around this, it might be clearer to do 
```
# Build up a set of tuples representing job names and ids for each job to add
jobs_to_add = {(('jobName', job_def['jobName']), 
                ('jobId', job_def['jobId'])) for job_def in job_defs}
# Update the observed jobs
observed_job_def_set |= jobs_to_add
```

bgyori:
Great! The only question that occurs to me is whether it is okay to use 'w' in several places to open files for logs, or whether the output mode and encoding should be made explicit (for instance open as 'wb' and encode the log as encode('utf8')). It's hard to see if there's a possibility for unicode, maybe not?

pagreene:
It actually depends on the location a bit. I'll take a look. I think that
all the files should in general be bytes.

indra/tools/reading/submit_reading_pipeline_aws.py
pagreene: That is so that if a job meeting the criteria comes up after `wait_for_complete` has started. The idea is that anything the script "waits for" gets their logs stashed.

indra/tools/reading/read_db_aws.py
pagreene: Oh, true. I almost never use glob. Is it a standard part of Python, or does it need special installation?

PR #384, opened 2018-01-24
Belief rule subtype priors

djmilstein:
Allows the belief engine to have random noise priors for specific statement subtypes (such as biopax databases and reach rules), falling back on the overall source random noise priors when not specified.


data/reach_rule_substrings.p
```diff
@@ -0,0 +1,364 @@
...
+p177
+aS'hydrolysis_token_2'
+p178
+aS'hydrolysis_decl1'
+p179
+aS'hydrolysis_pass1'
+p180
+aS'hydrolysis_subjnom1'
+p181
+a.
```
bgyori: I think this file should live in indra/sources/reach, and could just be a raw .txt file since it just contains a list of strings.

indra/belief/evidence_subtype_tagger.py
```diff
@@ -0,0 +1,76 @@
...
+
+    best_match_length = None
+    best_match = None
+    for ss in reach_rule_regexp:
+        if re.search(ss, event_name):
+            if best_match is None or len(ss) > best_match_length:
+                best_match = ss
+                best_match_length = len(ss)
+
+    return best_match
```
bgyori: I would actually move this function into indra.sources.reach.processor, and change the API to take the event_name as an argument. The Belief Engine could then just import this function and call it with evidence.annotations['found_by'] to get the subtype.

indra/belief/evidence_subtype_tagger.py
```diff
@@ -0,0 +1,76 @@
...
+    if source_api == 'biopax':
+        subtype = annotations['source_sub_id']
+    elif source_api == 'reach':
+        subtype = determine_reach_subtype(evidence)
+    elif source_api == 'geneways':
+        subtype = annotations['actiontype']
+    else:
+        subtype = None
+
+    return (source_api, subtype)
```
bgyori: I think it isn't necessary to keep this function in a separate file, I think it should be simply moved into indra/belief/__init__.py

djmilstein:
I agree with these suggestions! I just pushed commits that implement them
and won't make further changes until requested.

PR #386, opened 2018-01-28
Update printing of activities

### LINKED ISSUE #352: Update pretty printing of activitites
`Phosphorylaion(X(kinase: true)) ...)` can just say `Phosphorylaion(X(kinase)) ...)` unless it's explicitly false. I'll make a PR for this

cthoyt:
This now lists the activities more succinctly. They will be implicitly "True", and only show "False" when that is set.

Closes #352

Side note - I was curious about why += was being used with lists, rather than either using list.append or list.extend in this function

### DIFF
--- a/indra/statements.py
+++ b/indra/statements.py
@@ -801,8 +801,11 @@ class Agent(object):
             mod_str += ', '.join(['%s' % m for m in self.mods])
             attr_strs.append(mod_str)
         if self.activity:
-            attr_strs.append('%s: %s' % (self.activity.activity_type,
-                                         self.activity.is_active))
+            if self.activity.is_active:
+                 attr_strs.append('%s' % self.activity.activity_type)
+            else:
+                 attr_strs.append('%s: %s' % (self.activity.activity_type,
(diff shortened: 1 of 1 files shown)


bgyori:
Thanks and sorry for the slow merge. I tend to use append when adding a single element to the list, but almost always use += for list extension with another list. I think it's just a matter of style since there isn't a considerable difference between the two that I'm aware of, or is there?

cthoyt:
I'm not sure if += causes the creation of a new list. If it does, then extend would be much preferred.

I think in the code I saw a couple lines written like `x += [y]` so I was a bit unsure. Maybe changing those to append would save some processing power, and the trees!

bgyori:
No, there is no such difference. And I'm not aware of x += [y] in the code where y is a single element, I think you might be mistaking it for cases where [...] is a list comprehension.

bgyori:
Except this line :) 
```
attr_strs += ['location: %s' % self.location]
```
which I think is just a mistake.

PR #389, opened 2018-02-02
Added CAG assembler

adarshp:


bgyori:
In the commits above I changed the following things:
- removed making the graph from the constructor, and renamed the function to make and return the graph to `make_model` (the standard for other assemblers)
- made some style changes to satisfy PEP8
- changed to networkx 1.11 compatibility from networkx 2 (INDRA and some of its dependencies are currently only compatible with networkx 1.11, with 2 not being backwards compatible)
- fixed an issue related to unicode characters in Python 2
Tests now seem to be passing. Does this look okay, @adarshp?

A general question I have is, why are nodes for derivatives added? They don't seem to be connected by edges to any of the other nodes.

adarshp:
Oversight on my part, the derivative nodes should not be there (I use the derivatives in the DBN). I can change it later tonight.

--
Adarsh Pyarelal
School of Information 
University of Arizona
Office: Harvill 454
http://adarsh.cc

adarshp:
Ok, I have taken out the partial derivatives. The other changes you made look good to me too. Thanks @bgyori !

adarshp:
Ok, I think I fixed the bugs. I also changed how the Eidos processor handles missing adjective and polarity information. How does it look now?

bgyori:
Thanks! Since it's only the Agents from the Statements that are added as nodes, I simplified the implementation. This is now very similar to some of the other graph-like assemblers, but perhaps it is worth keeping as a separate assembler to have increased flexibility in how metadata and styling is set for the specific use case.

adarshp:
Sounds good, thanks for making the changes and merging!

--
Adarsh Pyarelal
School of Information 
University of Arizona
Office: Harvill 454
http://adarsh.cc

Issue #396, opened 2018-02-20
Eidos update breaks INDRA-Eidos API


adarshp:
Recent updates to Eidos have broken the functioning of the INDRA-Eidos API.

From looking at the changes in the codebase, the main culprit is the change in the type signature of the extractFrom function in EidosSystem.scala - it used to be String -> Vector[Mention], but is now String -> AnnotatedDocument (which shows up in org/clulab/wm/serialization/json/JLDSerializer.scala). Any idea how to fix this?

bgyori:
Thanks, will work on this tomorrow!

adarshp:
I think I fixed it - just submitted a PR.

Issue #394, opened 2018-02-16
Find a way to use Reach/Eidos jars simultaneously


bgyori:
Currently the fat jars for Reach and Eidos can't simultaneously be on the CLASSPATH because they pack conflicting dependencies that lead to errors. This has prevented me from adding tests for Eidos without breaking Travis builds. A solution would be to introduce REACHPATH and EIDOSPATH variables that are added to the CLASSPATH from within the respective APIs to make sure the right one is on the CLASSPATH at the right time.

jmuhlich:
I think you can use separate explicit class loaders in Java to manage this, rather than using the default class loader.

https://stackoverflow.com/a/6108133
https://en.wikipedia.org/wiki/Java_Classloader#JAR_hell

bgyori:
That stackoverflow suggestion is good but that piece of code would have to be integrated into pyjnius to work (since pyjnius is the one that actually ends up loading a class you specify). I'll have to dig into it a bit more.

bgyori:
The issue is that as soon as `autoclass` is imported, the Java VM is started (which reads the CLASSPATH in whatever state it is in when `autoclass` loads) and it's not possible to effectively change the CLASSPATH afterwards to use JARs that weren't initially on it. This is also discussed in https://github.com/kivy/pyjnius/issues/316, and there doesn't seem to be a solution yet.

ctrueden:
> it's not possible to effectively change the CLASSPATH afterwards to use JARs that weren't initially on it.

If all you want is to add JARs to the classpath of a running Java, apparently that is possible; see https://github.com/kivy/pyjnius/issues/434#issuecomment-515693635. Not sure if changes would be needed on the pyjnius side, though.

bgyori:
Thanks, @ctrueden, I guess the issue is that I would need to both add _and remove_ paths from the CLASSPATH during runtime (see https://github.com/kivy/pyjnius/issues/411#issuecomment-540463639) to solve this issue.

ctrueden:
@bgyori Ah, yeah, in that case you'd need to use custom class loaders on the Java side. And you can't wrap those dynamically loaded classes on the Python side (see kivy/pyjnius#316); if you try you'll see something like:
```
Traceback (most recent call last):
  File "jnius/jnius_export_class.pxi", line 766, in jnius.JavaMethod.__call__
  File "jnius/jnius_export_class.pxi", line 845, in jnius.JavaMethod.call_method
[...]

PR #397, opened 2018-02-20
Update eidos_reader.py for new Eidos version

adarshp:
I updated the eidos_reader module to be compatible with the new Eidos version with the new type signature for the extractFrom function.


bgyori:
Thanks! I also pushed an update to the package name in which EidosSystem lives, which has just changed yesterday. This fixes #396 .
I have an idea how to solve #394 so that we can run Eidos tests on Travis and then we will know if integration ever breaks - will work on this later today.

PR #404, opened 2018-02-25
Update INDRA to work with networkx 2.1

bgyori:
This PR updates various submodules of INDRA to work with networkx version 2.1.


bgyori:
It looks like the test are passing but there's an issue https://github.com/pybel/pybel/issues/277 (probably unrelated) that results in an error in Python 2.

PR #405, opened 2018-02-26
REACH and Eidos together

### LINKED ISSUE #359: Add tests for Eidos processing and Influence Statements
These are not tested yet.

### LINKED ISSUE #394: Find a way to use Reach/Eidos jars simultaneously

bgyori:
This PR adds some CLASSPATH handling to allow more conveniently using REACH and Eidos and allow testing both on Travis. REACHPATH and EIDOSPATH point to the JAR file of each system. These two paths can either be on the CLASSPATH set by the user, or, if not on the CLASSPATH, the right one is added to the CLASSPATH in the respective reach_reader.py or eidos_reader.py.

One change is that REACHPATH now points to the REACH JAR rather than the folder that the REACH JAR is in, this is updated in the various readers. A simultaneous update in indra_deps_docker is ready to be pushed if this is merged.

This fixes #394 to the extent that both readers (and Biopax) can be used with a single fixed setting of environment variables, but due to fundamental limitations in jnius, using both readers in the same session is still not possible (i.e. `from indra.sources import reach, eidos` and then using both wouldn't work). It also fixes #359.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,9 @@ install:
   - tar -xf trips_reach_batch4.gz
   - cd $TRAVIS_BUILD_DIR
   - wget http://sorger.med.harvard.edu/data/bachman/reach-82631d-biores-e9ee36.jar -nv
-  - export REACHPATH=$TRAVIS_BUILD_DIR
+  - export REACHPATH=$TRAVIS_BUILD_DIR/reach-82631d-biores-e9ee36.jar
+  - wget http://sorger.med.harvard.edu/data/bgyori/eidos-assembly-0.1-SNAPSHOT.jar -nv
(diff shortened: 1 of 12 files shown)


bgyori:
Despite these changes it doesn't seem possible to test both REACH and Eidos with a single `nosetests` call, even with the `--with-isolation` flag. Therefore I ended up separating out Eidos tests from the other tests.

PR #411, opened 2018-03-01
Add eidos rule used for extraction to evidence annotations

djmilstein:


bgyori:
@djmilstein thanks, please add a test for this to make sure this keeps working as Eidos is developed.

PR #412, opened 2018-03-01
Create a flask app for accessing the database content

pagreene:
This pull request add a python app which allows for simple access to the raw statements produced by the reading pipeline and placed in the database.


bgyori:
Looks good!

Issue #417, opened 2018-03-10
Add ability to parse Eidos' JSON-LD format


adarshp:
Is this functionality something that is already being worked on? If not, I could try to implement it.

bgyori:
Fixed in #418

PR #418, opened 2018-03-11
Json LD parsing functionality

adarshp:
Added JSON-LD parsing functionality

### DIFF
--- a/indra/sources/eidos/__init__.py
+++ b/indra/sources/eidos/__init__.py
@@ -1,2 +1,3 @@
-from .eidos_api import process_text, process_json_str, process_json_file, \
-                       process_json_ld_file, process_json
+from .eidos_api import process_text, \
+    process_json_str, process_json_file, process_json, \
+    process_json_ld_str, process_json_ld_file, process_json_ld
--- a/indra/sources/eidos/__init__.py
+++ b/indra/sources/eidos/__init__.py
@@ -1,2 +1,2 @@
 from .eidos_api import process_text, process_json_str, process_json_file, \
-                       process_json
+                       process_json_ld_file, process_json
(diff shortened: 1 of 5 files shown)


adarshp:
#417

bgyori:
This looks great, thanks @adarshp! Since the processing code of JSON and JSON-LD dicts is distinct, I'm wondering if we should make two separate classes inside process_eidos.py, i.e. EidosJsonProcessor and EidosJsonLdProcessor which can then each contain the relevant methods for the two formats - I can do this if you agree. Another nice addition would be to add another option to the EidosReader to call EidosSystem to produce JSON-LD output (rather than its current default of the simple JSON of OdinMentions) - again, I could add this pretty easily but you can also work on it if you want. Let me know!

adarshp:
1. I think two separate classes sounds fine. An alternative that might keep things more unified would be to split the logic by checking the extension of the input file - if it is .json, it could be processed as a JSON, and if it is .jsonld, it could be processed as JSON-LD. However, I'm not sure if this might be too brittle an approach or not.

2. I think it would definitely be good to add the option to have EidosSystem produce JSON-LD output.

I'll take you up on your offer to implement these features, thanks!

bgyori:
Alright, I added the following things:
- separated the two processors, both in processor.py
- implemented reading into a JSON-LD string via the EidosReader (this will work once https://github.com/clulab/eidos/pull/177 is merged)
- added the extraction of Evidence from JSON-LD - currently evidence text and the rule used to get the extraction are added to the evidence of each Statement
- added a test

There is now an optional argument to `process_text` called `out_format` which defaults to the "old" `json` but can also be set to `json_ld`, for instance
```
from indra.sources import eidos
ep = eidos.process_text('fuel prices decrease water trucking', out_format='json_ld')
```

PR #419, opened 2018-03-12
CAG Assembler improvement and bug fix

adarshp:
Fixed bug in CAG assembler where the word "simulable" was being treated as a node, some refactoring, some additional metadata for cytoscape JS viz.


bgyori:
I also added a test for the CyJS export

PR #420, opened 2018-03-12
This pull request adds the option to the Eidos JSON-LD processor to set a minimum grounding threshold for merging entities.

adarshp:


bgyori:
Hey @adarshp, this is an interesting addition but I'm not convinced (yet) that it is the best way to handle the issue of grounding. I would prefer an approach where we represent the output of reading (including various groundings) as closely as possible to the original output, and then apply things like grounding normalization and thresholds in a separate step - ideally at an assembly stage. In any case, I think this is a more conceptual issue and so I would appreciate a chance to discuss before adding this change.

adarshp:
That sounds good to me. Basically, what I would like to do is to store the grounding information from the Eidos JSON-LD output to ground the Agents in the Influence statements. Two possible alternative routes come to mind:

1. Adding a ‘groundings’ slot to the Agent class.
2. Adding ‘subj_groundings’ and ‘obj_groundings’ slots to the ‘Influence’ class.

Do either of these sound like what you had in mind? If not, let’s discuss further tomorrow.

--
Adarsh Pyarelal
School of Information 
University of Arizona
Office: Harvill 454
http://adarsh.cc

bgyori:
Yes, Agents have a `db_refs` dictionary which is meant to contain mappings to databases/ontologies. It is always used when dealing with sources in the biology domain and has been adequate for those use cases. We'll need to discuss whether db_refs as it currently stands, is adequate for representing grounding from Eidos / BBN / other inputs or whether it needs to be generalized (I suspect it does, but not sure yet, how, exactly).

bgyori:
Closing this, superseded by #423 .

PR #423, opened 2018-03-13
Integrate Eidos grounding and updates to CagAssembler

adarshp:
Moved grounding information to db_refs, modified CAG assembler to include edge colors and tooltips for provenance and modfiers, some refactoring

### DIFF
--- a/indra/assemblers/cag_assembler.py
+++ b/indra/assemblers/cag_assembler.py
@@ -4,6 +4,7 @@ from builtins import object, dict, str
 import logging
 import networkx as nx
 from indra.statements import Influence
+import json
 
 # Python 2
 try:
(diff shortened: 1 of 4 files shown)


bgyori:
Thanks! Yes, I just reviewed in more detail how grounding is represented in the Eidos JSON-LD, and I think this approach to propagating the information in INDRA is perfect. It then makes sense to choose a strategy for preferred grounding and name standardization in the output assembler (in this case the CagAssembler). Though this doesn't happen yet, I imagine the scored grounding matches will be part of the redundancy resolution logic during internal knowledge assembly, and will need to be handled in the Preassembler - this will partially relieve the output assembler of having to make these choices.

bgyori:
Alright, this looks good! I added some more tests and made some small changes mainly to handle missing information (i.e. grounding or provenance missing). I noticed that the way the EidosReader's JSON-LD reading interface is currently implemented, it's not outputting any groundings - that will have to be fixed in a separate branch but it's good to be aware of this limitation for now.

Issue #176, opened 2017-02-24
Problem with stmts_to_json


bgyori:
Only the uuid of supports/supported_by statements is serialized. But if those statements referred to are not in the list of statements then they will be lost.

bgyori:
Addressed in #428

PR #428, opened 2018-03-14
Fix stmts from json bug

pagreene:
This fixes a couple of bugs surrounding serialization of indra statement:
- When converting pre-assembled statements from json, uuids that can't be resolved are handled in 3 different ways, controlled by a keyword argument (see documentation for details).
- When converting any statement from json, the copy lists rather than pass references from the json, preventing complicated cross-contamination of the input json when modifying the resulting statements.
- Implement test to check these features.

### DIFF
--- a/indra/statements.py
+++ b/indra/statements.py
@@ -2615,6 +2615,12 @@ def stmts_from_json(json_in, on_missing_support='handle'):
     return stmts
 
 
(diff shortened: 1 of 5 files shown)


bgyori:
Great, I believe this is also an adequate fix for #176.

PR #429, opened 2018-03-15
Added null handling functionality to Eidos processor to deal with cases when no events are found in the document

adarshp:


bgyori:
Thanks! I simplified the return to be done up front, and also realized that pulling out the function definitions up front is better than defining (and many times redefining) functions inside the loop so I pulled those out. I think the code looks cleaner now.

adarshp:
I think those are good changes, thanks for implementing them!

--
Adarsh Pyarelal
School of Information 
University of Arizona
Office: Harvill 454
http://adarsh.cc

PR #432, opened 2018-03-16
Check for NDEx credentials in the environment

cthoyt:
I would like to post some configuration files for my INDRA machines publicly, but without the username and password for my NDEx account in it. One way to specify these is through the environment. This commit updates so the code will only check the environment if the user and password are not explicitly set in the NDEx credentials in the config.yaml file for a given INDRA machine.

References cthoyt/indra-machines#2


bgyori:
This looks great, thanks! At some point we are planning to collect all the various settings that INDRA uses and make those configurable in a single file.

PR #434, opened 2018-03-19
Assembly engineered for the database

pagreene:
Takes initial steps towards an assembly pipeline for the database.
- Implements methods for creating preassembly results for the database
- Prototypes the data structures that will be used in the database, and implements required changes in indra statements and preassembly.
- Refactors some of the code in statements, preassembly, grounding mapper, and site mapper.
- In general changes the way we populate agents tables, so that agents are added for all statements that lack them.


pagreene:
I was originally going to drop out commig 7457c7c because I couldn't actually add a meaningful test. The changes aren't necessary, but they do tidy it. Let me know if you want me to blow it away.

PR #435, opened 2018-03-19
Adding text to CAG edge tooltips in viz

adarshp:


indra/assemblers/cag_assembler.py
```diff
@@ -82,16 +82,22 @@ def make_model(self, grounding_threshold = None):
...
+                targetArrowShape, linecolor = ('triangle', 'maroon')
+
 
             # Add edge to the graph with metadata from statement
             if s.evidence:
                 provenance = s.evidence[0].annotations.get('provenance', {})
             else:
-                provenance = {}
+                provenance = [{}]
+            provenance[0]['text'] = s.evidence[0].text
```
bgyori: This line seems to assume that there is an evidence list and that the first element of that has a text entry - which is almost certainly the case except if there's a bug somewhere. But then the previous if-else block could also assume that s.evidence has content without checking.

adarshp:
You are right - I've pushed a commit that removes that if-else block.

bgyori:
Great, thanks! In general, a Statement can have multiple pieces of evidence (i.e. len(stmt.evidence) > 0) with different (1) `source_api` (eidos, bbn, etc.) (2) `text` and (3) `pmid` (more generally a document ID, the name will soon be generalized). This step of aggregating evidence happens when one runs INDRA's Preassembler on a set of Statements and duplicates are discovered.
If tooltips can support multiple pieces of evidence, you could take the whole evidence list with the relevant attributes and add all of them to the tooltip.

adarshp:
In theory, the tooltips can display any text - at some point I'll try to get them to work well when nodes/edges are merged...

PR #436, opened 2018-03-20
Added a to_json method for Influence statements

adarshp:


bgyori:
Thanks, I don't think this changes what already exists - these serialization steps are already inherited by Influence from its parent classes. What _is_ missing is adding the subj_delta and obj_delta attributes to the serialization which I'm happy to add in another commit.

adarshp:
True, but in its current form it is outputting a field corresponding to the Systems Biology Ontology (SBO), that I felt should not be there. Of course, I can do `del['sbo']` or something like that in my own code if you prefer.

bgyori:
I see, I will remove that

adarshp:
Thanks!

bgyori:
Alright, I added serialization/deserialization with all the attributes for Influences, and without the SBO annotations. I'll let the tests run and then merge this if everything passes.

PR #443, opened 2018-03-22
Update REACH Output

cthoyt:
First, this merge has a bit of code cleanup for the INDRA machine. More importantly, it adds an extra kwarg to the processing functions in the REACH API  (`indra.sources.reach.reach_api`) that allows for the directory where output gets stuck to be specified. 

Before, it just stuck the `reach_output.json` in whatever folder was the current working directory. This only doesn't cause a problem if you run the INDRA machine from in the same directory as where the `config.yaml` file and `jsons` folder are. Now, it can be run from an arbirary point on the file system, and still know where to put `reach_output.json`.

### DIFF
--- a/indra/sources/reach/reach_api.py
+++ b/indra/sources/reach/reach_api.py
@@ -34,10 +34,10 @@ except Exception:
 
 reach_text_url = 'http://agathon.sista.arizona.edu:8080/odinweb/api/text'
 reach_nxml_url = 'http://agathon.sista.arizona.edu:8080/odinweb/api/nxml'
-reach_output_fname = 'reach_output.json'
+default_output_fname = 'reach_output.json'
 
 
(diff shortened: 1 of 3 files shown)


bgyori:
Thanks! I like this but I think it would be even better to parameterize the specific output json file that is created rather than the directory (so instead of the current `directory` argument, we could have `output_fname` or something like that). We have a similar argument in some other input interfaces, just never implemented it for REACH. If you want to add that it's great but I'm happy to do it too.

cthoyt:
Agreed, that makes a lot of sense. If you want this today, go ahead and make the change yourself. Otherwise I'll take care of it tomorrow :)

bgyori:
I made the changes -- hope this will work with your setup!

PR #448, opened 2018-03-26
Trips cwms

djmilstein:


bgyori:
Very nice, thanks @djmilstein and @pagreene !

PR #449, opened 2018-03-26
Biogrid from file

djmilstein:
Written by @bgyori: This PR adds BioGRID as a source in `indra.sources.biogrid`. Though BioGRID is part of PathwayCommons, the current PC release is buggy, and also, having a separate client to BioGRID has advantages. The API allows passing the path to a local BioGRID tabular file, or if not specified, will download a release from the BioGRID website and process it.


bgyori:
I think the test is a little bit too heavy as implemented - we don't necessarily want to download the file and process the whole thing each and every time in tests. How about taking a small sample of the overall file adding it to this repo and using that for testing?

djmilstein:
Yes, sir!

bgyori:
Great, thanks. One more thing: since this is a new knowledge source, we should have an entry for it in the documentation, so please add an entry in doc/modules/sources/index.rst and make a folder called doc/modules/sources/biogrid and add an index.rst with the necessary sections.

djmilstein:
Thank you! I have done so.

bgyori:
Looks like the commit that changes the tests to use a local smaller file got lost somehow, not sure what happened, @djmilstein can you reinstate it, perhaps from a local branch?

djmilstein:
Yep yep. Just did so!

On Wed, Mar 28, 2018 at 6:02 PM, Benjamin M. Gyori <notifications@github.com

PR #452, opened 2018-03-27
Added functionality to deal with the case when Eidos groundings are not present, and when an entity is not truly grounded (Eidos just presents a list of ontology concepts with grounding score = 0.0)

adarshp:


bgyori:
Thanks! I didn't notice that Eidos sometimes returns groundings with 0.0 scores. In that special case, I think it might make sense to not propagate those groundings into Statements at all, i.e. ignore them in the input processor itself, because they aren't meaningful matches. Let me add this and then merge.

PR #457, opened 2018-03-28
Signor complexes families

djmilstein:


indra/sources/signor.py
```diff
@@ -22,9 +22,27 @@
...
+    fname = os.path.join(dirname(__file__), '../resources/famplex_map.tsv')
+    raw_map = read_unicode_csv(fname, '\t')
+
+    m = {}
+    for row in raw_map:
+        m[(row[0], row[1])] = row[2]
+    return m
+famplex_map = _read_famplex_map()
+
+def _read_signor_complex_map():
```
johnbachman: do you think that you could set this up to pull the latest set of complexes from the SIGNOR website, rather than a file added to /resources? The SIGNOR processor could make a post request to the SIGNOR website for this data (see how the interaction data is downloaded).

indra/sources/signor.py
```diff
@@ -194,24 +216,61 @@ def __init__(self, signor_csv=None, delimiter='\t'):
...
+            agents = []
+            for c in components:
+                db_refs = {}
+
+                name = uniprot_client.get_gene_name(c)
+                if not name:
+                    print('Could not look up', c, '(maybe the signor ' + \
+                            'complex component is not a UNIPROT id)')
+                    db_refs['SIGNOR'] = id
+                if name:
```
johnbachman: Make this an else?

indra/sources/signor.py
```diff
@@ -194,24 +216,61 @@ def __init__(self, signor_csv=None, delimiter='\t'):
 
     @staticmethod
     def _get_agent(ent_name, ent_type, id, database):
```
johnbachman: This function doesn't need to return a list of Agents for complexes--rather, it should return a single agent with the other agents incorporated as BoundConditions. For simplicity, it's fine to just use the first protein listed in the complex as the "main" agent, the rest as BoundConditions.

indra/sources/signor.py
```diff
@@ -240,134 +299,142 @@ def _get_evidence(row):
...
-        # the appropriate active form.
         no_mech = False
-        if mech_stmt_type and issubclass(mech_stmt_type, Modification):
-            if not row.RESIDUE:
-                # Modification
-                mod_stmt = mech_stmt_type(agent_a, agent_b, None, None,
-                                          evidence=evidence)
-                stmts.append(mod_stmt)
+
+        for combo in combinations:
```
johnbachman: It's not necessary to generate statements for every possible combination of subject/object members. Instead, can just use the single agents with BoundConditions.

indra/sources/signor.py
```diff
@@ -194,24 +216,61 @@ def __init__(self, signor_csv=None, delimiter='\t'):
...
-        # Other possible groundings are PUBCHEM and SIGNOR
-        elif gnd_type is not None:
-            assert database in ('PUBCHEM', 'SIGNOR')
-            db_refs = {database: id}
-            name = ent_name
-        # If no grounding, include as an untyped/ungrounded node
+        # Returns a list of agents corresponding to this id
+        # (If it is a signor complex, returns Agent objects corresponding to
+        # each complex constituent)
+        if database == 'SIGNOR' and id in complex_map:
```
johnbachman: One additional thing to do for agents involving complexes is to (perhaps in a separate step) create a Complex Statement involving all of the members.

PR #461, opened 2018-03-30
Process bound conditions

djmilstein:
Process bound conditions in the grounding mapper, site mapper, and various filtering functions in assemble_corpus.py.


indra/preassembler/grounding_mapper.py
```diff
@@ -101,6 +101,13 @@ def map_agents_for_stmt(self, stmt, do_rename=True):
         mapped_stmt = deepcopy(stmt)
         # Iterate over the agents
         mapped_agent_list = mapped_stmt.agent_list()
+
+        # Add any agents in the bound condition to the agent list
+        for agent in mapped_agent_list:
+            if agent is not None:
+                bc_agents = [bc.agent for bc in agent.bound_conditions]
+                mapped_agent_list.extend(bc_agents)
+
```
bgyori: This way of constructing an agent list to map works well in most cases but can cause error when trying to call set_agent_list (see https://github.com/sorgerlab/indra/pull/461/files#diff-51eacc155ba7d90175bdc81b1043716aR120). This happens when using the so called "agent map" which takes text strings and maps them to stateful Agents from a json file. I added a test condition which triggers this case.

PR #462, opened 2018-03-30
Concepts, alternative hierarchies and the Eidos ontology

bgyori:
This PR:
- introduces the Concept class, a base class of Agent without the molecule-specific attributes (mods, mutations, activity, location)
- updates the Eidos, BBN and CWMS processors to produce Influence statements over Concepts rather than Agents
- builds up an RDF graph of the Eidos ontology such that it is compatible with the HierarchyManager and shows that Statements can be preassembled correctly with respect to this ontology
- generalizes some assemblers to work with the more general Concepts and only look at Agent-specific attributes when appropriate


bgyori:
I'm actually pleasantly surprised that this generalization could be done with just 3 lines added to the HierarchyManger and no changes at all to the Preassembler.

PR #463, opened 2018-03-31
Optional grounding map in assemble corpus and machine

### LINKED ISSUE #454: Add enhanced grounding mappers
I was just running my own INDRA machine for APP-related research and when outputting with the PyBEL Assembler, got logging like this:

```
...
WARNING: [2018-03-28 12:54:32] indra/pybel_assembler - Agent fibrillary_Abeta() has no grounding.
...
WARNING: [2018-03-28 12:54:32] indra/pybel_assembler - Agent scavenger_receptors(activity) has no grounding.
...
WARNING: [2018-03-28 12:54:32] indra/pybel_assembler - Agent amyloid_beta() has no grounding.
...
WARNING: [2018-03-28 12:54:32] indra/pybel_assembler - Agent RAGE_receptors(activity) has no grounding.
...
WARNING: [2018-03-28 12:54:32] indra/pybel_assembler - Agent tau_proteins() has no grounding.
...
WARNING: [2018-03-28 12:54:32] indra/pybel_assembler - Agent APOE_epsilon4() has no grounding.
...
```

For the entities I've highlighted, I think I can generate mappings to elements in BEL. For example, amyloid beta can be mapped to a ChEBI term, or a bel term like `p(HGNC:APP, frag(673_711))` (I probably got the numbers wrong, but you can see the idea)

In general, a domain expert might be able to provide mappings. Might we develop a way to load a mapping file to help INDRA with cases like this?

bgyori:
This PR makes it possible to parameterize machines with an optional grounding map. This is now also possible for IncrementalModel and assemble_corpus.

Fixes #454

### DIFF
--- a/indra/tests/test_assemble_corpus.py
+++ b/indra/tests/test_assemble_corpus.py
@@ -170,6 +170,19 @@ def test_map_grounding():
     assert(st_out[0].obj.db_refs.get('FPLX'))
     assert(st_out[0].obj.name == 'ERK')
 
+
+def test_map_grounding_user_map():
+    gm = {'MEK': {'XXX': 'YYY'}, 'ERK': {'FPLX': 'ERK'}}
+    a = Agent('MEK', db_refs={'TEXT': 'MEK'})
+    b = Agent('X', db_refs={'TEXT': 'ERK'})
+    st = Activation(a, b)
+    st_out = ac.map_grounding([st], grounding_map=gm, do_rename=True)
+    assert(len(st_out) == 1)
+    assert(st_out[0].subj.db_refs.get('XXX') == 'YYY')
+    assert(st_out[0].obj.db_refs.get('FPLX') == 'ERK')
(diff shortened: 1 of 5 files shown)


cthoyt:
Would it make sense to allow multiple grounding maps? For example, a user might want to use both the default INDRA grounding map plus add their own (without having to fork/modify FamPlex)

bgyori:
Yes, and this allows you to do that. You can create your own grounding map file on any path, load it, and use it in assembly. Or you can load INDRA's grounding map and override entries in it or extend it, e.g. by doing
```
from indra.preassembler.grounding_mapper import default_grounding_map as gm
my_grounding_map = {'X': {...}, 'Y': {...}}
gm.update(my_grounding_map)
```
and then using `gm` in subsequent assembly tasks.

bgyori:
Specifically for the machine, you would want to save the default grounding map updated with a custom map back into a TSV file so it can be loaded using the path specified in the config.

PR #464, opened 2018-04-02
INDRA config file

djmilstein:
A single configuration file has API keys, dependency paths, database locations, credentials, and other settings.


indra/resources/default_config.json
```diff
@@ -0,0 +1,26 @@
...
+    "BIOGRID_API_KEY": "",
+    // Credendials for ndexbio.org
+    "NDEX_USERNAME": "",
+    "NDEX_PASSWORD": "",
+    // Addresses to INDRA database instances
+    "primary": "",
+    "test-1": "",
+    "test-2": "sqlite:///indra_test.db"
+}
+
```
bgyori: I just added this to delete actual values from the config file, and to make the point that we may want a format in which comments are allowed (JSON doesn't actually allow commenting). Perhaps configparser (https://docs.python.org/3/library/configparser.html) or YAML could work. Also, the keys for the database-related settings are too generic this way (primary, test-1, test-2), they could have more descriptive names.

PR #466, opened 2018-04-02
Added sanitizing of eidos evidence sentence text - basically, unescaped parentheses

adarshp:


bgyori:
Thanks, this is good. Not 100% related but I realized I accidentally introduced a bug in PR #463 where I flipped the order of grounding tuples from Eidos. I'll push a fix for that after merging this.

PR #478, opened 2018-04-04
Added a guard to deal with missing evidence (like for manually constructed Influence statements)

adarshp:


bgyori:
Thanks, I fixed a bug with provenance not being declared if there's no evidence, and added a regression test to make sure this will keep working.

PR #482, opened 2018-04-05
Handle multiple sites in site text, broader handling of mutation patterns

djmilstein:
Note: we may or may not be handling multiple site patterns in the desirable way, see details below.

This PR:
1. Handles mutation patterns with residues described with more than one letter (such as Leu174Arg)

2. Handles site patterns with multiple sites, separated by either a comma or a slash (such as S199/S202/T205 or 992,1068). If the site text describes a modification to an agent, adds a ModCondition to agent.mods for each entry in the comma or slash separated site list. If the site text modifies a modification statement, creates one modification statement per site list entry (is this the desired behavior?)

### DIFF
--- a/indra/tests/test_reach.py
+++ b/indra/tests/test_reach.py
@@ -16,35 +16,45 @@ def test_parse_site_text():
             'threonine residue 185', 'T185']
     assert unicode_strs(text)
     for t in text:
-        residue, site = ReachProcessor._parse_site_text(t)
+        sites = ReachProcessor._parse_site_text(t)
+        assert(len(sites) == 1)
+        residue, site = sites[0]
         assert(residue == 'T')
         assert(site == '185')
         assert unicode_strs((residue, site))
 
 def test_parse_site_text_number():
     t = '135'
-    residue, site = ReachProcessor._parse_site_text(t)
+    sites = ReachProcessor._parse_site_text(t)
+    assert(len(sites) == 1)
+    residue, site = sites[0]
     assert(residue is None)
     assert(site == '135')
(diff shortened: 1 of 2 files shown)


indra/tests/test_reach.py
```diff
@@ -53,11 +63,52 @@ def test_parse_site_residue_only():
...
+    assert(sites[0][0] == 'Y')
+    assert(sites[0][1] == '1221')
+    assert(sites[1][0] == None)
+    assert(sites[1][1] == '1222')
+
+    sites = ReachProcessor._parse_site_text('Tyr-577/576')
+    assert(len(sites) == 2)
+    assert(sites[0][0] == 'Y')
+    assert(sites[0][1] == '577')
+    assert(sites[1][0] == None)
```
bgyori: Could you actually make this 'Y' instead of None? The pattern implies that both of those residues are tyrosine (though possibly on different isoforms of the protein agent).

indra/tests/test_reach.py
```diff
@@ -53,11 +63,52 @@ def test_parse_site_residue_only():
...
+    assert(sites[0][0] == None)
+    assert(sites[0][1] == '992')
+    assert(sites[1][0] == None)
+    assert(sites[1][1] == '1068')
+
+    sites = ReachProcessor._parse_site_text('Y1221/1222')
+    assert(len(sites) == 2)
+    assert(sites[0][0] == 'Y')
+    assert(sites[0][1] == '1221')
+    assert(sites[1][0] == None)
```
bgyori: This should also be 'Y' instead of None

PR #487, opened 2018-04-10
Added functionality to display CAG in Jupyter notebook cell

adarshp:


bgyori:
Thanks! I refactored this quite a bit:
- separated out the JavaScript template into a separate file
- separated out the default style/layout into a separate json
- used a more primitive way to fill the template string so that we don't need to add Jinja as a dependency (which would require updating setup, documentation, tests, etc.)
I also updated my previous demo Jupyter notebook and added a section for this interactive visualization.

bgyori:
The failing test is unrelated so I'm merging this

adarshp:
All good ideas, thanks!

PR #488, opened 2018-04-10
Added null handling for missing Eidos @type key

adarshp:


bgyori:
Closing this as it's included in #487

PR #491, opened 2018-04-12
Add a tool for getting the ideal set of statements for preassembly

pagreene:
This PR adds the functionality in `db_util` to get a corpus of statement from reading which eliminates exact duplicates and selects from the best versions of content for each ref. in the function `get_reduced_stmt_corpus`.

### DIFF
--- a/indra/db/util.py
+++ b/indra/db/util.py
@@ -563,8 +563,10 @@ class NestedDict(object):
         return result_list
 
 
-def get_reduced_stmt_corpus(db, get_full_stmts=False):
+def distill_stmts_from_reading(db, get_full_stmts=False, clauses=None):
     """Get a corpus of statements from clauses and filters duplicate evidence.
+
+    Note that this will only get statements from reading.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


indra/db/util.py
```diff
@@ -472,6 +471,241 @@ def make_stmts_from_db_list(db_stmt_objs):
...
+    ----------
+    db : :py:class:`DatabaseManager`
+        A database manager instance to access the database.
+    get_full_stmts : bool
+        By default (False), only Statement ids (the primary index of Statements
+        on the database) are returned. However, if set to True, serialized
+        INDRA Statements will be returned. Note that this will in general be
+        VERY large in memory, and therefore should be used with caution.
+
+    Return
```
bgyori: This should be `Returns`

indra/db/util.py
```diff
@@ -472,6 +471,241 @@ def make_stmts_from_db_list(db_stmt_objs):
...
+    for trid, tcid, src in info_tuples:
+        if trid not in tr_best.keys() or is_better(src, tr_best[trid][0]):
+            tr_best[trid] = (src, tcid)
+        if always_include is not None and src in always_include:
+            tcids.add(tcid)
+    tcids |= {tcid for _, tcid in tr_best.values()}
+    return tcids
+
+
+class NestedDict(object):
```
bgyori: I think a couple of unit tests would be nice to make sure this works as intended! I still wonder if something like `defaultdict(defaultdict)` could achieve a similar structure.

pagreene: I agree. I'm actually working on those now. The `get`, `get_path`, `gets`, and `get_paths features` would not be available from the nearest equivalent using `defaultdict`. I'm still not completely convinced that the added features are worth adding the features, though. They could certainly make life easier and more convenient in some cases, but I'm not sure this is that case.

PR #503, opened 2018-04-17
Flag missing config keys

pagreene:
Flag any mismatches in keys between the user's config file and the default config file with warnings. Also give config tools their own file.

### DIFF
--- a/indra/__init__.py
+++ b/indra/__init__.py
@@ -1,12 +1,7 @@
 from __future__ import print_function, unicode_literals
 import logging
 import os
-import shutil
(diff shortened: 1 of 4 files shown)


bgyori:
I think this is good! Because this is so central, I am wary of merging it before Travis starts working again.

pagreene:
For what it's worth, I did run the full suite of tests locally, and didn't
get any related errors.

On Tue, Apr 17, 2018 at 5:40 PM, Benjamin M. Gyori <notifications@github.com

PR #506, opened 2018-04-19
Cleanse reach input

pagreene:
This PR implements a couple of optional (but performed by default) checks on the content being read by REACH in the hopes of reducing the number of hang-ups. Specifically, the following types of content will be automatically skipped:
- Content that is long (> 5e5 by default). It is unclear whether such long content creates genuine stalls or is simply timed out before it can be read. In addition, from examining a few example cases, over-long text can often be a symptom of other anomalies.
- content that is more than 50% spaces. This targets a particular bug in some of the Elsevier content in which spaces are inserted between every character, but would also be an indicator of something odd in itself.

*Note that this PR adds two new config settings to the config file*

### DIFF
--- a/indra/db/reading_manager.py
+++ b/indra/db/reading_manager.py
@@ -82,7 +82,7 @@ if __name__ == '__main__':
     args = parser.parse_args()
 
 from indra.tools.reading.db_reading import read_db as rdb
-from indra.tools.reading.readers import get_reader
+from indra.tools.reading.readers import get_reader_class
 
 logger = logging.getLogger('reading_manager')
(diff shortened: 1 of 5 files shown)


bgyori:
This looks good, with one request: I don't think REACH_CHARACTER_LIMIT and REACH_MAX_SPACE_RATIO should be defined in the config file. These are the kinds of parameters that are (1) not specific to a given user's environment (i.e. not a local path on a machine) and (2) we or another user may wish to change the values from one application to another. Therefore I would prefer if these were passed in as arguments to the reading job, with these values available as defaults locally in the reading script.

PR #507, opened 2018-04-23
Implement simple client for INDRA's Database REST API

pagreene:
This PR adds a simple wrapper around the REST API for the database. A small change is also made to the `config.py` tool, which allows for an exception to be raise internally if a key is not resolved, depending on an keyword argument. The default behavior remains unchanged, so there should be no issue with backwards compatibility. Two new keys were also added to the config file.


bgyori:
Great! I removed my commit to move this submodule to `indra.databases` from `indra.sources`, we can revisit that issue later. In the interest of time I'll merge it like this.

PR #510, opened 2018-04-24
Medscan processor can process a directory full of csxml

djmilstein:


indra/sources/medscan/api.py
```diff
@@ -48,14 +136,25 @@ def process_file(filename, medscan_resource_dir, num_documents=None):
...
     mp = MedscanProcessor(medscan_resource_dir)
-    logger.info("Parsing %s to XML" % filename)
+
+    if progress_updates:
+        logger.info("Parsing %s to XML" % filename)
     with open(filename, 'rb') as f:
         mp.process_csxml_from_file_handle(f, num_documents)
     return mp
+
+if __name__ == '__main__':
```
bgyori: This block should be removed

indra/sources/medscan/api.py
```diff
@@ -48,14 +136,25 @@ def process_file(filename, medscan_resource_dir, num_documents=None):
     num_documents : int
         The number of documents to process, or None to process all of the
         documents within the csxml file.
+    progress_updates: bool
+        Whether to print progress updates to the console
```
bgyori: The typical solution for this is to change the logging level of the logger for this module. If a user wants to suppress logging here, they can set the level to logging.WARNING, for instance. So I don't think this needs to be an API argument.

indra/sources/medscan/api.py
```diff
@@ -1,13 +1,101 @@
...
+    s_dict = defaultdict(list)
+    mp = process_directory(directory_name, medscan_resource_dir)
 
-def process_file(filename, medscan_resource_dir, num_documents=None):
+    for statement in mp.statements:
+        s_dict[statement.evidence[0].pmid].append(statement)
+    return s_dict
+
+def process_directory(directory_name, medscan_resource_dir):
+    """Processes a directory filled with CSXML files. For each file, first
```
bgyori: Docstring should start with single line describing what the function does/returns. More text can be provided after a blank line below that. The parameters and returns should also be documented.

indra/sources/medscan/api.py
```diff
@@ -1,13 +1,101 @@
...
+import pickle
+from indra.sources.medscan.fix_csxml_character_encoding import \
+        fix_character_encoding
+import time
 
 logger = logging.getLogger('medscan')
 
+def process_directory_statements_sorted_by_pmid(directory_name,
+                                                medscan_resource_dir):
+    """Processes a directory filled with CSXML files. For each file, first
```
bgyori: Docstring should start with single line describing what the function does/returns. More text can be provided after a blank line below that. The parameters and returns should also be documented.

indra/sources/geneways/processor.py
```diff
@@ -107,15 +108,9 @@ def make_statement(self, action, mention):
                     sentences = ftm.find_matching_sentences()
                     if len(sentences) == 1:
                         text = sentences[0]
-                    else:
-                        text = None
-                else:
-                    text = None
             except:
```
bgyori: There should be a specific Exception type used here

djmilstein:
Thank you! I will work on these and write a comment on this thread once I
am done committing.

On Mon, Apr 30, 2018 at 1:41 AM, Benjamin M. Gyori <notifications@github.com

djmilstein:
I am done committing. I wasn't sure what exception to catch in the Geneways
processor so I left that be.

bgyori:
You can just use `except Exception` unless you want a more detailed one, but `except` alone is not so good. 

>Because except: catches all exceptions, including SystemExit, KeyboardInterrupt, and GeneratorExit (which is not an error and should not normally be caught by user code), using a bare except: is almost never a good idea. In situations where you need to catch all “normal” errors, such as in a framework that runs callbacks, you can catch the base class for all normal exceptions, Exception```

djmilstein:
This is helpful information to know! All feedback to date now incorporated
on this PR.

On Mon, Apr 30, 2018 at 5:04 PM, Benjamin M. Gyori <notifications@github.com

johnbachman:
Overall this looks great. Two requests:

- In `_urn_to_db_refs`, look up mappings from GO, MESH, ECCODE, and MEDSCAN against the FamPlex equivalences file to see if there are equivalent FPLX groundings. For comparison, see `_get_grounding_terms` in indra.sources.trips.processor, or `_get_db_refs` in indra.sources.reach.processor.

- Fix issue #527 and create a test for it using sample content.

johnbachman:
Excellent! Looks great. 👍

PR #511, opened 2018-04-26
Add CLI option to RAS Machine for setting number of days

cthoyt:
Sometimes, you just want to make that number bigger to get some more literature from even further back. 

This PR adds an option to the command line interface that lets the user override this number (even if it's explicitly set in the config.yaml, not to be confused with global configuration) when using the command line interface.

Example:

```sh
$ indra-machine run_with_search -d 40 /path/to/my/machine
```


bgyori:
Looks great, thanks! Since you submitted the PR from your master, I'd rather not merge it here. I just merged and pushed via git - so I'm closing the PR.

PR #514, opened 2018-04-27
Submodule dependencies

djmilstein:
Adds to setup.py the python module dependencies for various submodules (such as assemblers, processors, etc)


bgyori:
There are a few issues with this:
- `sympy` (via pysb) and `objectpath` (explicitly) are installed as standard dependencies, they can be removed from extras
- the correct dependency where applicable not `pyjnius` but `jnius-indra`
- please merge `ndex_cx` and `ndex_client` into just `ndex`, same for `bel` and `pybel_assembler` into just `bel`

bgyori:
Also, I suspect the list is incomplete. For instance, I know that the graph_assembler depends on pygraphviz but don't see it in this list.

bgyori:
The `yaml` package is actually called `pyyaml`

djmilstein:
Agreed and fixed

On Fri, Apr 27, 2018 at 7:54 AM, Benjamin M. Gyori <notifications@github.com

PR #518, opened 2018-05-01
Fix travis signor errors

djmilstein:
Fixes bugs in the signor processes so that the tests run on Travis; have Complex statements be generated in deterministic order so that a test reliably passes


bgyori:
Hooray! 🍾

PR #535, opened 2018-05-12
Medscan protein modification site

djmilstein:
When processing protein modification events, include the modification site if available


bgyori:
Looks good!

PR #537, opened 2018-05-14
Refactor output assemblers into their own submodules

### LINKED ISSUE #504: Restructure indra.assemblers module
Currently the `indra.assemblers` module is structured such that all specific assemblers (which typically each consist of a single file) are directly under the module. To avoid having to do very verbose imports like `from indra.assemblers.pysb_assembler import PysbAssembler`, all the Assembler classes are imported in `indra.assemblers.__init__`. But then to avoid errors due to missing optional dependencies, all the imports there are wrapped in try/except.

I propose a new module structure in which all the output assemblers get their own submodule, e.g. `indra.assemblers.pysb`, `indra.assemblers.sbgn`, `indra.assemblers.cag`. Then a typical import would look like `from indra.assemblers.pysb import PysbAssembler`. We could possibly simplify even further and rename all the assembler classes to just Assembler, in which case they could be imported as `from indra.assemblers.pysb import Assembler`. But in this case, if multiple assemblers are used in a script, one would need to also add `import ... as ...` to the import to give each assembler class a unique local name. 

I am also somewhat concerned that internal knowledge assembly (like `indra.preassembler`, `indra.belief`, `indra.mechlinker`, `indra.tools.assemble_corpus`) is often confused with output model assembly, and I'm not sure that the `indra.preasssembler` and `indra.assemblers` naming helps clarify this distinction.

bgyori:
This PR refactors `indra.assemblers` into submodules for each output assembler. A typical import pattern would now be
```
from indra.assemblers.pysb import PysbAssembler
```
This resolves #504. An important benefit of this change is that `indra.assemblers` doesn't import assembler classes at the module level anymore leading to less import issues and more transparency regarding custom/optional dependencies of various output assemblers. Once merged, dependencies will need to be updated simultaneously.


johnbachman:
So I think the key question here is if there are any assembler modules with substantial functionality where someone would want to import the module rather than just the class. For example, if there were lots of non-class functionality in `indra.assemblers.pysb`, this creates the situation where one would want to do `from indra.assemblers import pysb`, which would then lead to namespace conflicts with `pysb`. If in all the relevant cases the functionality is mainly in the class, then this is fine.

It is definitely nice that we eliminate the try/except blocks from the top-level API, and prevent importing everything even if one assembler is needed, which will presumably speed things up.

johnbachman:
Also this is a non-backward compatible change--are we going to increment the major version for this, or try to roll it into another release with any other pending/contemplated non-backward compatible changes?

bgyori:
I agree, in case of conflicts, the recommended approach would be something like ``import indra.assemblers.pysb.assemler as pysb_assembler```. We don't officially follow semantic versioning and it's easy to update dependencies (I've actually already prepared some PRs in other repos) so I think this could just be rolled into the next release, I guess 1.8.0.

johnbachman:
👍

bgyori:
This is somewhat disruptive and can cause issues in downstream applications so we should revisit it in a couple of months.

PR #539, opened 2018-05-15
Fix handling of Elsevier and clean up content handling in general in new reading

pagreene:
This PR's primary purpose is to fix the handling of Elsevier content, by extracting the text from Elsevier XML before reading. However, to enable this, and also to pull database dependencies out of the reading architecture, this PR also introduces a new Content object (`indra.tools.reading.readers.Content`), which is now the standard class used for content read by all the readers, whether the content is already loaded, e.g. from the database.

In addition, various small fixes were made.


pagreene:
I tested this on 100 randomly sampled Elsevier texts, and the Elsevier content was correctly produced. The general reading tests are also passing.

bgyori:
Great!

PR #540, opened 2018-05-15
Fix db api handling of FPLX and fruitless roles

pagreene:
This PR:
- Handles the FPLX->BE rift that currently exists in the database (this will be remedied in the near future, but must be patched for now).
- Correctly handles the case were no statements remain at the end of a check.


bgyori:
Looks good! This is really not a big deal but whenever you do a check on `len(stmts)` to see if the list is empty, those could be replaced by just `stmts` which is the recommended way to test for empty lists.

pagreene:
Oh? I have always preferred using `len`, because it avoids ambiguities. Like John always says: better explicit than implicit. Who recommends you not use `len`?

PR #546, opened 2018-05-19
REACH processor updates

bgyori:
This PR makes a number of changes to the REACH processor:
- eliminates the case where HGNC ID is set for a non-human protein
- eliminates redundant complexes that are duplicate extractions
- eliminates the Agent name sanitization which is not needed anymore in input processors
- adds complex controller extraction to RegulateAmounts and standardizes it with RegulateActivities


bgyori:
Looks like the way I filtered redundant Complexes is not so good after all, it gets rid of some legit Complexes. I'll try to find a better solution.

johnbachman:
Did you find another solution for the redundant Complexes?

bgyori:
Yes, I tested it on a bunch of examples and the way it's implemented now seems to be working as expected.

PR #547, opened 2018-05-19
Add web uploader to PybelAssembler

cthoyt:


bgyori:
Thanks, this is useful! 

Are the values (host and credentials) also configurable in a config file for PyBEL? If not, I would like to use INDRA's `indra.config.get_config` to fetch these values from the INDRA config file or an env variable before passing them to `pybel`. We would then also also add these keys to `indra/resources/default_config.ini`.

bgyori:
The `Returns`, which I believe is `requests.Response` could also be added to the docstrings.

bgyori:
I'm happy to add these, just let me know about the config

johnbachman:
Also rebase off the latest sorgerlab/master--I just merged the changes to the pybel_assembler that we put in while you were here. I don't expect any conflicts though.

cthoyt:
I've added another wrapper for `pybel.to_database` as well

@bgyori documentation updated - these functions make use of the PyBEL configuration system and I've added a bit more information on how to mess with them. Maybe we should talk a bit more to see if we should inject INDRA config in there so it can all be consolidated in INDRA optionally

@johnbachman I just used GitHub's resolve system to take care of this - quite elegant and easy since there weren't any conflicts

cthoyt:
@bgyori for example, if we wanted to consolidate configuration to the INDRA file, then the code could be updated to:
 
```python

import pybel
from pybel.constants import PYBEL_REMOTE_HOST, PYBEL_REMOTE_PASSWORD, PYBEL_REMOTE_USER

...

response = pybel.to_web(
   self.model, 
   host=host or get_config(PYBEL_REMOTE_HOST), 
   user=user or get_config(PYBEL_REMOTE_USER), 
   password=password or get_config(PYBEL_REMOTE_PASSWORD),
)
```

PR #550, opened 2018-05-20
Provide feedback when too many statements found.

pagreene:
This PR alters the REST API service to provide more detailed feedback to the user if there are too many statements, specifically returning a count of each statement type.


pagreene:
The service has already been updated with this alteration.

PR #551, opened 2018-05-20
Improve indra db REST client

pagreene:
This PR:
- allows the user to give a parent class of statement, and get the all the statements corresponding to it's children, e.g. given RegulateAmount, the `get_statements` function will now also return statements for `IncreaseAmount` and `DecreaseAmount`, and others.
- allows the client to break up queries that are too large into smaller chunks by statement type, thus circumventing the REST service's limit of 10,000 statements per request.

### DIFF
--- a/indra/sources/indra_db_rest/client_api.py
+++ b/indra/sources/indra_db_rest/client_api.py
@@ -29,17 +29,28 @@ class IndraDBRestError(Exception):
 
 def _make_stmts_query(agent_strs, params):
     """Slightly lower level function to get statements from the REST API."""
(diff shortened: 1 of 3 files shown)


pagreene:
All relevant tests verified passing locally.

pagreene:
I have now added four options for behavior "on_limit": 'sample' (default), 'truncate', 'persist', and 'error'. The previous behavior where the error message was analyzed, and all the statements were retrieved in many smaller queries is available through 'persist'. The rest are pretty self-explanatory, I think.

These options are available in part due to updates in the REST API which have already been propagated to the active service.

PR #556, opened 2018-05-23
JSON schema

djmilstein:
JSON schema for lists of JSON-serialized INDRA statements.

Created by manually "translating" the JSON-generation code in statements.py. Validated by:
* Confirming a random sample of 10,000 INDRA statements from REACH is valid
* Perturbing a valid statements in invalid ways to see if it breaks schema validation


cthoyt:
Maybe it's a bit overkill, but there's a Python package called `jsonschema` that you could use to unit test the schema.

CC @maestroe

bgyori:
Hooray!

cthoyt:
You might also want to check out https://github.com/mlakewood/hypo_schema, which takes a JSON schema and automatically generates instances (that you can convert into INDRA statements)

PR #560, opened 2018-06-04
Fix issue with CAG nodes not being grounded properly in viz

adarshp:
The grounding threshold was automatically being set to `None` in the `make_model` class method, even if it was specified during instantiation.


adarshp:
Superseded by #561.

PR #561, opened 2018-06-04
Additional eidos groundings

adarshp:
This PR adds functionality to process additional Eidos groundings (WDI, FAO)

### DIFF
--- a/indra/assemblers/cag_assembler.py
+++ b/indra/assemblers/cag_assembler.py
@@ -44,7 +44,7 @@ class CAGAssembler(object):
         """Add a list of Statements to the assembler."""
         self.statements += stmts
 
-    def make_model(self, grounding_threshold=None):
+    def make_model(self, grounding_ontology='un', grounding_threshold=None):
         """Return a networkx MultiDiGraph representing a causal analysis graph.
 
         Parameters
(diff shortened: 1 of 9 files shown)


bgyori:
Thanks, this is a good start but the scored groundings are used in a number of other assembly components and I will need to go in an reimplement those too to be able to handle a deeper dict structure.

adarshp:
Potentially you could save some work by flattening the structure - instead of having

```python
db_refs['EIDOS'] = {
'un': [('x', 1.0), ...]
'fao': [('y', 1.0), ...]
...
}
```

You could do:
```python
db_refs = {
'un': [('x', 1.0),...]
'fao': [('y', 1.0), ...]
...
}
```

Just a thought..

bgyori:
Yes, that's a good idea, I guess UN, FAO, WDI are not Eidos-specific so we can do away with the EIDOS key and use those instead - it would also solve this issue since I would just need to change EIDOS to UN in a couple of places. I'd capitalize it to be consistent with other db_refs keys, like
```
db_refs = {
'UN': [('x', 1.0),...]
'FAO': [('y', 1.0), ...]
```
What do you think? I can add this on top of your branch tonight of that sounds good to you.

adarshp:
Sounds good to me!

bgyori:
This should work now according to what we discussed above, let me know if you run into any issues!

adarshp:
Works great, thanks!

PR #563, opened 2018-06-04
Enable incremental preassembly on the database (extensive refactor).

pagreene:
This PR enables the automatic incremental update of preassembled statements in the database by:
- breaking evidence out of the statement json, to be repopulated on-load, by linking to "raw" statements using a separate id-pair table. This allows for pre-assembled statements to be traced in the database refs all the way back to, for example, a pmid, and vice-versa. This flat structure also enables the addition of new "evidence" to a pre-assembled statement as it becomes available without re-processing the entire corpus.
- changes the primary key of raw statements to uuids, for reproduceability, and similarly uses an altered matches key as the primary index of pa statements.
- similarly breaking support-supported by relations out of the statement jsons, and storing them in a separate table, with pairs of the modified matches keys, with the standard list-of-reference structure re-constructed on-load.
- Implements a batched pre-assembly pipeline for the database, incorporating the above features.
- Implements a batched incremental pre-assembly algorithm designed for the database, with only minimal and completely backwards compatible changes to the Preassembler code.
- Adds new features to Statements, to support all of the above.
- Adds some new funcitonality elsewhere, most notably the function `batch_iter`, which efficiently breaks up iterables into batches.
- This PR also fixes a known issue where "pmids" in evidence were actually text content ids, by filling the value with the actual pmid from the text ref on-load (made possible by the above changes).
- Implements rigorous and detailed tests, comparing the new pre-assmebly pipeline (both a single step, and an incremental build) and retrieval to the results of "classic" preassembly of a corpus, ensuring that the results are identical (all evidence matches, and all supports-supported by relations are the same).


bgyori:
Wow, this is quite a PR! I will look at changes to statements.py and preassembler.py but otherwise should be good to go.

bgyori:
There are a couple of issues that make tests fail and should be fixed before merging:
- test_db_preassembly fails in Python 2, the test pickles can't be loaded. We should probably skip all the tests in this file in Python 2.
- test_db_info_tuples.pkl needs to be added to the downloads in .travis.py
- it looks like some of the new tests (e.g. test_db_preassembly.test_statement_distillation) are quite slow and so should maybe be tagged as such so that Travis doesn't routinely run them

pagreene:
Yeah, I admit that I was putting off doing Python 2 checks, in part because I wasn't sure it would be worth maintaining compatibility especially if we break the project into a separate repo.

As for the tests, I think I might actually have two different runs, one with a very small set of statements that probably won't test most of the logic (small chance of duplicates, support, multiple evidence, etc), but will act as a smoke test to make sure the basic cod runs. Then for local running, have tests with actually usefully large samples of statements. The tests may still be slow, though. We'll see.

bgyori:
Is it possible that tests are failing because they are trying to access a shared, single db instance / connection? It might be better to have each test case start a completely fresh connection to avoid cases where files are loded out of synch.

pagreene:
It does look like that, but they _shouldn't_ be using the same connection. Certainly that isn't how it is written (hence it works locally), but maybe Travis is doing something funky? I'm digging into it now.

pagreene:
This is the problem: for some reason the test on Travis doesn't think it can import pgcopy.

```db_manager: WARNING: You are not using postgresql or do not have pgcopy, so this will likely be very slow.```

PR #565, opened 2018-06-06
Eidos CLI

bgyori:
This PR adds a Python CLI wrapper around Eidos to allow running it from within Python via the command line. This is complementary to using the Python-Java bridge and is meant to facilitate batch processing of large numbers of files with Eidos from a Python environment. Example usage:

```python
from indra.sources.eidos import eidos_cli
stmts = eidos_cli.extract_and_process('input_folder', 'output_folder')
```
runs `ExtractFromDirectory` on the `input_folder`, produces JSON-LDs in the `output_folder` and then processes those to return a list of INDRA Statements.

In addition, 
```eidos_cli.extract_from_directory```
 just runs reading without subsequent processing, and 
```eidos_cli.run_eidos``` 
is a generic interface allowing running any other endpoint of Eidos other than `ExtractFromDirectory`.

Could be useful for @adarshp

### DIFF
--- a/doc/modules/sources/eidos/index.rst
+++ b/doc/modules/sources/eidos/index.rst
@@ -21,3 +21,9 @@ Eidos Reader (:py:mod:`indra.sources.eidos.eidos_reader`)
 
 .. automodule:: indra.sources.eidos.eidos_reader
     :members:
+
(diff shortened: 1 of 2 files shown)


adarshp:
This does look useful, thanks for the heads-up!

PR #566, opened 2018-06-07
Clean some imports, add __all__ to statements.py.

pagreene:
Something I did a while ago that I still think would be worth adding to master. (Note that this went through a rebase since I last tested, so there are liable to be bugs, which I will fix as they show up on Travis).


bgyori:
This poses an unnecessary risk without having a clear contribution to functionality. We can revisit it as part of a larger refactoring.

PR #567, opened 2018-06-07
Add synonym capabilities to the uniprot client.

pagreene:
This will be used by the BioSense agent in CwC's Bob with Bioagents.

### DIFF
--- a/indra/databases/uniprot_client.py
+++ b/indra/databases/uniprot_client.py
@@ -265,6 +265,56 @@ def get_gene_name(protein_id, web_fallback=True):
         return gene_name
     return None
 
+
+def get_gene_alt_labels(protein_id):
+    """Return a list of other names for the protein."""
+    g = query_protein(protein_id)
+    if g is None:
+        return None
+    query = rdf_prefixes + """
+        SELECT ?name
+        WHERE {
+            ?gene skos:altLabel ?name .
+            }
+        """
+    res = g.query(query)
+    if res:
+        return [r[0].toPython() for r in res]
+    return None
+
+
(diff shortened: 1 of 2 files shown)


bgyori:
This is good, I'd like to ask for two things:
1. The prefLabel should also be extracted along with the altLabels e.g. from
```
<skos:prefLabel>KRAS</skos:prefLabel>                                           
<skos:altLabel>KRAS2</skos:altLabel>                                            
<skos:altLabel>RASK2</skos:altLabel>                                            
```

2. The names and docstrings are a little bit confusing, I suggest renaming them as:
```
def get_gene_synonyms(protein_id):
    """Return a list of synonyms for the gene corresponding to a protein."""

def get_protein_synonyms(protein_id):
    """Return a list of synonyms for a protein."""

def get_synonyms(protein_id):
    """Return all synonyms for a protein including synonyms for its associated gene."""
```

pagreene:
The reason I didn't get the "prefLabel" is that that is what is used by `get_gene_name`. As per the no-copying principal, it seemed better to not repeat the capability, and give an informative name that tried to clarify the point. So perhaps what you called `get_gene_synonyms` would be better named `get_alternative_gene_names`, thus making it's relation to `get_gene_name` clearer. Then perhaps I could name the other one `get_all_protein_names`? The last function could then be called `get_all_names`. What do you think?

bgyori:
Please include preflabel as a synonym (thus the gene name will also be considered a gene synonym). The distinction between name and synonyms is clear. Please follow the function names as suggested. Thanks!

PR #570, opened 2018-06-18
Api evidence

pvtodorov:
- Changes to REST API
  - New assembler endpoints
    - assemble English sentences from model
    - assemble CX and share it to NDEX

  - Get evidence from INDRA database
    - filter evidence to agent names in request

- Changes to CX Assembler
  - the `ndex_cred` param in `upload_model` is now optional. if not supplied, an attempt is made to retrieve it from the config file

- Changes to NDEX client
  - `create_network` now has an optional param `private` which is `True` by default. If set to false, the model will be uploaded to NDEX as a public network.

### DIFF
--- a/indra/assemblers/cx_assembler.py
+++ b/indra/assemblers/cx_assembler.py
@@ -168,7 +168,7 @@ class CxAssembler(object):
             cx_str = self.print_cx()
             fh.write(cx_str)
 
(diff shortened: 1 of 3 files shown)


bgyori:
Great, thanks! I rebased the branch and working on updating the service.

PR #572, opened 2018-06-18
Fix cwms tests

pagreene:
Capture ONT::INCREASE events, add tests for this case, and update some other tests.


pagreene:
Note that the failing tests are related to the pubmed issue noted elsewhere.

bgyori:
Thanks!

PR #574, opened 2018-06-19
Add PFAM regex in PyBEL assembler

cthoyt:
Also move PUBCHEM to a dictionary so its prettier when we add more


indra/assemblers/pybel_assembler.py
```diff
@@ -98,9 +98,7 @@ def __init__(self, stmts=None, name=None, description=None, version=None,
                   'namespace/swissprot/swissprot-20170725.belns',
             'IP': 'https://arty.scai.fraunhofer.de/artifactory/bel/'
                   'namespace/interpro/interpro-20170731.belns',
-            'FPLX': 'https://raw.githubusercontent.com/sorgerlab/famplex/'
```
cthoyt: Wait, noo! This was a mistake

PR #580, opened 2018-06-26
Update handling of Eidos and BBN ontologies

pagreene:
This adds the capability to get the BBN ontology, and also updates the handling of the Eidos ontology, and furthermore merges the two, as they are handled in nearly exactly the same way. Also includes the new/updated ontology rdfs.


bgyori:
This is great! Later we may want to change all entries in bbn_ontology.rdf to prepend the grounding with BBN (i.e. `http://...#BBN/...`). This is because in integrated applications multiple ontologies are loaded in a single hierarchy manager and the namespace for the Eidos and BBN ontologies are now shared. In practice this currently won't cause problems because the Eidos ontologies always use a prefix like UN.

PR #582, opened 2018-06-27
Sofia processor updates

bgyori:
This PR by @johnbachman updates the Sofia input processor to work with its latest output format, and extend the set of relations we extract.


bgyori:
Works perfectly, I just changed the subject polarity to be None by default instead of being assumed to be positive to match assumptions in other processors.

PR #586, opened 2018-06-28
New assembly filter for db refs.

pagreene:
Adds a new filter function to select statements based on their db refs.


bgyori:
I made a number of changes to make it work with the lists of scored groundings we get from Eidos.

PR #587, opened 2018-06-28
Add a couple more ontologies to cwms.

pagreene:
These came up when running the paragraph: "Education improves thinking. Thinking increases doubt. Doubt destabilizes the totalitarian government."

Words like "improves" and "destabilizes" don't strike me as exceptionally odd, or worth omitting, and the ontologies they are mapped to seem relevant.


bgyori:
Sure!

PR #592, opened 2018-07-05
Get statement essentials only

pagreene:
This PR give users the option to get only some essential metadata of statements, allowing fast light-weight queries.


pagreene:
Fixes #595

PR #594, opened 2018-07-09
Create function to list the hierarchy relation of a statement.

pagreene:
This is a feature requested by someone in the lab, which seems useful and was easy to implement.


johnbachman:
Interesting. What's the use case?

PR #597, opened 2018-07-11
NewsAPI client

bgyori:
This PR adds a basic client for NewsAPI which allows searching recent news articles for any topic of interest.


pagreene:
Cool! Should we start reading some of these?

bgyori:
Yes! We should experiment a bit and see what we can get that might be useful.

PR #599, opened 2018-07-17
Add agent activity to PySB rule name

bgyori:
For some reason agent activities were not made part of the rule name yet which could have resulted in inadvertent rule name collisions and made influence maps confusing.


bgyori:
So this turned out to be a bit more complicated. It revealed an issue in which assemblies of various state-changing rules would overwrite specified states of Monomers (e.g. if "active" RAF is activated by a rule, the left hand side RAF would be rewritten to be "inactive") - this is now changed to prioritize context that is pre-specified on an Agent/Monomer. This then revealed some tests that were buggy because they missed adding some activity pre-conditions, and so shouldn't have had proper mechanistic paths to a given Observable as previously reported - these are now fixed.

PR #603, opened 2018-07-24
Update call to pybel.to_database in PybelAssembler

cthoyt:
This is for the newest version of PyBEL, where all `None or str or pybel.Manager` arguments were changed to `None or pybel.Manager`


bgyori:
Thanks!

PR #604, opened 2018-07-24
Streamline assemble_corpus code with updated logic

cthoyt:
1. Collapsed the logic of the if statement directly to the assignment
2. Switched order of if statement to make the continue happen faster and decrease code indentation


bgyori:
Okay, I'll merge this but unless I'm missing something, neither the logic, nor the speed of the filter is meaningfully affected so I'm not sure about the goal.

cthoyt:
These were purely cosmetic. I figured that while I was grokking the code, I could make tiny improvements!

PR #605, opened 2018-07-26
Add python-sbml as an extra dependency

bgyori:
This PR adds `python-sbml` as an extra dependency which is a new PySB dependency needed to generate SBML exports of models. As an aside, an extra argument is added to the PysbAssembler that allows controlling the `name` attribute of the Model being generated (for SBML and some other exports, None name errors).


cthoyt:
@ddomingof you might want to check this

PR #606, opened 2018-07-26
Fix elsevier client.

pagreene:
I found the bug in the client that was allowing error messages through as valid content and fixed it. I also left an extra check in case there there are any genuine Elsevier bugs.


indra/literature/elsevier_client.py
```diff
@@ -129,8 +129,8 @@ def download_article(id_val, id_type='doi', on_retry=False):
     params = {'httpAccept': 'text/xml'}
     res = requests.get(url, params, headers=ELSEVIER_KEYS)
     if res.status_code == 404:
-        logger.debug("Resource for %s not available on elsevier."
-                     % url)
+        logger.info("Resource for %s not available on elsevier." % url)
+        return None
```
pagreene: This is where the problem was.

PR #616, opened 2018-08-10
Updating PyBEL processor for PyBEL 12

cthoyt:
Closes pybel/pybel#331

### DIFF
--- a/indra/statements.py
+++ b/indra/statements.py
@@ -1074,7 +1074,7 @@ class Statement(object):
 
     Parameters
     ----------
-    evidence : list of :py:class:`Evidence`
+    evidence : None or :py:class:`Evidence` or list of :py:class:`Evidence`
         If a list of Evidence objects is passed to the constructor, the
         value is set to this list. If a bare Evidence object is passed,
         it is enclosed in a list. If no evidence is passed (the default),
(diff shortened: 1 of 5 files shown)


cthoyt:
Sorry in advance, I had to remove all the parentheses around the tuples to make PyCharm stop complaining :p

indra/statements.py
```diff
@@ -1074,7 +1074,7 @@ class Statement(object):
 
     Parameters
     ----------
-    evidence : list of :py:class:`Evidence`
+    evidence : None or :py:class:`Evidence` or list of :py:class:`Evidence`
```
johnbachman: What would cause evidence to be None? If a statement is synthetic it should just be an empty list.

johnbachman: Never mind!

johnbachman:
One question--because the hash/key change is specific to pybel/develop, merging this will make INDRA incompatible with pip-installed pybel. Will this be merged into master and released in the near future?

cthoyt:
Changing the way the hashes are stored required changing a lot of functions that handled these operations. Luckily, I put a lot of time into unit testing each so now it's pretty easy to find any issues that pop up. I'd like to release these changes to PyPI soon (possibly with the networkx 2.1 update)

However, I'm unclear on how INDRA does dependencies. Do you have a version of PyBEL pinned, or is it always pointing to the latest on GitHub?

PR #623, opened 2018-08-27
Sources name refactoring

kkaris:
This PR renames a number of files from the format "module_api.py" to "api.py" (or similar with "module_processor.py" etc). Refactoring has been done in all affected files in the indra directory.


bgyori:
Looks good so far! In addition to the python files, I now realize, you'll also need to go into the `doc/modules/sources` folder, and look over the .rst files to change references to the renamed files.

indra/sources/trips/client.py
```diff
@@ -123,9 +123,9 @@ def save_xml(xml_str, file_name, pretty=True):
                                       ['string=', 'file=', 'output=', 'help'])
     for o, p in opts:
         if o in ['-h', '--help']:
-            print('String mode: python trips_client.py '
+            print('String mode: python client.py '
                   '--string "RAS binds GTP" --output text.xml')
-            print('File mode: python trips_client.py --file test.txt '
+            print('File mode: python client.py --file test.txt '
```
pagreene: Perhaps this message should be refactored for clarity? e.g. `python trips/client.py ....`, or maybe even `python -m indra.sources.trips.client ...`?

kkaris: Changed to what you suggested "-m indra.sources.trips.client"

kkaris:
:(
https://travis-ci.org/sorgerlab/indra/builds/421294535

bgyori:
Okay so so far I haven't found many problematic imports in our own packages that use INDRA (mainly because we always use imports like `from indra.sources import biopax` which don't change). 

But it looks like a few other repos contain code that do imports like `from indra.sources.biopax import biopax_api` which will have to be updated using one of the following import patterns
```python
from indra.sources import biopax # Preferred but may need downstream changes in code
import indra.sources.biopax.api as biopax_api # Not preferred but rest of code doesn't need to change
```
I found the following problematic uses of indra.sources imports:
- PyBEL: https://github.com/pybel/pybel/blob/master/src/pybel/io/indra.py#L126,  https://github.com/pybel/pybel-notebooks/blob/master/Converting%20BioPAX%20to%20BEL.ipynb, https://github.com/cthoyt/notebooks/blob/master/BioCreative%20with%20INDRA.ipynb
- Several notebooks in https://github.com/ndexbio/indra_me
- PyKAMI (deprecated): https://github.com/johnbachman/pykami/blob/master/pykami/indra_to_kami_example1.py#L1-L2
- Syndra: https://github.com/csvoss/syndra/blob/d3d5b13ca3c0fa7c52a45744ce04071175adb41f/interface_indra_to_syndra.py#L15

There are also a couple of places where things other than the APIs are used e.g. `trips_client` is used in BIoagents which I have a fix ready for.

indra/sources/trips/client.py
kkaris: I edited that here in github

indra/sources/trips/client.py
```diff
@@ -123,10 +123,10 @@ def save_xml(xml_str, file_name, pretty=True):
                                       ['string=', 'file=', 'output=', 'help'])
     for o, p in opts:
         if o in ['-h', '--help']:
-            print('String mode: python trips_client.py '
+            print('String mode: python -m indra.sources.trips.client.py '
```
kkaris: @pagreene Here is the change. I don't know why it didn't show up earlier...

PR #627, opened 2018-08-30
Db download optimize

pagreene:
This PR implements deep changes to the way statements can be queried from the database to allow for rapid and finely controllable access. This involves significant changes to `indra.db.client`, `indra.db_rest_api.api` (which uses the `indra.db.client`), and `indra.sources.indra_db_rest.client_api`, which is the local endpoint a REST api running `indra.db_rest_api.api`. Despite the depth of change, the behavior for the end user of the rest service is the same by default (e.g. `indra.sources.indra_db_rest.client_api` is backwards compatible), but there are now many new features, key among them:
- you control how much evidence you get for each statement
- you can control the number of statements you get back
- you have the ability to allow a query to be paged in the background, while you examine the initial response (you can have a hybrid slow/fast response).

Typical queries run in under a second, despite all this wonderful machinery. The dev REST api is running off of this branch. Production is not, but should be changed over at the same time this PR is merged.

### DIFF
--- a/db_rest_api/README.md
+++ b/db_rest_api/README.md
@@ -38,17 +38,27 @@ query parameters.
   - Example 3: you can specify the agent id namespace by appending
   `@<namespace>` to the agent id in the parameter, e.g.
   `subject=6871@HGNC`.
-- **`agent`**: This parameter is used if the specific role of the agent
+- **`agent*`**: This parameter is used if the specific role of the agent
 (subject or object) is irrelevant, or the distinction doesn't apply to the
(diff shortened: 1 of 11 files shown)


bgyori:
I think this all looks really great!

pagreene:
The error that remains for python 2.7 cannot actually be circumvented because it is a _syntax_ error, which means the file is not actually parsed even as far as the skip test before python gives up. I think we'll just have to accept this failure for the moment, and work to separate the db code into a separate repo.

PR #632, opened 2018-09-04
Refactor assemblers into their own submodules

### LINKED ISSUE #504: Restructure indra.assemblers module

bgyori:
This is a revival of #537  which refactors `indra.assemblers` into submodules for each output assembler. A typical import pattern would now be
```
from indra.assemblers.pysb import PysbAssembler
```
This resolves #504. An important benefit of this change is that indra.assemblers doesn't import assembler classes at the module level anymore leading to less import issues and more transparency regarding custom/optional dependencies of various output assemblers. Once merged, dependencies will need to be updated simultaneously.


bgyori:
Along with this, I tried to do a comprehensive review of all repos on Github using indra.assemblers and either pushed changes directly or submitted PRs to update those repos.

PR #636, opened 2018-09-14
Implement authentication for the REST API access to the database.

pagreene:
This will redact text snippets from Elsevier for those without an API key.


johnbachman:
Nice! Can you update the redaction to display the first 200 characters of Elsevier evidence text, and if the snippet is longer, show `ev.text[0:200] + ' [Elsevier content limited to 200 characters]'`?

PR #637, opened 2018-09-14
Calculate belief scores for the database

pagreene:
This PR adds a functionality to calculate belief for the full corpus of preassembled statements on the database, making use of `MockStatement` and `MockEvidence` classes, which contain only the minimal necessary information to calculate belief, thus allowing the corpus to be processed in a reasonable amount of memory, while also utilizing the existing BeliefEngine tools.

I have confirmed that this PR is compatible with both the `master` and `belief_scoring_function` (#553)  branches.


johnbachman:
Looks like it's failing on Travis due to a missing parameter in the belief engine for 'signor'. For some reason only failing in Python 3.5, passes in 2.7.

```
ERROR: indra.tests.test_db_belief.test_belief_calc_up_to_prior
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/sorgerlab/indra/indra/tests/test_db_belief.py", line 19, in test_belief_calc_up_to_prior
    be.set_prior_probs(test_stmts)
  File "/home/travis/build/sorgerlab/indra/indra/belief/__init__.py", line 108, in set_prior_probs
    self._check_prior_probs(statements)
  File "/home/travis/build/sorgerlab/indra/indra/belief/__init__.py", line 199, in _check_prior_probs
    raise Exception(msg)
Exception: BeliefEngine missing probability parameter for source: signor
ERROR: indra.tests.test_db_belief.test_belief_calc_up_to_hierarchy
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/sorgerlab/indra/indra/tests/test_db_belief.py", line 36, in test_belief_calc_up_to_hierarchy
    be.set_prior_probs(test_stmts)
  File "/home/travis/build/sorgerlab/indra/indra/belief/__init__.py", line 108, in set_prior_probs
    self._check_prior_probs(statements)
  File "/home/travis/build/sorgerlab/indra/indra/belief/__init__.py", line 199, in _check_prior_probs
    raise Exception(msg)
Exception: BeliefEngine missing probability parameter for source: signor
```

pagreene:
That's weird, on my build (pagreene) it failed at sparser...

bgyori:
I think there are a couple of related issues I found when looking into this:
- _PrePaDatabaseTestSetup in test_db_client tries to open db_pa_test_input_1M.pkl which is not in version control and not added to the build by Travis so this test normally errors
- test_db_client is intended to be skipped in Python 2 only but it is actually skipped in Python 3 too, so it never ran on Travis - hence no errors previously
- in fact it looks like even if nonpublic is not set (i.e. regular, non-PR builds), somehow we skip almost all test_db_* tests with "SKIP: Every test is nonpublic and a webservice.", see e.g. https://travis-ci.org/sorgerlab/indra/jobs/428816419
- the new test imports _PrePaDatabaseTestSetup which I would now expect to error (and it does for me locally) - this one test is skipped on Travis
- I couldn't reproduce the BeliefEngine parameter missing locally!

pagreene:
Ben, you are correct that very nearly all db tests are skipped on Travis, because getting Travis access to the database proved difficult, and for the primary database would be highly unwise I think. I'm not sure where you're seeing `test_db_client` fail on Travis, because it shouldn't be run at all on Travis. The new belief tests that require a database should not be run either.

The only real error I see is with the belief engine. And it looks like it is simply not finding the priors, because it looks like the first source it comes upon fails every time, and I have no idea why that would be.

bgyori:
Sure, no primary DB. But I thought the test DB is / can be made accessible. I was seeing the test_db_client error offline due to the db_pa_test_input_1M.pkl missing. I still don't clearly see where the tests are skipped, all I noticed is the IS_PY3 check and the nonpublic/webservice attributes. What makes most of the tests be skipped on Travis in Python 3?

Issue #638, opened 2018-09-25
FileNotFoundError: large_corpus.bel


remondulos:
When running the code below, we get a FileNotFoundError exception. 
It seems that Indra cannot find the large_corpus.bel file. I'm running Indra in Jupyter using an Anaconda environment. Could you please help?

```
from indra.tools.gene_network import GeneNetwork

gn = GeneNetwork(['TMEM173'])
bel_stmts = gn.get_bel_stmts()
```

**The complete error message is:**

INFO: [2018-09-25 10:50:26] indra/pybel.io.lines - Loading from path: C:\Users\dulosr\AppData\Roaming\Python\Python36\site-packages\indra\sources\bel\../../..\data\large_corpus.bel
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-5-99500a89dcd7> in <module>()
      3 gn = GeneNetwork(['TMEM173'])
      4 # biopax_stmts = gn.get_biopax_stmts()
----> 5 bel_stmts = gn.get_bel_stmts()

~\AppData\Roaming\Python\Python36\site-packages\indra\tools\gene_network.py in get_bel_stmts(self, filter)
     71         # No cache, so perform the queries
     72         else:
---> 73             bel_proc = bel.process_pybel_neighborhood(self.gene_list)
     74             bel_statements = bel_proc.statements
     75             # Save to pickle file if we're caching

~\AppData\Roaming\Python\Python36\site-packages\indra\sources\bel\bel_api.py in process_pybel_neighborhood(gene_names, network_file, network_type)
    104                                     'data', 'large_corpus.bel')
    105     if network_type == 'belscript':
--> 106         bp = process_belscript(network_file)
    107     elif network_type == 'json':
    108         bp = process_json_file(network_file)

~\AppData\Roaming\Python\Python36\site-packages\indra\sources\bel\bel_api.py in process_belscript(file_name, **kwargs)
    212     identifier_validation = kwargs.get('identifier_validation', False)
    213     kwargs['identifier_validation'] = identifier_validation
--> 214     pybel_graph = pybel.from_path(file_name, **kwargs)
    215     return process_pybel_graph(pybel_graph)
    216 

~\AppData\Local\Continuum\anaconda3\envs\Indra\lib\site-packages\pybel\io\lines.py in from_path(path, manager, allow_nested, citation_clearing, encoding, use_tqdm, disallow_unqualified_translocations, **kwargs)
     69     """
[...]

PR #639, opened 2018-09-25
Gene network fixes

pagreene:
Fix two bugs in the gene network tool:
1. Make the path to large corpus Windows compatible [#638]
2. Fix an undefined variable bug that I found.


bgyori:
This change is good but it isn't in itself a fix for #638. The issue there is not primarily with the folder separators but the fact that the `indra` package is installed using pip which only installs the internal `indra` module in site-packages and not the folders outside the module (like `data` for instance).

pagreene:
True. It would also mean the changes would not have any effect. However, the file path would also have caused a problem. If we can merge this before he updates his installation, it should work.

bgyori:
Let me push one more thing

PR #643, opened 2018-09-28
Updated parameter signature for stmts_to_json function

adarshp:


bgyori:
Thanks! I modified this a bit to use the lowercase types e.g. `list` which has been our convention for docstrings.

adarshp:
Got it! I'll keep that in mind for future PRs.

PR #647, opened 2018-09-30
Switch print statements to use logger

cthoyt:
Feel free to discard this PR if you disagree, but I think that the print statements in `indra.sources.indra_db_rest.client_api` would better be Python logging.


bgyori:
Sure, thanks! In fact we may want to make those debug, but I'll merge it like this for now.

PR #648, opened 2018-10-03
Context representation classes

### LINKED ISSUE #110: Standardize context annotations in INDRA
We should standardize the relevant types of contextual annotations in INDRA and have each input processor produce annotations using the same standard. As a reference, BEL currently uses the following name spaces to annotate context:
- Species: 
  - NCBI Taxonomy (http://www.ncbi.nlm.nih.gov/Taxonomy/taxonomyhome.html)
- Organ/Anatomy: 
  - In BEL 1.0: MeSH
  - In latest BEL: Uberon (http://uberon.org/) and MeSH
- Disease:
  - In BEL 1.0: MeSH
  - In latest BEL: DO (http://disease-ontology.org/)
- Cell type:
  - In BEL 1.0: MeSH
  - In latest BEL: CL (http://cellontology.org/)
- Cell line:
  - In BEL 1.0: ATCC (http://www.atcc.org)
  - In latest BEL: CLO (http://www.clo-ontology.org/) 
- Cellular location:
  - MeSH

bgyori:
This PR introduces a new representation for context in Statements. Previously, context-like annotations were picked up from various sources and put into the `annotations` dict attribute of Statement Evidences with no particular schema. This PR adds a dedicated `context` attribute to each Evidence which is either None (if no context is available), or is a `Context` class. There are two types of Context classes currently implemented: `BioContext` and `WorldContext`. BioContext is applicable to mechanisms in the biology space with cellular location / cell line / cell type / organ / disease / species represented, more or less following the BEL annotation classes. WorldContext is generally applicable to Statements identified in time and space. Most context entries can be identified directly in some ontology with the `RefContext` class, which has a simple text name, and a dictionary of `db_refs` identifiers. Time needs to be represented in a specific way with start/end or duration attributes via the `TimeContext` class. All the context classes come with JSON serialization and deserialization.

The following changes were made to input processors:
- The Eidos processor now extracts `timex` entries as `TimeContext` and adds them to Evidence under `context` rather than `annotations` (locations not yet available)
- The SOFIA processor now extracts time and location as WorldContext 
- The BEL/PyBEL processor extracts all context-like annotations using BioContext
- The SIGNOR processor extracts all context-like annotations using BioContext
- The REACH processor extracts all context-like annotations using BioContext

What this PR doesn't do:
- Provide special handling for geographical locations - will make sense to do that once Eidos provides GeoNames-identified locations
- Extract context from all sources, with some remaining candidates being e.g. Hume, TRIPS, CWMS.
- Standardize and map context DB identifiers in case the same concept is provided in different ontologies by different sources: this would require delving more deeply into mappings between ontologies like CLO, EFO, CVCL, GO, MESH, DOID, UBERON, etc. Nevertheless I think this PR is sufficient to close #110 since there is substantial overlap between e.g. REACH and BEL in the context ontologies they ground to.
- The JSON schema for Statements should probably be updated once the object model for Context is finalized
- More minor issues: the context class str/reprs could be a bit less verbose, though the advantage of the current implementation is that they work as code when pasted


bgyori:
I added a \_\_setstate\_\_ method to Evidence which sets context for Evidence objects that are loaded with "old" Statements, which makes this PR fairly safe to merge.

PR #657, opened 2018-10-10
Unload statements in order.

pagreene:
Micro-PR to preserve the ordering of statements returned by the REST client.


bgyori:
Looks good!

PR #662, opened 2018-10-11
Removing sbo field from JSON serialization of Influence statements

adarshp:


bgyori:
Thanks, I agree these annotations shouldn't appear in the JSON, I'll try to think of a more general way to turn them on/off as needed.

PR #666, opened 2018-10-14
Update rdflib version to 4.2.2

cthoyt:
I'm not sure if this will pass - we'll let travis decide that

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ import sys
 def main():
     # Only install functools32 if we're in Python 2 (it's not available
     # for Python 3)
-    install_list = ['pysb>=1.3.0', 'objectpath', 'rdflib==4.2.1',
+    install_list = ['pysb>=1.3.0', 'objectpath', 'rdflib==4.2.2',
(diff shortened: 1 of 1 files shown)


bgyori:
Hi @cthoyt, the reason we fixed rdflib to version 4.2.1 is that there was a serious increase in time taken for some of the SPARQL queries that INDRA does with newer versions of rdflib. I can't recall exactly which modules were affected but we would need to do more investigation and testing before considering to merge this. Is there a reason to increase the version that would motivate this?

bgyori:
I guess one solution would be to deprecate the BEL/RDF processor and just keep the PyBEL integration - I would want to check if the other places where we use SPARQL queries are also affected by 4.2.2 first so it could take a little while.

cthoyt:
Do we have a good way to check if PyBEL minimally duplicates all of the functionalities of the BEL/RDF processor?

bgyori:
Informally I'm pretty confident that it does but we could set up a formal comparison at the Statement level to check - good idea.

bgyori:
Given #982 this can now be closed

PR #668, opened 2018-10-16
Fix the version finding regex to handle merges.

pagreene:
Fix a bug in the `get_git_info` function's regex which prevented it from parsing merge commits.


pagreene:
The failure is an unrelated issue with installing java.

`The command "sudo apt-get install -y oracle-java8-installer" failed and exited with 100 during .`

PR #669, opened 2018-10-17
Get more elsevier content

pagreene:
This PR modifies the Elsevier client text processing to allow us to get more content. In particular, we can now extract content that comes from book chapters, and so-called "simple-article"s.


bgyori:
Looks, good, I really like the path finding implementation!

PR #674, opened 2018-10-24
Patch sparser's GO groundings.

pagreene:
Sparser omits the 'GO:' prefix to GO ids in at least a majority of cases. This PR simply adds it on.


pagreene:
As far as I know, this is now ready to go!

PR #677, opened 2018-10-29
Fix some minor issues with the db client api.

pagreene:
This PR:
- Allows the user to not have an api key
- Replaces the join functionality for Windows compatibility
- Removes the API key from the logs.


pagreene:
This PR was failing due to an issue with the cbio client, apparently.

PR #685, opened 2018-10-30
Track evidence

pagreene:
This PR makes two changes:
- Adds a `text_references` attribute and parameter to `Evidence`, to carry more general text ref info besides pmids.
- Adds a `source_hash` attribute and `get_source_hash` method to store and generate a hash based off of the text and source api. This could have major degeneracies for content we get from pathway databases if/when there isn't a `source_id`, but I don't think we can avoid that.

### DIFF
--- a/indra/statements.py
+++ b/indra/statements.py
@@ -1036,6 +1036,9 @@ class Evidence(object):
     epistemics : dict
         A dictionary describing various forms of epistemic
         certainty associated with the statement.
+    text_references : dict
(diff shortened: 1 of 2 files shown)


bgyori:
Could we make it `text_refs` to rhyme with `db_refs`?

johnbachman:
Should we include the PMID in the hash if there is no evidence text?

bgyori:
I changed the dict slightly to use capitalized keys similar to db_refs and added a handler for setting text_refs to {} in case of loading from old pickles.

PR #690, opened 2018-11-02
Html Assembler

johnbachman:
Introduces a new assembler, the HtmlAssembler, that formats a set of statements into a easily readable HTML report including evidence text and other metadata. Also incorporates Javascript allowing users to send feedback on statements to the INDRA DB REST API.


pagreene:
```======================================================================
FAIL: indra.tests.test_html_assembler.test_assembler
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/sorgerlab/indra/indra/tests/test_html_assembler.py", line 36, in test_assembler
    assert isinstance(result, str)
AssertionError```

This shows up in the python 2.7 test. A bytes issue perhaps?

File: indra/assemblers/html/template.html [hunk 1]
```diff
@@ -9,7 +9,7 @@
     <meta name="author" content="">
     <link rel="icon" href="../../favicon.ico">
 
-    <title>Results Browser</title>
+    <title>INDRA DB Query Results</title>
```
bgyori: I think since this is a general assembler, it ought to be more agnostic to the source of Statements it displays so in that sense Results Browser was not bad. Maybe the title of the page could be an argument to the assembler or make_model to allow customization.

johnbachman:
👍

pagreene:
I think this was the longest, most collaborative branch in INDRA history?

johnbachman:
Don't know about the longest (implementing Python 2/3 compatibility was pretty long) but definitely the most collaborative!

kkaris:
It's really cool that we initialized it only some month ago and have first version application out already!

PR #697, opened 2018-11-07
Grounding map docs

steppi:
This PR adds Docstrings to functions in grounding_mapper.py that were previously missing. The function load_grounding_map has also been given an optional argument allowing users to specify the line separator for csv files containing grounding map info.


bgyori:
I made some more changes but there's still more to go, please continue with those.

bgyori:
I think it's done now, thanks @steppi!

PR #707, opened 2018-11-13
Finish curation

pagreene:
This should wrap up the INDRA side of the curation pipeline, by adding the capability to submit a curation of a raw or preassembled statement from the INDRA DB REST client.

### DIFF
--- a/indra/sources/indra_db_rest/client_api.py
+++ b/indra/sources/indra_db_rest/client_api.py
@@ -429,11 +429,12 @@ def get_statements_for_paper(ids, ev_limit=10, best_first=True, tries=2,
 
 
 def submit_curation(level, hash_val, tag, text, curator,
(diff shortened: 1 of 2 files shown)


indra/sources/indra_db_rest/client_api.py
```diff
@@ -428,12 +430,17 @@ def get_statements_for_paper(ids, ev_limit=10, best_first=True, tries=2,
     return stmts_from_json(stmts_json.values())
 
 
-def submit_curation(level, hash_val, tag, text, curator,
-                    source='indra_rest_client'):
+def submit_curation(hash_val, tag, text, curator, source='indra_rest_client',
+                    ev_hash=None, is_test=False):
```
bgyori: Let's add some docstrings to explain the arguments here, that would be really useful!

PR #710, opened 2018-11-14
Move s3 lookup method from indra_db to aws util.

pagreene:
This PR adds a tool to util.aws which scans for all matching prefixes and returns and NestedDict object containing a file-tree-like structure of the s3 content and it's keys.


pagreene:
The travis error is an http fluke, with uniprot in this case.

PR #720, opened 2018-11-28
Update HUME processor

qiuhaoling:
The changes in the pull request are:
1) Change the Extraction "type" of relations to "relation", and make "subtype" to be "causation", "precondition","catalyst", etc. This is to be consistent with the Eidos Wiki description of the JSON-LD format.

2) We added a few helper functions which can extract "place" and "time" arguments for events, in Indra's Hume readers. We are not able to create WorldContext though, given our limited understanding of INDRA code. 

For places, in order for multiple readers to agree, it looks like there should be a shared database of Geonames - as a temporal solution, we provide the canonical name strings for each geopolitical entities.

### DIFF
--- a/indra/sources/hume/processor.py
+++ b/indra/sources/hume/processor.py
@@ -60,7 +60,7 @@ class HumeJsonLdProcessor(object):
                                'mitigation': -1, 'prevention': -1}
 
         # Restrict to known relation types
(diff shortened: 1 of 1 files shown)


pagreene:
I'm going to close this PR and make a new one with my changes rebased on top of yours: https://github.com/sorgerlab/indra/pull/722

PR #722, opened 2018-11-28
Update HUME processor.

pagreene:
This PR adjusts to various changes in the HUME processor, and also incorporates the the temporal and spacial context that's available.

### DIFF
--- a/indra/tests/test_hume.py
+++ b/indra/tests/test_hume.py
@@ -63,7 +63,7 @@ def test_bbn_on_ben_paragraph():
 
 def test_large_bbn_corpus():
     bp = process_jsonld_file(os.path.join(path_this,
-                             'wm_m12.v6.full.v2.json-ld'))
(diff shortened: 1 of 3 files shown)


indra/tests/test_hume.py
```diff
@@ -53,9 +53,18 @@ def test_negated_effect():
...
+                                          'hackathon_test_paragraph.json-ld'))
     assert bp is not None
     print(bp.statements)
     stmt_dict = {s.get_hash(shallow=False): s for s in bp.statements}
     assert len(stmt_dict) == 3, len(stmt_dict)
+
+
+def test_large_bbn_corpus():
+    bp = process_jsonld_file(os.path.join(path_this,
+                             'wm_m12.v6.full.v2.json-ld'))
```
bgyori: We should use the forthcoming test file here that we can add to version control

PR #727, opened 2018-11-29
REST API developments

bgyori:
This PR extends the INDRA REST API to expose:
- interfaces to the Eidos, Hume and CWMS readers
- preassembly with a custom belief scorer
- ontology mapping
- scored grounding filtering
- filtering by belief

It also adds documentation and tests for the various end points.

The PR also removes deprecated code from the Hume processor, and fixes other bugs.


bgyori:
An important note is that I started updating the Swagger documentation to reflect these changes but that is not done yet and and I haven't included it in this PR.

PR #737, opened 2018-12-04
Don't add act() around bioprocesses when outputting to PyBEL

### LINKED ISSUE #601: Biological processes shouldn't have activities when output to PyBEL
The PybelAssembler just needs a bit more logic, since being a biological process in BEL means there isn't an activity nor abundance.

I'll make a PR for this with extra tests!

cthoyt:
Closes #601 

Includes a new unit test that shows the scenario. Before, it would output things like

```
p(HGNC:YGF) -> act(bp(GO:YFBP))
```

and now it will correctly output

```
p(HGNC:YGF) -> bp(GO:YFBP)
```

### DIFF
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -450,7 +450,10 @@ def _get_agent_node_no_bcs(agent):
 
     if variants:
         node_data = node_data.with_variants(variants)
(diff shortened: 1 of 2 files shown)


cthoyt:
Great, thanks @johnbachman!

PR #738, opened 2018-12-06
Changed loop to handle changed json structure

kkaris:
This PR resolves a bug where new json stucture broke the handling of json statements in `stmts_from_json()` in `indra/statements`.

Signed-off-by: kkaris <karis.klas@gmail.com>

### DIFF
--- a/indra/statements.py
+++ b/indra/statements.py
@@ -3549,9 +3549,9 @@ def stmts_from_json(json_in, on_missing_support='handle'):
 
     stmts = []
     uuid_dict = {}
-    for json_stmt in json_in:
+    for stmt_hash in json_in['statements']:
         try:
-            st = Statement._from_json(json_stmt)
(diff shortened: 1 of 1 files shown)


pagreene:
Aha. I don't think that's actually a bug. The format of the JSON from the rest API is simply a different beast than the assumed _list_ of json that this function expects.

pagreene:
It may be worth letting `stmts_from_json` handle both formats, though, given the likely increasing prevalence of the hash format; just check if the thing passed is a list or a dict, and if it is a dict, iterate over `values()`.

kkaris:
> It may be worth letting `stmts_from_json` handle both formats, though, given the likely increasing prevalence of the hash format; just check if the thing passed is a list or a dict, and if it is a dict, iterate over `values()`.

Ok, I can fix that and push

PR #749, opened 2018-12-13
Hide Statement JSON in CX networks

bgyori:
This PR changes the attribute in which the INDRA Statement JSON is added to CX networks to hide it from the interface while still leaving it part of the network. It also integrates styling as an option directly in the network upload function.


bgyori:
We should keep in mind that this may need to be modified too: https://github.com/sorgerlab/bioagents/blob/master/bioagents/qca/qca_module.py#L85

PR #750, opened 2018-12-16
Make statements a module and split into multiple files

bgyori:
This PR takes statements.py, moves it into a `statements` module, and splits parts of it into separate files, as follows:
- statements.py: Statement class definitions and helper functions operating on Statement classes
- agent.py: Agent class definition and all associated classes representing Agent states (ModCondition, etc.)
- concept.py: Concept class definition
- evidence.py: Evidence class definition
- context.py: Context class definitions (BioContext, WorldContext, etc.)
- io.py: Input/output functions and serialization
- resources.py: loading and interacting with resource files (e.g. cellular components)
- util.py: other shared helper functions

To make this change smoother, the `__all__` declaration in `statements.py` is preserved, and propagated into `__init__.py`, meaning that import paths in other modules and packages do not need to be updated.


bgyori:
One inconvenience is that pickle files made after this change won't open with older versions of INDRA as I just encountered:
```
ModuleNotFoundError: No module named 'indra.statements.statements'; 'indra.statements' is not a package
```

PR #752, opened 2018-12-17
Trips shared objects

steppi:
This PR modifies the trips processor so that it no longer creates shared evidence and agent objects between some statements extracted from the same text. Originally, for a sentence such as
"A binds B and C" the extracted statements Complex(A, B), Complex(A, C) would share evidence objects. The agent object for A would also be shared between the statements. The problem has been solved by making deep copies of agent and evidence objects for all statement instantiations that could possibly lead to shared objects. This route was taken instead of instantiating new agent and evidence objects for each statement instantiation because it was easier to implement.

### DIFF
--- a/indra/sources/trips/processor.py
+++ b/indra/sources/trips/processor.py
@@ -6,6 +6,7 @@ import logging
 import operator
 import itertools
 import collections
+from copy import deepcopy
 import xml.etree.ElementTree as ET
 from indra.util import read_unicode_csv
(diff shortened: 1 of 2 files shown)


steppi:
@bgyori That makes a lot more sense. Then we wouldn't have to add additional deep copies if we want new statement types. Do you have a similar suggestion for cases where we are trying to avoid shared evidence?

PR #755, opened 2018-12-18
Added mapping from PCID to PUBCHEM

kkaris:
This PR adds the capability to the Sparser processor to map PCID to PUBCHEM.


kkaris:
I tested it on a set of 500 agents and then produced URLs from the db ids. I tested some URLs at random and they seemed to work out.

kkaris:
Re-pushed after PyCharm edited code without notice

PR #766, opened 2019-01-07
Remove unused import of NiceCXNetwork

cthoyt:
Apologies if I've missed something, but it's not obvious why this is being imported. If indra is installed without ndex2, there are several modules that can't be imported without it crashing.

### DIFF
--- a/indra/databases/ndex_client.py
+++ b/indra/databases/ndex_client.py
@@ -7,10 +7,6 @@ import time
 import requests
 import logging
 import ndex2.client
-try:
-    from ndex2.nice_cx_network import NiceCXNetwork
-except ImportError:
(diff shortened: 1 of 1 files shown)


cthoyt:
I think this failed on something irrelevant, but it's a big problem reading through the logs. Might we update the build so it only shows output for tests that fail?

In the Python 2 build, it was something with the PMC client and in the Python 3 build, downloading the eidos.jar file stalled.

bgyori:
Interesting, there must have been a reason for having that import there but I can't recall what it was. I'll look into it a bit more but it seems like this change is good!

bgyori:
Note however, that this change won't change that "If indra is installed without ndex2, there are several modules that can't be imported without it crashing" because there is another import from ndex2 above, right?

PR #767, opened 2019-01-07
Update BEL processor

cthoyt:
I've got some new content that's possible to get into INDRA after `pip install hbp_knowledge` using the following code:

```python
from hbp_knowledge import repository
stmts = repository.get_indra_statements()
```

Also see: https://github.com/pharmacome/knowledge

However, it uses lots of exotic BEL and we didn't plan for all possibilities with the current BEL processor. I've made a couple changes such that this importer doesn't error. There's still lots of normalization work to be done, but at least it doesn't crash.


bgyori:
Thanks!

PR #768, opened 2019-01-07
Travis update

pagreene:
Update Travis to:
- Use nose-notify to send notifications when tests fail
- Draw resources from s3 instead of O2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
   - "2.7"
   - "3.5"
 script:
+  # Credit: https://graysonkoonce.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/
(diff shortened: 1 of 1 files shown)


bgyori:
Very cool! One important thing to fix: currently even if a test fails, the test command exits with 0 as follows:
```
FAILED (SKIP=5, failures=1)
The command "python -m nose_notify --slack_hook $SLACK_NOTIFY_HOOK --label "$TRAVIS_REPO_SLUG - $TRAVIS_BRANCH" indra -v -a $NOSEATTR --exclude='.*tees.*' --exclude='.*eidos.*' --exclude='.*isi.*' --with-coverage --cover-inclusive --cover-package=indra --with-doctest --with-doctest-ignore-unicode --with-timer --timer-top-n 10" exited with 0.
```
nose_notify somehow has to raise 1 instead of 0 if there is a failure to make sure Travis knows about it.

pagreene:
This issue was a trivial one with `nose_notify`, in which the failure was only propagated if a notification was sent. It should be fixed now. Ironically, that should make the Travis tests fail.

PR #772, opened 2019-01-11
Merge sofia processors

kkaris:
This PR merges the SofiaJsonProcessor into the SofiaProcessor. The bulk of the work is now done in the new methods `_process_event()` and `_build_stmts()`.


indra/sources/sofia/api.py
```diff
@@ -27,7 +28,10 @@ def process_table(fname):
         rel_sheet = book['Causal']
     event_sheet = book['Events']
     entities_sheet = book['Entities']
-    sp = SofiaProcessor(rel_sheet.rows, event_sheet.rows, entities_sheet.rows)
+    sp = SofiaProcessor(input_object=(rel_sheet.rows, event_sheet.rows,
+                                      entities_sheet.rows))
+    if not sp.statements:
+        return None
```
bgyori: To match the standard behavior of other processors, we shouldn't do this return None here. sp.statements should just be an empty list in case no statements were extracted.

A _different_ scenario in which case some serious issue happened while processing the input  (e.g. the input file doesn't exist or the web service call failed) it may be appropriate to return a None processor.

kkaris: Fixed it in latest commit.

bgyori:
I refactored the implementation to have the shared code in SofiaProcessor, and the format-specific preprocessing steps in two separate, derived classes. This way the format-specific code isn't mixed and doesn't need to be chosen based on the type of input object.

PR #776, opened 2019-01-15
Bayesian belief scorer and real-time curation service

bgyori:
This PR implements a new belief scorer class called BayesianScorer (which inherits from SimpleScorer) that is built around a set of curation counts of positive and negative examples from each source or subtype. The Scorer can also accept updates during runtime to the positive/negative counts to update its probability estimates.

The PR also adds a new tool, implemented as a Flask app which allows loading a corpus of Statements and accepting positive/negative (i.e. correct/incorrect) curations on Statements in the corpus. Depending on the input arguments, the tool can re-run belief calculation with the updated counts and return an updated set of belief scores for the corpus.

### DIFF
--- a/indra/belief/__init__.py
+++ b/indra/belief/__init__.py
@@ -100,6 +100,10 @@ class SimpleScorer(BeliefScorer):
     """
     def __init__(self, prior_probs=None, subtype_probs=None):
         self.prior_probs = load_default_probs()
+        self.subtype_probs = {}
+        self._update_probs(prior_probs, subtype_probs)
+
+    def update_probs(prior_probs=None, subtype_probs=None)
         if prior_probs:
             for key in ('rand', 'syst'):
                 self.prior_probs[key].update(prior_probs.get(key, {}))
@@ -107,7 +111,6 @@ class SimpleScorer(BeliefScorer):
             logger.debug("Prior probabilities for %s errors: %s"
                          % (err_type, source_dict))
         self.subtype_probs = subtype_probs
-        return
 
     def score_evidence_list(self, evidences):
         def _score(evidences):
--- a/indra/belief/__init__.py
+++ b/indra/belief/__init__.py
@@ -103,7 +103,7 @@ class SimpleScorer(BeliefScorer):
         self.subtype_probs = {}
         self._update_probs(prior_probs, subtype_probs)
 
-    def update_probs(prior_probs=None, subtype_probs=None)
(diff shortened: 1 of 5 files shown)


indra/belief/__init__.py
```diff
@@ -205,6 +209,86 @@ def check_prior_probs(self, statements):
...
+    Parameters
+    ----------
+    prior_counts : dict
+        A dictionary of counts of the form [pos, neg] for
+        each source.
+    subtype_counts : dict
+        A dictionary of counts of the form [pos, neg] for
+        each subtype within a source.
+    """
+    def __init__(self, prior_counts, subtype_counts):
```
cthoyt: I feel like this `__init__` should be able to take advantage of the `SimpleScorer`'s `__init__` via `super(BayesianScorer, self).__init__()`. It might require a bit of changing of the interface, but 3/5 of the lines here are the same as `SimpleScorer`

indra/belief/__init__.py
```diff
@@ -205,6 +209,86 @@ def check_prior_probs(self, statements):
...
+                # Skip if there are no actual counts
+                if n + p == 0:
+                    continue
+                if source not in subtype_probs:
+                    subtype_probs[source] = {}
+                subtype_probs[source][rule] = \
+                    1 - min((p / (n + p), 1-syst_error)) - syst_error
+        # Finally we propagate this into the full probability
+        # data structures of the parent class
+        super().update_probs(prior_probs, subtype_probs)
```
cthoyt: Watch out for Python3 only usage of `super()` :)

indra/belief/wm_scorer.py
```diff
@@ -29,6 +46,6 @@ def get_eidos_scorer():
     subtype_probs = {'eidos':
                      {k: 1.0-min(v, 0.95)-syst_error for k, v
                       in zip(table['RULE'], table['% correct'])}}
-
+    print(subtype_probs)
```
cthoyt: Switch to log?

indra/belief/__init__.py
bgyori: The meaningful shared part of the code is in the update_probs method which does call the same method of the parent class.

indra/belief/__init__.py
bgyori: Yes, I knew I had to go back and fix that before merging :)

indra/belief/wm_scorer.py
bgyori: Removed

PR #790, opened 2019-01-29
Live statement curation tests

bgyori:
Add some tests for the curation app


bgyori:
Will reopen in another form.

PR #791, opened 2019-02-06
Improve the sorting and presentation of results in HTML assembler.

pagreene:
Implement general purpose function for grouping and sorting a list of INDRA Statements, and apply to the HTML Assembler. This PR also makes some upgrades to the formatting of the HTML in the HTML assembler, taking advantage of the new groupings of statements.


bgyori:
Awesome!

PR #792, opened 2019-02-10
Use logging instead of print in clockit

cthoyt:


bgyori:
Great! I was about to make the same change a few days ago but forgot about it :)

pagreene:
I've been meaning to make the change for a couple months, but likewise kept forgetting. :)

PR #793, opened 2019-02-10
Fix bug in PybelAssembler for nodes that shouldn't have variants

cthoyt:
Better handling for abundances and biological processes that accidentally get variants added

### DIFF
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -448,8 +448,10 @@ def _get_agent_node_no_bcs(agent):
     for mut in agent.mutations:
         var = hgvs(mut.to_hgvs())
         variants.append(var)
-
-    if variants:
+    
+    if variants and not isinstance(node_data, CentralDogma):
+        logger.warning('Node should not have variants: %s, %s', node_data, variants)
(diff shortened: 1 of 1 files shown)


cthoyt:
Maybe I should add some tests for this...

indra/assemblers/pybel/assembler.py
```diff
@@ -447,8 +447,10 @@ def _get_agent_node_no_bcs(agent):
     for mut in agent.mutations:
         var = hgvs(mut.to_hgvs())
         variants.append(var)
-
-    if variants:
+    
+    if variants and not isinstance(node_data, CentralDogma):
```
bgyori: This is interesting - looks like a valid concern but at least in the case of some modifications, they could apply to other things too. What kind of examples did you run into where this happened?

cthoyt: Good question: I only found instances of GO, CHEBI, and PUBCHEM agents that had phosphorylation or acetylation modifications.

cthoyt: I didn't look into which GO term it was, though. That might have been interesting...

PR #794, opened 2019-02-10
Update PybelAssembler to use information from epistemics['direct'] 

### LINKED ISSUE #788: PybelAssembler does not make use of epistemics['direct'] when generating BEL
The PybelAssembler currently does not make any use of the evidence.epistemics['direct'] flag when generating BEL, instead generating only directlyIncreases and directlyDecreases relations by default. This should be updated to reflect information about the presence of a physical interaction in the edge when it is available.

@cthoyt Any appetite for tackling this one?

cthoyt:
Closes #788

### DIFF
--- a/indra/statements/statements.py
+++ b/indra/statements/statements.py
@@ -237,6 +237,7 @@ class Statement(object):
     supported_by : list of :py:class:`Statement`
         Statements supported by this statement.
     """
+
     _agent_order = NotImplemented
 
     def __init__(self, evidence=None, supports=None, supported_by=None):
@@ -560,6 +561,7 @@ class Modification(Statement):
     evidence : None or :py:class:`Evidence` or list of :py:class:`Evidence`
         Evidence objects in support of the modification.
     """
+
     _agent_order = ['enz', 'sub']
 
     def __init__(self, enz, sub, residue=None, position=None, evidence=None):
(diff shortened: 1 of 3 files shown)


cthoyt:
Should be good to go, pending Travis-CI tests!

indra/statements/statements.py
```diff
@@ -858,117 +858,94 @@ class Dephosphorylation(RemoveModification):
...
 
 
 class Methylation(AddModification):
     """Methylation modification."""
-    pass
 
 
 class Demethylation(RemoveModification):
     """Demethylation modification."""
-    pass
```
bgyori: In what way were these passes a problem here?

cthoyt: There's incorrect or wrong with using `pass` here, but it's only necessary inside a control structure where there are no other interpretable statements. Since there's a docstring, this class definition isn't empty and it doesn't need a pass. I think that having these short class names really shows the elegant of the Statement hierarchy, and this is a 1/3 improvement on code efficiency! 

I really shouldn't have put this in this PR though... Let me know and I will revert these changes.

Related: my favorite use of pass is:

```python
try:
    do_something_dangerous()
except Exception:
    pass
```

If you think that's funny, check out https://github.com/ajalt/fuckitpy for a good laugh

PR #795, opened 2019-02-10
Add Gene Assembler

cthoyt:
This assembler extracts a list of genes that have HGNC database references and can be used for:

1. Generating a gene list file, useful for Broad Institute's GSEA software
2. Generating an Ideogram visualization that shows what genes are represented and with what density. See example below generated using the `hbp_knowledge` graph: 

```python
import hbp_knowledge
from indra.assemblers.genes import GeneAssembler

statements = hbp_knowledge.repository.get_indra_statements()
ga = GeneAssembler(statements)
ga.make_model()
ga.save_model('statements_ideogram.html', format='ideogram')
```

screenshot 2019-02-10 18 42 14

To Do:

- [x] How should RefSeq data be stored? Should it live in the repository, and if so, how should it get updated? Otherwise I'll include a downloader for it.
- [x] What about chromosome numbers?

### DIFF
--- a/doc/modules/assemblers/gene_assembler.rst
+++ b/doc/modules/assemblers/gene_assembler.rst
@@ -1,0 +1,4 @@
+Gene Set Assembler (:py:mod:`indra.assemblers.genes.assembler`)
+===============================================================
+.. automodule:: indra.assemblers.genes.assembler
+    :members:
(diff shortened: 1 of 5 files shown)


cthoyt:
@dianakolusheva what do you think about using this in EMMAA? what would we need to generalize here to make it directly useful there?

dianakolusheva:
> @dianakolusheva what do you think about using this in EMMAA? what would we need to generalize here to make it directly useful there?

I think we could probably do one of the following:
- Generate visualization during assembly and store it as one of the downloadable export formats.
- Embed the html into EMMAA model dashboard along with other charts.
- Add an endpoint to open it up in a new page.

Will there be any issues if the gene list is very large?

PR #797, opened 2019-02-11
Integrating protmapper package for site mapping, UniProt and PSP

bgyori:
**This PR effectively makes INDRA incompatible with Python 2.x**

This PR contains many important changes with the main goal of using the now standalone package called `protmapper` (https://github.com/indralab/protmapper) for site mapping. This has the following implications:
- The `phosphosite_client` is completely removed and used via `protmapper` instead.
- Almost all functionalities of the `uniprot_client` are removed and moved over to the `protmapper`. However, for the time being the imports are implemented in a way that the API for using UniProt via INDRA does not change.
- The core mapping code in the SiteMapper is now done in `protmapper`, and the surrounding code is refactored (pretty significantly) to use the MappedSite object model instead of passing around `invalid_sites` lists.


bgyori:
For some reason Travis tests (even ones independent of the PR) are stalling on this branch. Based on some debugging, it doesn't seem to have to do with changes to .travis.yml or installing the protmapper.

bgyori:
For some mysterious reason, this commit 6168026a6384b4ad3c7fb4e0f3e3fa39a7cbf2fe which removes the UniProt resources and client causes unrelated nosetests on travis to stall.

bgyori:
The stall is not related to nose-notify, the same thing happens with basic nosetests.

bgyori:
The key was to add the protmapper downloads explicitly as ```python -m protmapper.resources```. Not totally sure why the downloads didn't work on the fly.

bgyori:
Before merging, it would be good to eliminate the import chain that requires `protmapper` when importing `indra.statements`.

PR #802, opened 2019-02-13
Create a simple processor for RLIMS-P results.

pagreene:
Create a processor for RLIMS-P json, and an API to use their webservice.


pagreene:
Travis tests are failing for unrelated reasons.

indra/tests/test_rlimsp.py
```diff
@@ -0,0 +1,30 @@
...
+    rp = rlimsp.process_from_webservice('PMC3717945', with_grounding=False)
+    assert len(rp.statements) == 33, len(rp.statements)
+
+
+def test_grounded_endpoint_with_pmids():
+    pmid_list = ['16403219', '22258404', '16961925', '22096607']
+    stmts = []
+    for pmid in pmid_list:
+        rp = rlimsp.process_from_webservice(pmid, id_type='pmid',
+                                            with_grounding=False)
```
johnbachman: Looks like this test is for the grounded endpoint but uses `with_grounding=False`

pagreene: Ah yes. The name did not change to keep up with reality.

johnbachman: Then probably also worth adding at least one test of the grounded endpoint as well

pagreene: The first test is of the grounded endpoint. Do you mean a grounded endpoint with pmids?

pagreene: I added some more tests.

johnbachman: Got it--didn't notice that with_grounding=True is the default in the first test

PR #813, opened 2019-02-26
Refactor out indra_db references.

pagreene:
Part of the INDRA reading system still bore some vestigial references to the database reading pipeline that are not needed for any other reading process. This PR refactors those out.

### DIFF
--- a/indra/tests/test_reading_files.py
+++ b/indra/tests/test_reading_files.py
@@ -1,0 +1,33 @@
+from os import path
+from indra.tools.reading.read_files import read_files, get_readers
+
+from nose.plugins.attrib import attr
+
+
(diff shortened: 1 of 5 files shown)


indra/tests/test_reading_files.py
```diff
@@ -0,0 +1,33 @@
+from os import path
+from indra.tools.reading.read_files import read_files, get_readers
+
+from nose.plugins.attrib import attr
+
+
+@attr('slow', 'nonpublic')
```
bgyori: We might want to add 'notravis' here because not all readers are available on Travis.

pagreene: Done.

PR #814, opened 2019-02-27
Ensure TEES directory only referred to when it exists

steppi:
This PR attempts to fix the still failing issue with TEES tests by ensuring that files from TEES are only referred to if TEES has been installed. TEES is only installed if RUN_SLOW is set to true, but TEES_SETTINGS was being set to a file within TEES even when RUN_SLOW is false.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ install:
         aws s3 cp s3://bigmech/travis/TEES.tar.bz2 . --no-sign-request --no-sign-request;
         tar xjf TEES.tar.bz2;
         mv TEES ~/TEES;
+        export TEES_SETTINGS=~/TEES/tees_local_settings.py
     fi
   # Install nose notify
   - mkdir $HOME/.nose_notify;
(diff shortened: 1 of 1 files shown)


bgyori:
I'm not totally sure how this solves the issue though I might be missing something. The error says:
```
ERROR: indra.tests.test_tees.test_evidence_pmid
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/sorgerlab/indra/indra/tests/test_tees.py", line 208, in test_evidence_pmid
    tp = api.process_text(s, pmid)
  File "/home/travis/build/sorgerlab/indra/indra/sources/tees/api.py", line 89, in process_text
    python2_path)
  File "/home/travis/build/sorgerlab/indra/indra/sources/tees/api.py", line 148, in run_on_text
    raise Exception('Provided TEES directory does not exist.')
Exception: Provided TEES directory does not exist.
-------------------- >> begin captured stdout << ---------------------
Found python 2 interpreter at /opt/pyenv/shims/python2.7
--------------------- >> end captured stdout << ----------------------```
which is running in the `slow` condition and complains about the TEES directory not existing. How does moving the TEES_SETTINGS assignment from being done under all conditions to only being done in the `slow` condition solve that?

steppi:
sorry, i must not have understood the issue. i’ll look into it carefully tomorrow morning

bgyori:
Now that I'm looking at the Travis config file, I think I figured it out, will push to this branch.

PR #826, opened 2019-03-08
Add handling of cbn jgf files

kkaris:
This PR adds handling of Causal Bio Networks database files of jgif format.

### DIFF
--- a/indra/sources/bel/api.py
+++ b/indra/sources/bel/api.py
@@ -257,4 +257,4 @@ def process_jgif_file(file_name):
         bp.statements.
     """
     with open(file_name, 'r') as jgf:
-        return process_pybel_graph(pybel.from_jgif(json.load(jgf)))
+        return process_pybel_graph(pybel.from_cbn_jgif(json.load(jgf)))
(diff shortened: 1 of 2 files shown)


bgyori:
Since the function is generic for reading JGIF files (not specific to CBN), I renamed it to reflect this and also added some dosctrings. But I am actually not even sure that adding such a function is necessary, since it is trivial to replicate by calling `pybel.from_jgif` and then calling the existing `process_pybel_graph` function. The actual added value in "handling of cbn jgif files" might be to add a client to access the CBN networks, but even that is not necessarily needed.

bgyori:
Another question is, whether we should be using `pybel.from_cbn_jgif` specifically for CBN networks:
https://pybel.readthedocs.io/en/latest/io.html#pybel.from_cbn_jgif

kkaris:
Commented on a PyBel issue about the unhandled annotations warnings: https://github.com/pybel/pybel/issues/266#issuecomment-471711951

bgyori:
Looks great!

bgyori:
it turns out the test was checking for the type of the first Statement returned but the order is non-deterministic, so the test failed in certain cases. I pushed a fix for this ind2db250b79d1d7376182569b6ac3469fb77b7160.

PR #827, opened 2019-03-11
Medscan fixes

pagreene:
Fix handling of entity name boundaries to be compatible with the case that there is no entity, but only a tag (TEXT only db_refs).

This PR now also implements the ability, via the `lazy` argument in the api endpoints, to merely create a generator of statements rather than extract the statements all at once. This should allow statements to be processed and dumped in batches more easily.

These changes have been stress tested on our entire corpus of Medscan reading outputs, which had been failing previously.

### DIFF
--- a/indra/sources/medscan/api.py
+++ b/indra/sources/medscan/api.py
@@ -81,7 +81,7 @@ def process_file_sorted_by_pmid(file_name):
     return s_dict
 
 
-def process_file(filename, num_documents=None, lazy=False):
+def process_file(filename, interval=None, lazy=False):
     """Process a CSXML file for its relevant information.
 
     Consider running the fix_csxml_character_encoding.py script in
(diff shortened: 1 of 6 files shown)


bgyori:
Looks good! The only odd feature is that a lot of the folder manipulation and file handling has been moved into the processor which normally only deals with format-specific processing for the given source and not file handling. Is this because it is only possible to implement the iterator feature if that code is in the processor and not the API?

pagreene:
The core issue is that the files are in the general sense too big to read and need to be streamed. I could probably refactor it out of the class, but it would be a bit tricky. Also, there is some special handling performed on the files which is specific to the content.

johnbachman:
Re: Ben's comment, the alternative would be to go back to the previous approach where the API basically merged the contents of multiple processors.

pagreene:
Re: John, I might be able to come up with something better now, but it would take some time and thought.

bgyori:
See #830 to get ChEBI names from IDs just note that the ID passed to get a name shouldn't contain the CHEBI: prefix.

pagreene:
I think this is finally ready to merge!

johnbachman:
Nice!

PR #828, opened 2019-03-12
Complex Equality

pagreene:
Allow two complexes with members ordered differently to be considered equal by sorting the members. To make this easier, this PR also adds a special method for complexes: `sorted_members`, which returns the members sorted by matches key.

### DIFF
--- a/indra/statements/statements.py
+++ b/indra/statements/statements.py
@@ -384,17 +384,28 @@ class Statement(object):
             return str(self).encode('utf-8')
 
     def equals(self, other):
+        return self._types_equals(other) \
+            and self._agents_equal(other) \
+            and self._evidence_equals(other)
+
+    def _types_equals(self, other):
         if stmt_type(self) != stmt_type(other):
             return False
+        return True
+
+    def _agents_equal(self, other):
         if len(self.agent_list()) == len(other.agent_list()):
             for s, o in zip(self.agent_list(), other.agent_list()):
                 if (s is None and o is not None) or \
-                   (s is not None and o is None):
+                        (s is not None and o is None):
                     return False
                 if s is not None and o is not None and not s.equals(o):
                     return False
         else:
             return False
+        return True
+
+    def _evidence_equals(self, other):
         if len(self.evidence) == len(other.evidence):
             for s, o in zip(self.evidence, other.evidence):
                 if not s.equals(o):
--- a/indra/statements/statements.py
+++ b/indra/statements/statements.py
@@ -384,16 +384,16 @@ class Statement(object):
             return str(self).encode('utf-8')
 
     def equals(self, other):
-        return self._types_equals(other) \
-            and self._agents_equal(other) \
-            and self._evidence_equals(other)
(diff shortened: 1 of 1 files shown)


johnbachman:
Looks good, but there should definitely be a regression test!

bgyori:
Actually, I think `equals` was implemented this way on purpose. This is because `equals` is meant to test for exact equality. Mechanistic equality is tested using `matches` which does account for the ordering of Agents in the complex. Is there some context in which `equals` needs to be used for testing mechanistic equality? Could it be replaced by `matches` instead?

pagreene:
This is for use in the medscan processor. We want to check of statements contain new information, which includes all db refs, hence matches doesn't work. However, a difference in ordering in a complex carries no meaning. In this case, "A and B  bind" gets extracted initially as Complex("A and B", "A and B"), where the product of all db ref combinations exists. The entity names get fixed by the id grounding, however you still get two statements for on pieces of evidence that are identical. We want to use equal to detect this case and skip the extra extractions.

pagreene:
This is for use in the medscan processor. We want to check of statements contain new information, which includes all db refs, hence matches doesn't work. However, a difference in ordering in a complex carries no meaning. In this case, "A and B  bind" gets extracted initially as Complex("A and B", "A and B"), where the product of all db ref combinations exists. The entity names get fixed by the id grounding, however you still get two statements for on pieces of evidence that are identical. We want to use equal to detect this case and skip the extra extractions.

bgyori:
I see, I understand the motivation but I don't fully understand this part: "We want to check of statements contain new information, which includes all db refs, hence matches doesn't work." - wouldn't the two redundant complexes that need to be eliminated have the same agents (including all db_refs) just in different order, and therefore would `match`? 

Here is an example where I believe I solved the same (or a very similar) issue in another processor: https://github.com/sorgerlab/indra/blob/master/indra/sources/biopax/processor.py#L71-L83. Here redundant statements that share the same evidence and also match (including complexes that have members in different orders) are eliminated.

pagreene:
The issue is that for every other kind of statement, equality is what we want, because we want to catch differences in ungrounded refs. I could simply add special handling for complexes, but it seemed that, given the order does not in general relate to the evidence, it shouldn't be relevant to equals.

bgyori:
I see, I think some custom logic in the processor would be the best solution in this case. I wouldn't want to change the general semantics of `equals` for this because it is currently consistent across Statement types and elements of Statements, for instance, consider an example of another set of lists where `equals` is implemented to match exactly, including ordering: https://github.com/sorgerlab/indra/blob/master/indra/statements/agent.py#L262-L275.

pagreene:
I can remove it. The net effect of this PR would then be to refactor the equals operation into smaller parts (has no effect, but could be useful) and removing the old stub overwrite of equals that didn't actually do anything.

That said, I think I will promote the internal methods for equals into public methods. Having the ability to check certain parts of equality will essentially provide the capability we need:
```s1.matches(s2) and s1.evidence_equals(s2)```

PR #829, opened 2019-03-12
Mesh client update

johnbachman:
As it turns out in certain cases (in particular Medscan) it is important to be able to query MeSH for terms and get back canonical IDs and names. This PR adds a feature to the MeSH client to retrieve IDs and names from the MeSH resource file (already in indra/resources) or connect to the MESH SPARQL endpoint for terms not found in the resource file.


bgyori:
👍

PR #833, opened 2019-03-13
Create a simple client to get CHEBI names from a rest api.

pagreene:
The name says it all. There are other things we could extract from this REST endpoint, if we wanted. Of the various possible choices, I found this to be the fastest to respond by far, and has the level of information we really need. It is also confirmed to get names that are not picked up by the TSV.

### DIFF
--- a/indra/databases/chebi_client.py
+++ b/indra/databases/chebi_client.py
@@ -158,7 +158,7 @@ def _read_relative_csv(rel_path):
     return csv_reader
 
 
-@lru_cache(maxsize=50)
+@lru_cache(maxsize=5000)
 def get_chebi_name_from_id_web(chebi_id):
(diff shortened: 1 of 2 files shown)


pagreene:
Should I use this as an optional back-up attempt when getting a chebi name from an ID?

indra/databases/chebi_client.py
```diff
@@ -148,6 +153,44 @@ def _read_relative_csv(rel_path):
     return csv_reader
 
 
+@lru_cache(maxsize=50)
```
johnbachman: Link this function to `get_chebi_name_from_id` so it is called when the ID is not found in the local resource file

johnbachman:
👍

PR #834, opened 2019-03-14
ChEBI names v3

### LINKED ISSUE #831: Some synonyms should be included as ChEBI names
The name resource file from ChEBI (names_3star.tsv) has a variety of name TYPEs including the simple NAME type which provides a single standard name for many compounds. 

There are also compounds (in what I found, specifically, drugs) that don't have a NAME listed at all, and instead have a number of SYNONYMs, BRAND NAMEs, and INNs (International Nonproprietary Name). In most cases, for our standard naming purposes, INNs seem to be the best fit, e.g., vemurafenib, tofacitinib, etc. (these were added in #830).

Unfortunately, there are yet other compounds that don't have NAMEs, INNs, or BRAND NAMEs at all, and just come with a list of SYNONYMs. Here is the set of entries for PLX-4720 in names_3star.tsv:
```
960476	90295	SYNONYM	ChemIDplus	N-[3-(5-chloro-1H-pyrrolo[2,3-b]pyridine-3-carbonyl)-2,4-difluorophenyl]propane-1-sulfonamide
960506	90295	SYNONYM	ChemIDplus	PLX 4720
960507	90295	SYNONYM	ChemIDplus	PLX4720
960514	90295	IUPAC NAME	IUPAC	N-{3-[(5-chloro-1H-pyrrolo[2,3-b]pyridin-3-yl)carbonyl]-2,4-difluorophenyl}propane-1-sulfonamide
```
Intriguingly, _none_ of these actually matches the name shown on the ChEBI website (https://www.ebi.ac.uk/chebi/searchId.do?chebiId=CHEBI:90295): "PLX-4720". One is left wondering, where do they get the single name displayed on the website as ChEBI Name?

In any case, to cover compounds like PLX-4720, we will need to extend chebi_names.tsv and implement some kind of prioritization in which a preferred single name is identified for each ChEBI ID that we then standardize to.

bgyori:
In another attempt to tackling this issue, I realized that using the `names` table was not the right choice, and instead, the `compounds` table is the right place to get standard names from. Constructing chebi_names.tsv is now combined with constructing chebi_to_primary.tsv in a single function, derived from the same resource file. It looks like this now contains 

As a sanity check, I sampled 100 random ChEBI IDs and checked the name returned by the web service vs what is in the updated chebi_names.tsv table. I actually found one inconsistency, which is pretty interesting:
```
CHEBI: 139561
web: 'archaeal dolichyl alpha-D-glucosyl phosphate(1-)'
table: 'archaeal dolichyl beta-D-glucosyl phosphate(1-)'
```
Otherwise we get the same results from both. The web service fallback can of course still remain in place in case an ID is missing from the resource file.

This also resolves #831 since the names reported to be missing there are now also covered.


pagreene:
This is great! Yesterday afternoon we actually ran into an issue with the web service cutting me off because they thought I was spam, so this is great news.

PR #851, opened 2019-04-05
Update lit clients

steppi:
This PR removes filtering of paragraphs from core extract_text functions within the literature clients as suggested by bgyori. This filtering is deft specific and should be handled within deft_tools instead. To facilitate this, the extract_text endpoints continue having the same behavior, but each calls a function that extracts paragraphs and then joins these paragraphs. Text processing functions for deft can now call this extract paragraphs function and then do their own filtering. Filtering is deft specific and may need to change in unforeseen ways. It is bad to have to touch the core extract_text functions for deft specific filtering. These functions are used widely throughout Indra.

### DIFF
--- a/indra/literature/elsevier_client.py
+++ b/indra/literature/elsevier_client.py
@@ -233,20 +233,29 @@ def get_article(doi, output_format='txt'):
     return xml_string
 
 
-def extract_text(xml_string, contains=None):
+def extract_text(xml_string):
     """Get text from the body of the given Elsevier xml."""
+    paragraphs = extract_paragraphs(xml_string)
+    if paragraphs:
+        return '/n'.join(paragraphs) + '/n'
+    else:
+        return None
+
+
+def extract_paragraphs(xml_string):
+    """Get paragraphs from the body of the given Elsevier xml."""
     assert isinstance(xml_string, str)
     xml_tree = ET.XML(xml_string.encode('utf-8'), parser=UTB())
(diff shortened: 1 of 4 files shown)


indra/literature/elsevier_client.py
```diff
@@ -233,20 +233,29 @@ def get_article(doi, output_format='txt'):
     return xml_string
 
 
-def extract_text(xml_string, contains=None):
+def extract_text(xml_string):
     """Get text from the body of the given Elsevier xml."""
+    paragraphs = extract_paragraphs(xml_string)
+    if paragraphs:
+        return '/n'.join(paragraphs) + '/n'
```
bgyori: This is probably meant to be \n, right?

indra/literature/pmc_client.py
```diff
@@ -122,6 +122,26 @@ def extract_text(xml_string, contains=None):
     str
         Extracted plaintext.
     """
+    paragraphs = extract_paragraphs(xml_string)
+    if paragraphs:
+        return '/n'.join(paragraphs) + '/n'
```
bgyori: Why are /n-s used here? The original code uses ' '.join(paragraphs).strip(), we should preserve that unless there is a good reason not to.

indra/literature/elsevier_client.py
steppi: Wow, that’s pretty boneheaded. Yeah, my goal was to produce identical results to the existing implementation,

PR #856, opened 2019-04-08
Events in Influences and standalone Events

bgyori:
This PR changes the representation of causal _Influence_ Statements to take _Event_s instead of _Concept_s as arguments. Events carry a single Concept (or Agent which is a subclass) argument and can represent _delta_s (polarity, adjectives) and _context_ (World or Bio with details like time, location, etc.). Importantly, Events can also be instantiated as "first-class" Statements and carry their own evidence, allowing the representation of Events that aren't arguments of Influences. 

With this in mind, this PR does the following:
- Add a new Event class in indra.statements.statements and adapt all the surrounding code in the module
- Finally (long overdue), Influence is no longer derived from IncreaseAmount, it is its own top-level Statement class
- Preassembly is done for Influences or standalone Events as normal no code in the preassembler needed to be changed (the implementation of refinements, contradictions, etc. is wrapped in these classes)
- Major overhaul of all causal reader processors (eidos, cwms, hume, sofia), especially Eidos, which is now set up to extract standalone Events in addition to Influences
- Update all the relevant assemblers (PySB, CX, CAG, English, etc.) to use the new representation
- Update all the tests to construct Influences and check them as needed after the change


bgyori:
Tests are passing! Possible further developments:
- `delta`, currently an attribute of Event could be represented as a class and allowed to have other subclasses, the current one perhaps being called "QualitativeDelta" or something like that. A class-based representation would also simplify some of the code where the delta dict structure needs to be manipulated at the Event level.
- Standalone Events could be extracted not only from Eidos but also other readers

PR #860, opened 2019-04-11
Ensure cython is installed before pyjnius

steppi:
This PR implements a temporary fix to get travis tests working again. For unknown reasons pyjnius installation was being attempted before cython was installed. Cython is now explictly installed in .travis.yml to ensure this doesn't happen. It remains to discover what caused the tests to begin failing.


bgyori:
I'm running another build, this time with caches deleted to see if this fix is needed or not, I'll post the results.

bgyori:
Looks like this is needed because something about Cython/jnius has changed!

steppi:
Woah, I wasn’t wrong. I guess my travis has a different cache.

PR #862, opened 2019-04-15
More tweaking to try to fix Travis

bgyori:
I'm not sure anymore which of these lines is necessary or sufficient but the latest builds seem to pass.


bgyori:
Based on the latest tests, these changes don't seem to be necessary so I'm closing this.

Issue #246, opened 2017-06-27
Integrate ChEBI entries into entity hierarchy


bgyori:
After some digging I think the best place to start would be this file: ftp://ftp.ebi.ac.uk/pub/databases/chebi/ontology/chebi_lite.owl
which has ontology relationships and a single standard name (label) for each entry. The only piece missing here is the mapping from secondary to primary IDs. A strategy that might work in practice is to assume that the primary ID will show up in addition to secondary IDs in the lists of groundings we get as noted in #245. If only secondary IDs appear without the corresponding primary ID then additional resources need to be integrated.

bgyori:
Fixed in #865

Issue #245, opened 2017-06-27
Ambiguous ChEBI entries from BioPAX


bgyori:
For instance, for CO2 we get
```
'CHEBI': ['13282', '13285', '46787', '48829', '77746', '13285', 
'78433', '13283', '35620', '23011', '38867', '3283', '23014', 
'75771', '23011', '75772', '64708', '13282', '77974', '76971', 
'13284', '13283', '48706', '13284', '48829', '78017', '3283', '16526', '76413']
```
Here a lot of the IDs are ChEBI secondary IDs that map to the same primary ID, e.g. 13282 and 13285 are secondary IDs that map to the same primary ID, 16526. Some of the IDs like 13285 are duplicated, this can be eliminated easily. Many of the IDs refer to super-categories like 64708 which is "one-carbon compound". 

Due to IDs referring to categories, different chemicals often share some ChEBI IDs for instance in
```
NADP+
'CHEBI': ['77746', '58945', '58349', '18009', '75772']
NADPH
'CHEBI': ['59724', '57783', '77746', '16474', '75772']
```
Here [75772](https://www.ebi.ac.uk/chebi/chebiOntology.do?chebiId=CHEBI:75772) is a very general super-category called "Saccharomyces cerevisiae metabolite". 77746 is the same for "human metabolite". 59724 is another general category called "ribonucleoside triphosphate oxoanion". 

This indicates that to find the specific groundings for chemicals, and for assembly purposes, we need to make use of the [ChEBI ontology](http://ftp.ebi.ac.uk/pub/databases/chebi/ontology/).

This issue is also related to #150.

bgyori:
Fixed in #865

Issue #150, opened 2017-01-18
Standardize chemical names


bgyori:
Currently chemicals (even if they are correctly grounded) are not name-standardized and therefore can appear under different names (e.g. PIP3, PIP-3) in assembled models. This would require some kinds of reverse lookup (either as web service or as resource file) from ChEBI IDs to names.

bgyori:
Fixed in #865

Issue #138, opened 2016-11-18
Ambiguous ChEBI mappings


bgyori:
Some ChEBI entries map to multiple PubChem and ChEBML entries. When loading these resource files into dictionaries, one entry is picked up and the others are ignored. Someone should look at these and see why these ambiguities exits and if there is a good way to prioritize which mapping to choose.

johnbachman:
Can you give some examples? I'm wondering if this is only for the ones that are for classes of things, like "Kinase inhibitors".

bgyori:
Below is a list of ambiguous entries. The first column is the ChEBI ID, the second and third are PUBCHEM IDs (note that in the current state of things, compound IDs and substance IDs can both appear here and are not differentiated in any way).
```
3504 87305645 96099266
3856 160711983 172120884
4946 160710300 171571832
5776 160711189 171571833
6698 136349241 172120885
10038 160711702 172120886
10353 160711511 223438276
10429 160709810 162012068
11515 8146716 93581525
15467 121269965 8143338
15469 85240320 99313653
15577 58106738 8145802
15701 123058728 160645907
16055 160709399 8145029
16126 135610377 24398008
16271 136349393 8143455
16870 8145747 92741042
17013 160962786 8145733
17038 57581474 8144710
17135 162169346 171571834
17197 11533126 162169359
17277 81058753 81058761
17392 123058730 160645906
17439 8144844 99437568
17631 111978144 8144072
17785 8145325 92741778
17810 85164689 92741314
17950 135610192 99313678
18347 135610263 8144529
18348 56464096 8145455
20833 223445492 252149670
22788 92729978 99313687
28066 104222399 8144404
28103 163835680 164174508
28107 10318900 121269854
28323 163835630 223438372
28353 163835685 223438374
28454 121269913 252149681
28610 164174758 223438381
28733 135668154 164174673
29503 163725485 163835620
30041 160709401 8145270
30042 160709400 8145269
31111 160711814 163626652
31998 26744233 56394948
32359 172121088 8147834
33196 14717811 49658847
33855 253662781 255509810
34778 160712527 223438442
36528 124403845 50139201
36702 124403633 92741551
36707 135610739 99437367
45996 223438479 8147972
46229 223438480 57581424
50043 135610928 49742745
50998 160644804 56394811
51013 56394810 57304837
52402 58106724 85164744
53028 85240324 99437564
53604 252162763 87246475
55376 87351936 99437544
57258 92741115 93581526
[...]

PR #865, opened 2019-04-17
Extend and improve grounding logic for assembly

### LINKED ISSUE #548: ValueErrors used in places where they cannot be handled by user
There really shouldn't be ValueErrors thrown in any of these 4 cases:
https://github.com/sorgerlab/indra/blob/master/indra/preassembler/grounding_mapper.py#L57-L85
since these are basically impossible to handle by the user who typically calls `GroundingMapper.map_agents`  or `assemble_corpus.map_grounding` which both iterate over a set of Statements.

bgyori:
This PR makes several important changes to handling grounding upon assembly in the bio domain. The first main improvement is that the PR enables entity matching for chemicals and biological processes using grounding (rather than name), and in addition, is able to assign standardized names to chemicals and biological processes (given that they are grounded to CHEBI, GO or MESH). The second main improvement is the inclusion of the ChEBI ontology into the entity hierarchy enabling hierarchical assembly for chemicals.
The following changes are important to note:
- entity_hierarchy.rdf now contains HGNC IDs instead of symbols to correspond to the internal representation of the HGNC namespace and eliminate the need for reverse lookups during runtime
- The default grounding map is loaded such that HGNC symbols (curated in FamPlex) are converted to HGNC IDs during load time eliminating the need for the special handling of mapped db_refs dictionaries and reverse lookups
- The Agent entity_matches_key now uses get_grounding and uses the returned value (either a DB NS/ID tuple or name) to construct the matches key. The side effect of this is that virtually all Statement hashes will change after this PR is merged.
- The get_grounding method does not do ID mapping between UP and HGNC anymore, this is the responsibility of input processors and preassembly components (by which I mean the GroundingMapper). It just goes by what is in the Agent's db_refs and prioritizes HGNC over UP in case of redundancy.
- The grounding mapper now standardizes IDs across PUBCHEM and ChEBI based on improved mappings between the two, including using InChiKeys. It also maps HMDB to ChEBI.
- The PR significantly restructures the ChEBI related resource files and updates all resource files to their newest state

Fixes #548, #150, #138, #107, #246, #245


bgyori:
Further improvements as follow-ups to this branch could be:
- Extend the MESH ID to name mapping resource file for more efficient name standardization
- Extend the standardize_db_refs method of the GroundingMapper (potentially as a new module) to do ID mappings beyond just HGNC and UP (e.g., GO <-> MESH, PUBCHEM <-> CHEBI, etc.)

bgyori:
The PR now also implements back and forth standardization between CHEBI and PUBCHEM IDs in the GroundingMapper.

bgyori:
I think I'm done with this!

PR #867, opened 2019-04-17
Implementing Association

dianakolusheva:
This PR extends the implementation of Association statement and does the following changes:
- Make sure Eidos processor extracts Associations (correlations) with Event arguments
- Extend Association class that inherits from Complex and takes a list of 2 Events as an argument
- Make modifications to English Assembler and Graph Assembler to work with Association statements
- Add and modify relevant tests


dianakolusheva:
This PR doesn't implement a test of Association statement's refinement_of, equals and matches methods. The test should be very close to Influence statement's test_influence_refinement_of. I am thinking of parameterizing the statement type and running the same test for both statements.

PR #876, opened 2019-04-30
Add translocation to list of special cases.

pagreene:
Fixes a bug where query strings generated for translocations didn't work.


pagreene:
@bgyori and @johnbachman, can you think of any other special cases that need to be added off the tops of your heads?

PR #885, opened 2019-05-23
Proposal: Sitemapper tweaks

pagreene:
This is what I implemented while trying to debug the sitemapper issues. Seems like it could be a useful functionality for future debugging.

While I was touching the file I also removed the python 2-3 compatibility code.


johnbachman:
Interesting--one thought though is that any errors occurring during sitemapping (from the protmapper) should be caught there and then encoded in the ERROR_CODE of the MappedSite object returned. So in principle after the recent tweaks to the protmapper the extra exception handling would not be required within the indra sitemapper. Of course, Murphy's Law and all that :)

pagreene:
Should we just close this one?

PR #887, opened 2019-05-29
Deft to Adeft

steppi:
This PR updates the grounding_mapper in response to deft's name changing to adeft. deft is changed to adeft in imports, variable names, and docstrings in which it appeared. The travis config has been updated to install adeft instead of deft.


bgyori:
Based on the tests, it looks like there is still a reference to `deft`, in `indra.tools.assemble_corpus`, and in the module `indra.literature.deft_util`.

steppi:
It should work now. Hope all the checks pass

bgyori:
:+1:

PR #892, opened 2019-06-04
Tag the EBS volumes.

pagreene:
When tagging batch instances, also tag the EBS volumes.

### DIFF
--- a/indra/util/aws.py
+++ b/indra/util/aws.py
@@ -102,7 +102,7 @@ def tag_instance(instance_id, **tags):
         vols = instance.volumes.all()
         for page in vols.pages():
             for vol in page:
-                vol.create_tages(Tags=tag_list)
+                vol.create_tags(Tags=tag_list)
     else:
         logger.info('No new tags from: %s' % str(tags))
     return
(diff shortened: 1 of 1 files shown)


indra/util/aws.py
```diff
@@ -99,6 +99,10 @@ def tag_instance(instance_id, **tags):
         logger.info('Adding project tags "%s" to instance %s'
                     % (filtered_tags, instance_id))
         instance.create_tags(Tags=tag_list)
+        vols = instance.volumes.all()
+        for page in vols.pages():
+            for vol in page:
+                vol.create_tages(Tags=tag_list)
```
bgyori: Isn't this supposed to be `create_tags`?

pagreene: Indeed. For some reason I make that typo a lot.

indra/util/aws.py
```diff
@@ -99,6 +99,10 @@ def tag_instance(instance_id, **tags):
         logger.info('Adding project tags "%s" to instance %s'
                     % (filtered_tags, instance_id))
         instance.create_tags(Tags=tag_list)
+        vols = instance.volumes.all()
+        for page in vols.pages():
```
jmuhlich: Can't this be simplified to `for vol in vols` and omit the explicit page traversal? Or even combine it with the previous line, `for vol in instance.volumes.all():`

pagreene: Possibly. This is just how I got it to work in ipython. `vols` didn't seem to be iterable, although I suppose I didn't try just iterating over it. But the only thing that returned an apparent iterable was `vols.pages()`. It seemed obtuse, but it worked. :man_shrugging:

PR #901, opened 2019-06-26
Add publication_date in literature/pubmed_client

RohitChattopadhyay:
Added **publication_date** in function **_get_journal_info** in file [literature/pubmed_client.py](https://github.com/sorgerlab/indra/blob/master/indra/literature/pubmed_client.py).

Thanks

PR #902, opened 2019-06-26
Add publication_date in literature/pubmed_client.py

RohitChattopadhyay:
Related to #901 
Added **publication_date** in function **_get_journal_info** .
@johnbachman , please let me know if I have done it properly or anything else is required.

Thanks

### DIFF
--- a/indra/literature/pubmed_client.py
+++ b/indra/literature/pubmed_client.py
@@ -267,19 +267,22 @@ def _get_journal_info(medline_citation, get_issns_from_nlm):
     return {'journal_title': journal_title, 'journal_abbrev': journal_abbrev,
             'issn_list': issn_list, 'journal_nlm_id': nlm_id}
 
+
 def _get_pubmed_publication_date(pubmed_data):
     # Look for pubmed as PubStatus in PubmedPubDate
-    pubmed_pub_date =  pubmed_data.find('./History/PubMedPubDate[@PubStatus="pubmed"]')
+    pubmed_pub_date = \
+                pubmed_data.find('./History/PubMedPubDate[@PubStatus="pubmed"]')
     # Get date elements from extracted pubmed_pub_date element
     year = _find_elem_text(pubmed_pub_date, 'Year')
     month = _find_elem_text(pubmed_pub_date, 'Month')
     day = _find_elem_text(pubmed_pub_date, 'Day')
     # Build and return result
     return {
-        "year"  : None if (year  is None) else int(year),
+        "year"  : None if (year is None) else int(year),
         "month" : None if (month is None) else int(month),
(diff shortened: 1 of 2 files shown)


johnbachman:
Thank you! I pushed some minor commits to the branch to fix a few things in the code as I was taking a look at the PR. The most important thing is that I added a test, `test_get_pub_date` in `tests/test_pubmed_client.py` and in doing so I noticed a that the date information returned by PubMed is often inconsistent. For example, for PMID 27123883 the month comes back as "08" and the day is None (not present in the XML); for PMID 27121204 the month comes back as "May"; for PMID 27115606 only the year is included and both month and day are None.

@bgyori and I discussed and decided that probably the best thing to do in this case is to normalize the date information coming from the PubMed XML into integers for year, month, and day, with `None` used for missing information. I've written `test_get_pub_date` to indicate how the information should come back from the client--can you normalize the date information coming back from PubMed so that this passes? Thanks!

RohitChattopadhyay:
Thanks for the review.
I went through the documentation which explains the XML Schema ([link](https://www.nlm.nih.gov/bsd/licensee/elements_descriptions.html#pubdate)). It mentions, the record may contain **<MedlineDate>**, so we need to support that as well? 
For conversion of Month, it is mentioned in the documentation, 3 character abbreviation is used, so shall I use a dictionary consisting of abbreviated month names and corresponding number?

RohitChattopadhyay:
@johnbachman , requesting review for https://github.com/sorgerlab/indra/pull/902/commits/c218e6dd1a6bf5c6455acd27dfefe87120ca491b
Requesting you to add tests for **MedlineDate**.
Thanks

RohitChattopadhyay:
@johnbachman requesting review for https://github.com/sorgerlab/indra/pull/902/commits/862316d988b88930629b5da037b59a1978566b40

Following the advice of my mentor @cannin, I am extracting the publication date from **/History/PubMedPubDate** instead of **JournalIssue/PubDate**. The advantage of this method is that it assures the presence and consistency of the Date components and eliminates the requirement of `MedlineDate`.  

[Gist](https://gist.github.com/RohitChattopadhyay/09811f951dc41d156f442bda00430c02) having an example of two articles, one has MedlineDate and other has Date components. You can see that `History/PubMedPubDate[@PubStatus="pubmed"]` is giving the desired date components.  
[Schema Explanation for PubMedPubDate](https://www.nlm.nih.gov/bsd/licensee/elements_descriptions.html#pubmedpubdate)  

Requesting your views on this.   
Also, the Travis CI build is failing due to these changes, requesting you to make necessary changes in the values of the assert statements.  
Thanks

johnbachman:
Looks good!

PR #906, opened 2019-07-09
Update INDRA for new REST API

pagreene:
Relatively minor changes to keep up-to-date with the indra DB REST service.


indra/databases/hgnc_client.py
```diff
@@ -1,22 +1,19 @@
...
-except ImportError:
-    from functools32 import lru_cache
-from indra.util import read_unicode_csv, UnicodeXMLTreeBuilder as UTB
+from functools import lru_cache
 
+from indra.util import read_unicode_csv, UnicodeXMLTreeBuilder as UTB
 
 logger = logging.getLogger(__name__)
 
+try:
```
bgyori: @pagreene could you clarify in what context this is needed? Is the Lambda API so resource constrained that installing requests is an issue?

pagreene: Actually, the HGNC client is no longer used in the REST Service. I tried to purge the requests dependency from the Lambda to avoid what seemed to be a versioning issue (with urllib actually) when processing credentials.

So, in summary, this isn't really needed at this point, and could probably be dropped. On the other hand, the HGNC client mostly works without web lookups, so having the module doesn't need to rely on requests.

bgyori: Alright, in case it doesn't matter for the REST service, we could go back to just the plain `import requests` without try/except - since it is a basic INDRA dependency so there shouldn't be a scenario where it isn't available normally.

pagreene: Done

PR #911, opened 2019-07-16
Developments in the PyBEL processor

bgyori:
This PR improves grounding in the PyBEL processor by adding GO, MESH, PUBCHEM, etc. name spaces, and improves the handling of ones already used. It also fixes some implementations that were assuming a specific form of BEL content.


bgyori:
There are two interesting changes in the test results: we get 20 Statements instead of 26 from the test CBN networks due to 6 of these: 
```
indra.sources.bel.processor: INFO: Could not find GO ID for cell proliferation
```
It turns out `cell proliferation` isn't really a valid GO label.

Similarly, in `test_activation_bioprocess`, we try to work with a bioprocess called `apoptosis`, but that's not a valid GO label (the closest one is `apoptotic process` which I changed it to).

Now that we handle GO groundings, it also means that invalid entries will be thrown away, instead of propagated as ungrounded. We should look into whether this has any negative effects on processing the large corpus. Two possible solutions: look at the accompanying belns file for the BEL graph, if available, and attempt to find mappings through that, or propagate some of these entities as ungrounded (currently the statements they are involved in are dropped).

bgyori:
I compared processing large_corpus.bel on master and on this branch. There are a total of 946 Statements lost due to the issue described above, and 872 of those is specifically due to `cell proliferation`. Some of the remaining ones are for `necrosis` (the official label is `necrotic cell death`), `glucose transport`, `synaptic transmission`, `death`, `invasive growth`, and a few others.

Issue #255, opened 2017-07-07
Grounding of micro-RNAs


bgyori:
We actually get a pretty good amount of micro-RNAs from BioPAX and reading. The reading systems pick them up as named entities but don't have the resources to ground them. BioPAX grounds micro-RNAs to "mirtarbase", e.g. hsa-miR-26b-5p is grounded to "MIRT006307". This is an interaction database, not a micro-RNA database (like miRBase) and so we would probably need to map these IDs. So we could:
- introduce a new db_refs entry key for micro-RNAs, probably miRBase
- map mirtarbase BioPAX grounding to miRBase db_refs
- integrate micro-RNA resources into reading systems or grounding-map them ourselves

cthoyt:
Heads up to Sumit (@maestroe) since this is relevant to our discussion for next week

cthoyt:
@lingling93 did a lot of work parsing miRBase in https://github.com/bio2bel/mirbase. Specifically, downloading the terminology (which includes names, identifiers, synonyms, and xrefs) can be found at https://github.com/bio2bel/mirbase/blob/master/src/bio2bel_mirbase/parser.py

There's something else at https://github.com/bio2bel/mirtarbase that also gets all of the data from miRTarBase, but not as nicely in a JSON object as this

cthoyt:
After the last two PR's get reviewed, we might consider calling this done :)

PR #912, opened 2019-07-17
Add miRBase client

cthoyt:
This PR adds a client for miRBase, which is a nomenclature for micro-RNAs. It adds a function for downloading and structuring the resource as a TSV that has cross-references to ncbigene and other organism specific databases. The client has lookup for terms by miRBase name, identifier, or HGNC. This PR also updates the PyBEL processor to handle nodes named with miRBase.

A few questions about adding this client:

- ~Should I add the resources in a versioned way? The .gitignore seems to reject stuff in the resources folder~ I added it with 8211c4a because I couldn't pass the tests and it looks like everything else is already added
- Should I update the hgnc_client to have a `is_mirna()` function like the `is_kinase()` and related?
- What else is necessary to finish #255?


cthoyt:
Looks like the failing tests are from code that has to do with PubMed client... Unrelated to this PR

indra/resources/update_resources.py
```diff
@@ -662,11 +664,13 @@ def _process_mirbase_file(lines):
...
             if not element.startswith('DR'):
                 continue
-            db, identifier, name = [e.strip() for e in element[len('DR'):].lstrip().split(';')]
+            db, identifier, name = [e.strip() for e in \
+                                       element[len('DR'):].lstrip().split(';')]
             yield mirbase_id, mirbase_name, db, identifier, name.rstrip('.')
 
 
 if __name__ == '__main__':
+    """
```
cthoyt: Hey, what's going on here?

johnbachman: D'oh, commented stuff out for testing and then forgot to uncomment 🤦‍♂

johnbachman: Fixed

PR #913, opened 2019-07-19
IndraNet

dianakolusheva:
This PR introduces two new classes: 
- IndraNet - network representation of INDRA Statements. This class inherits from networkx.MultiDiGraph class and uses INDRA Statements specific data as nodes and edges attributes. 
- IndraNetAssembler - extracts the necessary metadata from INDRA Statements into a DataFrame and calls IndraNet classmethods to assemble the network.

Next steps include adding more methods to IndraNet to convert it to other graph types (signed, digraph, etc.)


indra/assemblers/indranet/assembler.py
```diff
@@ -0,0 +1,147 @@
+import logging
+import pandas as pd
+from .net import IndraNet
+from indra.statements import *
+from itertools import permutations
+from collections import OrderedDict
+
+
+logger = logging.getLogger(__name__)
+NS_PRIORITY_LIST = ('FPLX', 'HGNC', 'GO', 'MESH', 'HMDB', 'CHEBI', 'PUBCHEM')
```
bgyori: We could follow the priority used in Agent.get_grounding here, which would be: FPLX, HGNC, UP, CHEBI, GO, MESH. Other groundings like HMDB and PUBCHEM can be in arbitrary order at the end of the list. In practice, this may make a difference in at least two cases: chemicals that have both MESH and CHEBI grounding, there we want to prioritize CHEBI, and UP-grounded proteins that don't have human gene grounding.

kkaris:
Hmm, maybe we should change to have edges keyed by an index instead of hashes? When working with the assembled graph, it's hard to know a priori what hash you need to use to just get an edge between two nodes.

Using indices, one could also imagine a case where the edges can be sorted on and placed in some given order, which would not be possible with hashes.

Any thoughts @bgyori, @dianakolusheva (and @johnbachman if you're available)?

dianakolusheva:
I agree. We can move hashes to edge data dictionary.

indra/assemblers/indranet/assembler.py
kkaris: @dianakolusheva fixed it in https://github.com/sorgerlab/indra/pull/913/commits/e9f64393873dc274725dc331976fb33664fcf0d9

kkaris:
The docstrings in the [IndraNet class](https://github.com/sorgerlab/indra/pull/913/files#diff-fe7f1869f6d37611c5769fc537d3d14e) (indra/assemblers/indranet/net.py) refuse to show up in the sphinx html build. Trying to add the module explicitly using available [options](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) does only give partial results. The only way to get the full docstrings with proper formatting so far have been to replace `class IndraNet(MultiDiGraph)` with `class IndraNet(object)` or `class IndraNet()`. Maybe we should file an issue (both here and at sphinx) and just merge?

kkaris:
> The docstrings in the [IndraNet class](https://github.com/sorgerlab/indra/pull/913/files#diff-fe7f1869f6d37611c5769fc537d3d14e) (indra/assemblers/indranet/net.py) refuse to show up in the sphinx html build. Trying to add the module explicitly using available [options](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) does only give partial results. The only way to get the full docstrings with proper formatting so far have been to replace `class IndraNet(MultiDiGraph)` with `class IndraNet(object)` or `class IndraNet()`. Maybe we should file an issue (both here and at sphinx) and just merge?

I just double checked with a minimum working example and confirmed that the docstrings do show up when `MultiDiGraph` is not inherited.

PR #914, opened 2019-07-21
Add missing staticmethod annotation to Delta.from_json()

cthoyt:


bgyori:
Thanks! Neither staticmethod, nor classmethod was actually necessary there since Delta.from_json(...) worked without either of those (see test at https://github.com/sorgerlab/indra/blob/master/indra/tests/test_delta.py#L56). For consistency with other from_json methods, I changed it to classmethod.

PR #915, opened 2019-07-21
Fix typo in BioContext.__eq__()

cthoyt:
It looks like it was always comparing its own attrs to its own attrs, so this function would always return true when comparing any BioContext to any object at all.


bgyori:
Good catch! Fortunately this wasn't used yet so you found it before it caused any mysterious bugs :)

PR #916, opened 2019-07-21
Add hypothesis and context.species information in index_cards

RohitChattopadhyay:
@bgyori , @johnbachman  , I have added `hypothesis_information` and `context.species` following the email conversation.
As the `evidence` is of `list` type, hence to maintain the order in both lists ie `evidence` and `hypothesis_information` ( or `context` ) I have used the `get_evidence_text()` method.
Requesting feedback regarding the stability of my implementation.
Also, it would be helpful if you could add some tests for this.
Please note that `None` ( `null` in JSON ) is returned if any of these fields are absent.

Thanks

### DIFF
--- a/indra/assemblers/index_card/assembler.py
+++ b/indra/assemblers/index_card/assembler.py
@@ -64,6 +64,8 @@ class IndexCardAssembler(object):
                 continue
             if card is not None:
                 card.card['meta'] = {'id': stmt.uuid, 'belief': stmt.belief}
+                card.card['interaction']['hypothesis_information'] = _get_hypothesis_information(stmt)
+                card.card['interaction']['context'] = _get_context_information(stmt)
                 if self.pmc_override is not None:
                     card.card['pmc_id'] = self.pmc_override
                 else:
@@ -105,6 +107,7 @@ class IndexCard(object):
             'submitter': None,
             'interaction': {
                 'negative_information': False,
+                'hypothesis_information' : None,
                 'interaction_type': None,
                 'participant_a': {
                     'entity_type': None,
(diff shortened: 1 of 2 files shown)


bgyori:
Great, thanks! I think you are on the right track with trying to follow the order of the evidence sentences, but, it might make sense to refactor things a bit so that one doesn't have to match the ordering by looking for equivalences in evidence text. Let me look into it and see if I can come up with a solution, I'll add it as an extension to your PR.

bgyori:
Alright, I refactored the assembler to make getting context and hypothesis status much simpler, and improved some other stuff. @RohitChattopadhyay could you please test with some relevant statements if the cards come out to way you expect them to? If it works well, I'll merge this.

indra/assemblers/index_card/assembler.py
```diff
@@ -397,30 +407,44 @@ def get_pmc_id(stmt):
...
+            else:
+                ev_txts.append('%s%s' % (prefix, ev.text))
+
+            if ev.context and ev.context.species:
+                species = ev.context.species
+                obj = {}
+                obj['name'] = species.name
+                obj['taxonomy'] = species.db_refs.get('TAXONOMY') \
+                    if species.db_refs is not None else None
+                contexts.append(obj)
```
RohitChattopadhyay:
```suggestion
            contexts.append(obj)
```

indra/assemblers/index_card/assembler.py
```diff
@@ -397,30 +407,44 @@ def get_pmc_id(stmt):
...
-    return ev_txts
+                ev_txts.append(
+                    '%sEvidence text not available for %s entry: %s' %
+                    (prefix, ev.source_api, ev.source_id))
+            else:
+                ev_txts.append('%s%s' % (prefix, ev.text))
+
+            if ev.context and ev.context.species:
+                species = ev.context.species
+                obj = {}
```
RohitChattopadhyay:
```suggestion
```

indra/assemblers/index_card/assembler.py
```diff
@@ -397,30 +407,44 @@ def get_pmc_id(stmt):
...
-            if txt not in ev_txts:
-                ev_txts.append(txt)
-    return ev_txts
+                ev_txts.append(
+                    '%sEvidence text not available for %s entry: %s' %
+                    (prefix, ev.source_api, ev.source_id))
+            else:
+                ev_txts.append('%s%s' % (prefix, ev.text))
+
+            if ev.context and ev.context.species:
```
RohitChattopadhyay:
```suggestion
            obj = {}
            if ev.context and ev.context.species:
```

RohitChattopadhyay:
Thanks, I tested the changes using PUBMED ID `25695838`, and following is the output:
```
{
 "pmc_id": "PMC4346620",
 "submitter": "indra",
 "interaction": {
  "negative_information": false,
  "hypothesis_information": [
   null
  ],
  "interaction_type": "increases",
  "participant_a": {
   "entity_text": [
    "LAG3"
   ],
   "identifier": "UNIPROT:LAG3_HUMAN",
   "entity_type": "protein"
  },
  "participant_b": {
   "entity_text": [
    "TGF-\u03b23"
   ],
   "identifier": "",
   "entity_type": "protein"
  },
  "context": []
 },
 "meta": {
  "id": "70502ca4-bcb1-4528-814a-b7716fe5297f",
  "belief": 1
 },
 "evidence": [
  "We herein demonstrate that LAG3 (+) Treg produce high amounts of TGF-\u03b2 3 in an Egr2- and Fas dependent manner."
 ]
}
```
Is the empty list in `context` desired? I feel that this will make it difficult to get context information of an evidence as the order of evidence and context will not be the same.
To maintain the order, I think making the following changes at [assembler.py#L427](https://github.com/sorgerlab/indra/pull/916/files#diff-3e285462075de4c9023e09690198122eR427) should serve the purpose.
```
            obj = {}
            if ev.context and ev.context.species:
                species = ev.context.species
                obj['name'] = species.name
                obj['taxonomy'] = species.db_refs.get('TAXONOMY') \
                    if species.db_refs is not None else None
            contexts.append(obj if bool(obj) else None)
```
I have used the fact that empty dictionary is boolean `False` in python.
I have suggested these changes for your convenience.

Thanks

bgyori:
Yes, that's a good point, I didn't notice that the empty values weren't added. I changed is such that the length of context will be the same as the length of evidence witch identical order.

bgyori:
I'll merge this now, if there's anything else we need to add or change, we can start a new branch.

PR #919, opened 2019-07-21
Add miRTarBase processor

cthoyt:
References #255 

This doesn't update the BioPAX processor, but at least it gives a sure-fire set of miRNA-target interactions. Right now it's lacking in the species specificity or the propagation of the experimental conditions. The database even has confidence information associated with each, so this could be used to calculate BELIEF scores.

Question: how should this be tested?

### DIFF
--- a/doc/modules/sources/index.rst
+++ b/doc/modules/sources/index.rst
@@ -41,6 +41,7 @@ Standard Molecular Pathway Databases
    biogrid/index
    hprd/index
    trrust
+   mirtarbase/index
 
 Custom Knowledge Bases
 ----------------------
--- a/doc/modules/sources/mirtarbase/index.rst
+++ b/doc/modules/sources/mirtarbase/index.rst
@@ -1,0 +1,4 @@
+miRTarBase (:py:mod:`indra.sources.mirtarbase`)
+===============================================
+.. automodule:: indra.sources.mirtarbase
+    :members:
--- a/indra/sources/mirtarbase.py
+++ b/indra/sources/mirtarbase.py
@@ -2,6 +2,8 @@
 
 """An INDRA processor for miRTarBase."""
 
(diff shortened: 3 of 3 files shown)


bgyori:
Hey, this is interesting but we already get good interactions via BioPAX from this resource, so I'm not sure this is critical.

cthoyt:
I might have misunderstood but I thought somebody (you or @johnbachman) mentioned that there was room for improvement on the BioPAX import of miRTarBase. Perhaps it was just due to the lack of grounding.

bgyori:
No, they actually come out reliably grounded and name standardized, we haven't noticed any issues with it. Looking back, #255 is perhaps not totally clear, what I meant there is that INDRA doesn't (didn't) have a standardized way to deal with micro-RNA grounding, not that the BioPAX source produced ungrounded agents.

cthoyt:
I think I was actually remembering our discussion about the WikiPathways and KEGG pathways coming from BioPAX, then I mentioned WikiPathways has some good coverage of miRNA interactions. Sorry for the confusion - #255 is pretty clear (except for its completion conditions)

bgyori:
Given the new BioPAX implementation in INDRA, I'll take another look at this to see if processing miRTarBase directly has advantages - in many cases we found that going to the source directly is better so I suspect the same might be the case here. More updates coming soon.

bgyori:
I actually never gave up on this one and just a couple of months ago, did another test of using this processor (with some modifications I have on a local branch) and comparing to what we get from the latest state of the BioPAX processor via Pathway Commons. I actually found that the diff was non-trivial and it would have taken much more work to drill down to figure out what differs exactly and why, and which interpretation is better. I think if any specific use case comes up where we have to rely on microRNA target information, we can revisit this and see how it works.

PR #923, opened 2019-07-24
Update Elsevier client

dianakolusheva:
This PR updates Elsevier client `get_piis_for_date` according to new Elsevier ScienceDIrect search API format. P.S. This API is also used by `get_dois` function which also needs to be updated.


dianakolusheva:
I made a general function `search_science_direct` that can take any field name and only return values from that field. `get_piis_for_date` and `get_dois` are now calling it with "pii" and "doi" arguments respectively.

PR #926, opened 2019-07-31
Refactor grounding mapper and add misgrounding resources

bgyori:
This PR adds a new component to grounding mapping: one that allows removing known incorrect grounding without otherwise discarding the statement it was in. With this addition, the GroundingMapper has the following components:
- A grounding map that asserts manually curated grounding for a set of strings
- An ignores list that contains degenerate/irrelevant entity strings which, when present result in the statement being filtered out
- A misgrounding map that asserts a set of incorrect groundings
- Adeft, which is invoked if an ambiguous string is encountered for which a disambiguation model exists

Given all these moving parts, it made sense to restructure the code. The PR makes the following changes:
- Move grounding resource files into `indra/resources/grounding` (except grounding_map.csv, which comes from famplex and I chose to leave it in `indra/resources/famplex`)
- Make the ignores list just a list, instead of formatting it like the grounding map, and apply it separately from the grounding map (for historical reasons, ignores were loaded into the grounding map resulting in some complicated code).
- Split `indra.preassembler.grounding_mapper.py` into a module called `indra.preassembler.grounding_mapper` with the following files
  - mapper.py: the core GroundingMapper code
  - analysis.py: various (relatively old) functions to analyze the grounding of statements, and generate benchmarks
  - adeft.py: functions used when invoking Adeft for disambiguation
- Rename the GroundingMapper's map_agents method to map_stmts, which to me is a much more intuitive name since it takes a list of Statements as input
- Make the instantiation of the GroundingMapper more intuitive by avoiding having to separately import the class and a grounding map (and ignores, an misgroundings), and rather instantiating it will default parameters if nothing is provided.

Further potential changes:
- The db_refs standardization method and the name standardization method (both static) are now pretty cleanly separated from the mapping functionality of the GroundingMapper. It may make sense to refactor these into module-level functions, or wrap them into other classes separate from the GroundingMapper.


johnbachman:
Looks good

PR #933, opened 2019-08-06
Add MeSH supplementary concepts

### LINKED ISSUE #932: Add MeSH supplementary concepts to INDRA resources
Currently, the update_resources script doesn't take into account the MeSH supplementary concepts. These can be found at `ftp://nlmpubs.nlm.nih.gov/online/mesh/2018/xmlmesh/supp2018.gz` and have (mostly) the same structure as the actual descriptors file.

Will send a PR!

cthoyt:
Closes #932


indra/resources/update_resources.py
```diff
@@ -602,33 +604,60 @@ def update_lincs_proteins():
...
+    for concept in concepts:
+        # We only look at the preferred concept here
+        if concept.attrib['PreferredConceptYN'] == 'Y':
+            terms = concept.findall('TermList/Term')
+            for term in terms:
+                term_name = term.find('String').text
+                if term_name != name:
+                    all_term_names.append(term_name)
+    # Append a list of term names separated by pipes to the table
+    term_name_str = '|'.join(sorted(all_term_names))
```
cthoyt: Note I added a sort here to make sure it's always consistent

indra/resources/update_resources.py
```diff
@@ -602,33 +604,60 @@ def update_lincs_proteins():
...
+    supp_rows = []
+    for record in tqdm(supp_et.iterfind('SupplementalRecord')):
+        uid = record.find('SupplementalRecordUI').text
+        name = record.find('SupplementalRecordName/String').text
+        term_name_str = _get_term_name_str(record, name)
+        supp_rows.append((uid, name, term_name_str))
+
     fname = os.path.join(path, 'mesh_id_label_mappings.tsv')
-    write_unicode_csv(fname, rows, delimiter='\t')
+    write_unicode_csv(fname, itt.chain(rows, sorted(supp_rows)), delimiter='\t')
```
cthoyt: The descriptors seem to come out of XML in order by identifier but the supplement did not

indra/resources/update_resources.py
```diff
@@ -6,6 +6,8 @@
 import rdflib
 import logging
 import requests
+from tqdm import tqdm
```
cthoyt: While maintaining the christmas tree, I had to do a `from ...` import before a normal one. Is this okay? What's proper christmas tree etiquette?

bgyori:
Thanks, looks good! But the resource file itself is missing from the PR somehow. Using tqdm is a bit questionable because it's a new dependency (i.e., without further steps, the readthedocs build would fail when trying to import this file) while not being critical - I'd prefer not to use it. The Christmas tree convention is fine to not follow :)

cthoyt:
Somehow the resource didn't get updated! I added it and removed tqdm

cthoyt:
Also, FYI this seems to solve the problem we were having with Gilda looking up xrefs to MeSH

bgyori:
Oh wow, I didn't realize the supplementary concepts would increase the file by 10-fold! This is a little bit inconvenient because it puts the distribution size of INDRA well above the 60MB limit for PyPI, though with some tweaking I might be able to fit it below the limit.

bgyori:
The other issue that I jus realized is that these supplementary concepts are highly redundant with other resources like ChEBI. Unless we also put in place a whole other set of things to comprehensively map these to ChEBI, it would break the expected standardization of lots of entities. Given that none of INDRA's input sources currently ground to supplementary concepts that I'm aware of, I wonder if this is necessary.

cthoyt:
Yeah it seems there's quite a bit in there, and mapping MeSH to chemicals is notoriously difficult. I've had to go through PubChem then SMILES in the past to do this.

One option might be to split the `mesh_id_label_mappings.tsv` file to have the supplementary information in a different file, that the user could presumably build if they wanted that stuff (then it wouldn't have to be distributed with INDRA). Again, one use case would be for Gilda, but not necessarily for INDRA.

bgyori:
Yes, that's a good solution! I looked into it a bit deeper, and actually found that we have the same redundancy issue even with the original set of MeSH terms. This is perhaps more of a Gilda issue than an INDRA issue so we can discuss how to handle this over there. Overall, I think this PR is good and I'll merge it soon!

cthoyt:
@bgyori great. So it's up to you if you want to remove the TSV for the supplementary id to name mappings before you make the merge

PR #943, opened 2019-08-14
ModelChecker refactor

dianakolusheva:
This PR refactors the ModelChecker class and extends its functionality to support 3 new model formats. Implementation includes a parent ModelChecker class and 4 child classes: PysbModelChecker (replaces previous ModelChecker), SignedGraphModelChecker, UnsignedModelChecker, and PybelModelChecker.
Each child class is responsible for preprocessing the given model (using `get_graph`) and the test statements (using `process_statement` and `process_subject`) into a standard form used by parent class for path finding. Note that the search is now conducted on graphs with signed nodes instead of signed edges. This allowed replacing custom `_find_sources_with_paths` with Networkx `shortest_simple_paths` function and simplifying the code. Also note that only standard breadth first search is now supported and sampling paths is not supported.


indra/tests/test_model_checker.py
```diff
@@ -1120,57 +1125,57 @@ def test_check_transphosphorylation():
...
-    # +/- 1
-    # This data should ensure that the path through B should be more highly
-    # ranked than the path through C
-    data = {b_phos: 1, c_phos: -1, d_phos: 1}
-    paths = results[0][1].paths
-    scored_paths = mc.score_paths(paths, data)
-    assert scored_paths[0][0] == p1
-    assert scored_paths[1][0] == p2
-    assert scored_paths[0][1] > scored_paths[1][1]
+# NOTE: skip path score tests for now
```
bgyori: Let's use @unittest.skip('Skip path score tests for now') instead of commenting here.

indra/tests/test_model_checker.py
```diff
@@ -1231,181 +1236,182 @@ def test_prune_influence_map_degrade_bind():
...
-                     ('MAPK3_phospho_phosphorylation_JUN_phospho', 1),
-                     ('JUN_phospho_p_obs', 1))] == 49, path_ctr
-    assert path_ctr[(('MAP2K1_phosphorylation_MAPK1_S218', 1),
-                     ('MAPK1_phosphoS218_phosphorylation_JUN_phospho', 1),
-                     ('JUN_phospho_p_obs', 1))] == 31, path_ctr
-    assert path_ctr[(('MAP2K1_phosphorylation_MAPK1_S222', 1),
-                     ('MAPK1_phosphoS222_phosphorylation_JUN_phospho', 1),
-                     ('JUN_phospho_p_obs', 1))] == 20, path_ctr
+# NOTE: skip sampling tests for now
+# def test_weighted_sampling1():
```
bgyori: Here also, let's use @unittest.skip instead of commenting

bgyori:
Very nice! In terms of the big picture, I think it might make sense to turn model_checker.py into a submodule called model_checker, in which `__init__.py` (or another file that init imports) would contain the parent class and any shared functions, and the individual model checker classes would be in their own files (e.g., `indra/explanation/model_checker/pybel.py`). `__init__.py` could then import all the model checker classes so that imports like `from indra.explanation.model_checker import ModelChecker` would work).

indra/tests/test_model_checker.py
dianakolusheva: Fixed this

indra/explanation/model_checker/model_checker.py
```diff
@@ -0,0 +1,490 @@
+from __future__ import print_function, unicode_literals, absolute_import
+from builtins import dict, str
+from future.utils import python_2_unicode_compatible
```
johnbachman: I'll remove some of this Python 2-specific boilerplate.

indra/explanation/model_checker/model_checker.py
```diff
@@ -0,0 +1,490 @@
...
+        Whether to use breadth-first search or weighted sampling to
+        generate paths. Default is False (breadth-first search).
+    seed : int
+        Random seed for sampling (optional, default is None).
+
+    Attributes
+    ----------
+    graph : nx.Digraph
+        A DiGraph with signed nodes to find paths in.
+    """
```
johnbachman: In the case of simple DiGraphs, it's not a requirement that the nodes be signed, correct?

PR #944, opened 2019-08-14
Custom flattened belief and weight

kkaris:
This PR adds options allowing for custom mapping functions for the belief and weight of the flattened edges in a flattened IndraNet graph given just the graph and the edge attributes (including the list of statements making up the flattened edge) of the flattened graph.

_Note that for flattened weight, the default behavior is that no weight is calculated._

In addition to allowing for custom belief calculation, two pre-defined functions for mapping belief to the flattened edge are provided:
- Simple scorer: This uses the belief score of a mock Statement created with all the sources in the flattened edge as evidence. _This is currently the default behavior for flattened belief._
- Complementary belief: This mapping calculates a complementary belief score using the belief scores of the individual statements making up the flattened edge according to <a href="https://www.codecogs.com/eqnedit.php?latex=\inline&space;\text{flattened&space;belief}&space;=&space;1&space;-&space;\prod_i&space;\left(1&space;-&space;\text{belief}_i&space;\right)" target="_blank"></a>

### DIFF
--- a/indra/assemblers/indranet/net.py
+++ b/indra/assemblers/indranet/net.py
@@ -132,6 +132,12 @@ class IndraNet(nx.MultiDiGraph):
         """
         G = nx.DiGraph()
         for u, v, data in self.edges(data=True):
+            # Add nodes and their attributes
+            if u not in G.nodes:
+                G.add_node(u, **self.nodes[u])
+            if v not in G.nodes:
+                G.add_node(v, **self.nodes[v])
+            # Add edges and their attributes
             if G.has_edge(u, v):
                 G[u][v]['statements'].append(data)
             else:
--- a/indra/assemblers/indranet/net.py
+++ b/indra/assemblers/indranet/net.py
@@ -170,9 +170,9 @@ class IndraNet(nx.MultiDiGraph):
 
     @staticmethod
     def _update_edge_belief(G, flattening_method):
-        """G must be or be a child of an nx.Graph object. If provided,
-        flattening_method must be a function taking at least a graph G and
-        return a number (the new belief for the flattened edge).
+        """G must be or be a child of an nx.Graph object. If provided as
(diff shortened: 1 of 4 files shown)


dianakolusheva:
We should probably also update IndraNetAssembler.make_model - when choosing  signed and digraph options user needs to provide flattening method

indra/assemblers/indranet/net.py
```diff
@@ -128,6 +145,12 @@ def to_signed_graph(self, sign_dict):
             as positive edges and Inhibition and DecreaseAmount are added as
             negative edges, but a user can pass any other Statement types in
             a dictionary.
+        flattening_method : str|function
```
johnbachman: Give examples of what the valid string values are.

kkaris: Fixed in https://github.com/sorgerlab/indra/pull/944/commits/c192fa3f6427261c6a7adec3074dd5e609a9a12a

kkaris:
We should probably merge https://github.com/indralab/indra_db/pull/67 first. It adds the mapping of the source names used in `_simple_scorer_update` in `indra/assemblers/indranet/net.py` https://github.com/sorgerlab/indra/pull/944/files#diff-fe7f1869f6d37611c5769fc537d3d14eR255-R265

kkaris:
> Add arguments to IndraNetAssembler make_model function allowing custom flattening method to be specified.
> 
> Also move default_sign_dict out the IndraNetAssembler class and into a top level variable in assembler.py; and have this be the default argument for IndraNet.to_signed_graph when no arg is provided.

Fixed in last four commits (from https://github.com/sorgerlab/indra/pull/944/commits/d21ab5c709648e339cf702daea7dce2741d30baa to https://github.com/sorgerlab/indra/pull/944/commits/8be69b3963a45b6791bc843a7b84f94551e43d6e)

PR #957, opened 2019-08-29
Refactor html assembler templates

pagreene:
Refactor and rearrange how the jinja templates in the HTML assembler are used, replacing a bespoke loader with a standard FileSystemLoader.


pagreene:
The current test failure is irrelevant to the PR (something about memory allocation and Eidos).

PR #959, opened 2019-08-30
Fix indentation error

dianakolusheva:
This PR fixes a small bug in `stmts_from_pybel_path`.


bgyori:
Thanks @dianakolusheva, as a follow up, we should later add a couple of tests for cases like #958 and this PR.

PR #961, opened 2019-09-03
Test reporting

dianakolusheva:
This PR expands ModelChecker tests to test reporting functions (returning statements from path) and fixes a bug in Pybel reporting function. This should fix the issue of duplicated stmts in Pybel path.


bgyori:
Great, it's good that we have lots of tests now!

PR #975, opened 2019-09-27
Fix handling of the persist option, allowing for waiting.

pagreene:
The meaning of "persist" had come to have something of a dual meaning, and in the previous PR morphed someone from the original intent, causing a test to fail in certain cases (depending on how long the query took). The behavior should now be consistent with the original intent.


pagreene:
The test failures in this case appear to be pysb related an irrelevant to the PR.

PR #977, opened 2019-10-04
Let the API determine if it is done.

pagreene:
Tiny change which makes the paging over queries more robust in light of content redactions.


pagreene:
The corresponding change to indra_db has already been pushed.

PR #978, opened 2019-10-06
Normalize equivalent and opposite concepts with respect to an ontology

bgyori:
This PR adds new preassembly functionalities to process and handle `is_equal` and `is_opposite` relationships between entries of an ontology. The WM ontology is updated to reflect these changes. Two new optional normalization methods in the preassembler change grounding to standardize to a single grounding from a group of equivalent ones or mutually opposite ones. If a grounding is replaced by its opposite, the corresponding statement's polarity is flipped. 
This PR is conditional on https://github.com/WorldModelers/Ontologies/pull/25/ being fixed and merged.


bgyori:
The ever shifting ontology has yet again been updated and now a number of test conditions are not valid anymore (e.g., the duplicate flooding entry has been eliminated from the ontology so can't be used as a test condition).

PR #980, opened 2019-10-08
Update Eidos's processor.py

phosseini:
Removing duration. This fixes `KeyError: 'duration'`


bgyori:
Based on the issue https://github.com/clulab/eidos/issues/686, this error is simply due to that fact that you need to update your local version of INDRA. The current implementation of _get_duration calculates duration rather than attempting to get it from the Eidos JSON-LD, therefore this PR should not be applied.

PR #982, opened 2019-10-09
Updates to BEL API

bgyori:
This PR implements a long overdue face lift for the BEL API. Here is a list of the main changes:
- Two new API functions are added: `process_large_corpus` and `process_small_corpus` that fetch the corresponding Selventa corpora from https://github.com/cthoyt/selventa-knowledge, and process them. This replaces the previous mode of operation in which these corpora were processed from local BEL script files.
- Removed the large and small corpus BEL Script files from `data` since these are now available elsewhere, as noted in the previous bullet point. 
- The `process_pybel_neighborhood` function, when called with default parameters, behaves as before, but now has more options on what network to process in which format.
- A new function is exposed called `process_pybel_network`, which allows allows processing a given network file/URL with a given format. I don't consider this to be a critical API function to have (because one could just call a more specific function like `process_json_file` directly) but it is basically a freebie to expose given `process_pybel_neighborhood`, which I didn't want to remove.
- The `process_belrdf` function is not fully removed just in case it is ever needed again, but now comes with a deprecation warning and a warning about rdflib versions.
- The rdflib==4.2.1 restriction is removed so later versions of rdflib will be installed by default. This makes #666 redundant so we can close that. 
- The BEL/RDF tests are removed since the corresponding functionalities are deprecated, and the tests wouldn't work well with rdflib>4.2.1 anyway.
- All BEL/RDF files from the `data` folder are removed, though, for now, I left all the corresponding BEL Script files there (noting that they are probably not needed anymore).

FYI @cthoyt


indra/sources/bel/processor.py
```diff
@@ -786,9 +784,12 @@ def _get_translocation_target(node_modifier_data):
     to_loc_ns = to_loc_info.get(pc.NAMESPACE)
     to_loc_name = to_loc_info.get(pc.NAME)
     # Only use GO Cellular Component location names
-    if to_loc_ns != 'GOCC' or not to_loc_name:
+    if to_loc_ns not in ('GOCC', 'GOCCID') or not to_loc_name:
```
cthoyt: Should this also accept `'GO'`?

bgyori: I guess that wouldn't hurt, I haven't come across that yet but I'll add it

bgyori:
Thanks! I agree about the pickle issue, I guess there just isn't a perfect alternative currently, since processing the BEL script file (though smaller in size) takes a long time, whereas the pickle contains a ready to use BEL Graph. Maybe a gzipped JSON could work?

I think it's generally enough if INDRA provides the functions to process BEL content, not necessarily to fetch them from specific repos (that can be done in a separate script that imports indra). The Selventa corpora were an exception in my mind since we've worked with them so extensively, it seemed like a good idea to add functions to fetch them specifically.

PR #984, opened 2019-10-10
Move files out of the data folder and remove it

bgyori:
This PR eliminates the `data` folder (a folder that is outside the `indra` package but within the repo) by moving each file to a more appropriate location, or removing it entirely. The following changes were made:
- MFS_MSB_*_gene_list.csv: moved to `indra_apps/fallahi_eval` where they are used
- RAS_neighborhood*.bel: moved to `indra_apps/ras_demo` where they are used
- full_abstract1.bel: this is a Selventa example file and is not relevant anymore, removed
- biopax_test.owl: moved into `indra/tests` which is a more appropriate location for it
- ras_pathway_proteins.csv: the only file left here that is actually used in many different places, even recently, and therefore I moved it into `indra/resources`, and updated all references to it in `indra_apps`


johnbachman:
Looks good!

PR #985, opened 2019-10-10
Load BEL corpora from json.gz instead of pickle over network

bgyori:
This PR switches from loading pickles to loading gzipped JSONs of PyBEL graphs.

### DIFF
--- a/indra/sources/bel/api.py
+++ b/indra/sources/bel/api.py
@@ -1,4 +1,4 @@
-import os
+import zlib
 import json
 import pybel
 import pickle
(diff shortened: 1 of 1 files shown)


cthoyt:
I just had the thought now, should you pin the commit for the pickle? Or do you think it's okay to always grab the most recent?

bgyori:
I implemented it to just load the latest master intentionally - in case an older specific version is needed, the `process_pybel_network` function can be called with the appropriate URL to achieve that.

PR #1001, opened 2019-11-12
Add ISI Reader

pagreene:
This PR adds a Reader class for ISI, supporting bulk reading both locally and on a special docker image. This PR also includes significant renovation of the `read_files` CLI and underlying functionality.

The PR also:
- significantly restructures the `readers.py` file into a python module with submodules for each reader.
- allows for different readers to use different job definitions on AWS Batch
- slightly changes how Elsevier text is extracted by replacing excess whitespace with single spaces, so ISI can read the content effectively


pagreene:
For some reason tests are failing with RLIMSP errors, which _seem_ to be unrelated.

indra/tools/reading/read_files.py
```diff
@@ -1,28 +1,46 @@
...
+from os import path, listdir
 
-logger = logging.getLogger('indra.tools.reading.read_files')
-
+from indra.tools.reading.readers.core import dump_readings
 from indra.tools.reading.util.script_tools import get_parser
-from indra.tools.reading.readers import _get_dir, get_reader_classes, Content
+from indra.tools.reading.readers import get_dir, get_reader_classes, Content
+
+logger = logging.getLogger(__name__)
```
bgyori: The logger path was set explicitly here because this script will typically be run as `__main__`

PR #1008, opened 2019-11-25
Path stmts

dianakolusheva:
This PR introduces new object PybelEdge used to represent 'hasVariant' and 'hasComponent' edges of Pybel model. Function `belgraph_to_signed_graph` (and method `get_graph` calling it in PybelModelchecker) is parameterized to separately include/exclude hasVariant/hasComponent edges and their reversed versions. EnglishAssembler is updated to handle PybelEdges.

### DIFF
--- a/indra/assemblers/english/assembler.py
+++ b/indra/assemblers/english/assembler.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 from builtins import dict, str
 import logging
 import indra.statements as ist
+from indra.explanation.reporting import PybelEdge
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


indra/explanation/model_checker/pybel.py
```diff
@@ -27,14 +27,20 @@ def __init__(self, model, statements=None, do_sampling=False, seed=None):
         super().__init__(model, statements, do_sampling, seed)
         self.model_agents = self._get_model_agents()
 
-    def get_graph(self):
+    def get_graph(self, include_variants=False, symmetric_variant_links=False,
+                  include_components=True, symmetric_component_links=True):
```
dianakolusheva: By default hasComponent edges are included in both directions and hasVariant edges are not included at all in PybelModelChecker

indra/assemblers/english/assembler.py
```diff
@@ -75,6 +76,8 @@ def make_model(self):
                 stmt_strs.append(_assemble_gap(stmt))
             elif isinstance(stmt, ist.Conversion):
                 stmt_strs.append(_assemble_conversion(stmt))
+            elif isinstance(stmt, PybelEdge):
+                stmt_strs.append(_assemble_pybel_edge(stmt))
```
bgyori: I am not convinced that handling PybelEdges here is the best option. Would it be possible to iterate over edges in a path in the reporting module, and deciding if the EnglishAssembler or a separate PyBEL-edge to English converter should be called for the given edge?

dianakolusheva: That makes sense. I will create a separate function converting PybelEdge to English in reporting module. For now the only place it's going to be called is in EMMAA, so it can be easily handled there.

Issue #177, opened 2017-02-24
Extract specific phosphosite binding information from Phospho.ELM


jmuhlich:
Phospho.ELM is full of detailed information like "EGFR phospho-Y1092 binds the SH2 domain of GRB2". This could help us build a better picture of when certain binding events should be exclusive/competitive as well as properly handle multiple binding (e.g. GRB2 also binds EGFR phospho-1172).

PR #1009, opened 2019-11-25
Phosphoelm processor

kkaris:
This PR introduces the phosphoelm processor that adds phosphorylations from the phospho.ELM database http://phospho.elm.eu.org/dataset.html


indra/sources/phosphoelm/api.py
```diff
@@ -0,0 +1,107 @@
+import csv
+import logging
+import requests
+from unidecode import unidecode
+
+from .processor import PhosphoElmProcessor
+
+try:
+    from bs4 import BeautifulSoup
+    has_soup = False
```
bgyori: This should probably be `has_soup = True`

indra/sources/phosphoelm/api.py
```diff
@@ -0,0 +1,107 @@
...
+except ImportError:
+    BeautifulSoup = None
+    has_soup = False
+
+logger = logging.getLogger(__file__)
```
bgyori: This should be `logging.getLogger(__name__)`

indra/sources/phosphoelm/api.py
```diff
@@ -0,0 +1,107 @@
...
+
+    fname : str
+        File path to the phospho.ELM file dump. If none is provided,
+        the file will be downloaded from S3.
+    delimiter : str
+        The delimiter to use for csv.reader
+
+    Returns
+    -------
+    ppp : indra.sources.phosphoELM.PhosphoElmProcessor
```
bgyori: The above line should just be `indra.sources.phosphoelm.PhosphoElmProcessor` (to fix that capitalization, also, the ppp : part is not needed, and it doesn't match the variable name).

indra/sources/phosphoelm/api.py
```diff
@@ -0,0 +1,107 @@
...
+    if fname is None:
+        s3 = _get_s3_client()
+        s3_obj = s3.get_object(Bucket=s3_bucket, Key=ppelm_s3_key)
+        csv_reader = csv.reader(
+            s3_obj['Body'].read().decode('utf8').splitlines(True),
+            delimiter='\t'
+        )
+    else:
+        with open(fname, 'r') as f:
+            csv_reader = csv.reader(f.readlines(), delimiter=delimiter)
```
bgyori: Usually one can just do `csv.reader(f, ...)` is the `readlines()` part needed for some reason?

indra/sources/phosphoelm/api.py
```diff
@@ -0,0 +1,107 @@
...
+    ppelm_json = _get_json_from_entry_rows(csv_reader)
+    return PhosphoElmProcessor(file_dump_json=ppelm_json)
+
+
+def _get_json_from_entry_rows(row_iter):
+    """Loop body to generate a json friendly structure"""
+    ppelm_json = []
+    columns = next(row_iter)
+    for entry in row_iter:
+        row_dict = {columns[n]: entry[n]
```
bgyori: A bit better as `row_dict = {c: e for c, e in zip(columns, entry)}`

indra/sources/phosphoelm/api.py
```diff
@@ -0,0 +1,107 @@
...
+                    kinase_table.find_all('th'))
+    kinase_entries = {}
+    for tr in kinase_table.find('tbody').find_all('tr'):
+        entry = dict(zip(headers, (unidecode(td.get_text()).strip() for td
+                                   in tr.find_all('td'))))
+        kinase_entries[entry['Name']] = entry
+    return kinase_entries
+
+
+def _get_s3_client(unsigned=True):
```
bgyori: Looks like `_get_s3_client` is a more general implementation than what the processor needs and is not specific to it. Would make sense to move it into `indra.util.aws` and import it from there.

bgyori: Also note that boto3 is not a standard INDRA dependency so maybe it should be added to the phosphoelm extras list in setup.py.

indra/sources/phosphoelm/processor.py
```diff
@@ -0,0 +1,139 @@
+import logging
+import requests
+
+from indra.statements import Phosphorylation, Evidence, Agent
+
+gilda_url = 'http://grounding.indra.bio/ground'
+logger = logging.getLogger(__file__)
```
bgyori: Should be `__name__`

indra/sources/phosphoelm/processor.py
```diff
@@ -0,0 +1,139 @@
...
+        return entity_str, db_ns, db_id
+    else:
+        if res.status_code != 200:
+            logger.warning('Gilda service responded with status code %d' %
+                           res.status_code)
+        return entity_str, None, None
+
+
+class PhosphoElmProcessor(object):
+    def __init__(self, file_dump_json=None, keep_empty=False):
```
bgyori: Can file_dump_json really be None? What does the processor do if that's the case?

indra/sources/phosphoelm/processor.py
```diff
@@ -0,0 +1,139 @@
...
+        """The PhosphoElmProcessor processes data dumps from the phospho.ELM
+        database. See http://phospho.elm.eu.org/dataset.html
+
+        file_dump_json : list(dict)
+            JSON compatible list of entries from a phospho.ELM data dump
+        keep_empty : bool
+            If true, also create statements when upstream kinases in
+            entry['kinases'] are not known.
```
bgyori: The default value/behavior for the parameters should be described

indra/sources/phosphoelm/processor.py
```diff
@@ -0,0 +1,139 @@
...
+            logger.warning('Gilda service responded with status code %d' %
+                           res.status_code)
+        return entity_str, None, None
+
+
+class PhosphoElmProcessor(object):
+    def __init__(self, file_dump_json=None, keep_empty=False):
+        """The PhosphoElmProcessor processes data dumps from the phospho.ELM
+        database. See http://phospho.elm.eu.org/dataset.html
+
```
bgyori:
```
Parameters
-----------
```
heading missing

indra/sources/phosphoelm/processor.py
```diff
@@ -0,0 +1,139 @@
...
+        file_dump_json : list(dict)
+            JSON compatible list of entries from a phospho.ELM data dump
+        keep_empty : bool
+            If true, also create statements when upstream kinases in
+            entry['kinases'] are not known.
+        """
+        self.statements = []
+        self.statements.extend(self._from_file_dump_json(file_dump_json,
+                                                         keep_empty))
```
bgyori: Most standard processors are implemented such that the constructor doesn't immediately do processing. Rather, the API calls one or more processor functions to populate the statements attribute (e.g. `process_phosphorylations`). The advantage is that the processor can be instantiated without starting any specific processing. I believe this may have to do with my comment on a None input argument to the constructor above.

indra/sources/phosphoelm/processor.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def _from_file_dump_json(self, fd_json, keep_empty=False):
+        """Create Phosphorylation statements from the json entries
+
```
bgyori: Parameters header missing

indra/sources/phosphoelm/api.py
kkaris: Using only `f` makes `csv_reader` an object that has to be consumed within the `with-open` part, the idea is to provide `csv_reader` either from the s3 client or from the local file, but I can change it so we have `ppelm_json` be produced inside either of the if-else clauses instead.

indra/sources/phosphoelm/processor.py
```diff
@@ -0,0 +1,145 @@
...
+    phosphoelm_data : list(dict)
+        JSON compatible list of entries from a phospho.ELM data dump
+
+    Attributes
+    ----------
+    statements : list(indra.statement.Phosphorylation)
+        A list of the phosphorylation statements produced by the entries
+        in phosphoelm_data
+    """
+    def __init__(self, phosphoelm_data=None):
```
bgyori: I don't think the `phosphoelm_data` argument should be optional - the processor is not meaningful to instantiate without something to process. Consequently, the 
```
        if self._phosphoelm_data is None:
            logger.warning('No phosphoelm data to process. No statements '
                           'are produced')
            return
```
block below can also be removed.

indra/tests/test_phosphoelm_processor.py
```diff
@@ -0,0 +1,66 @@
...
+human_kinase2 = ['O14746', 'FAKEPRCRAVRSLL', '12', 'S', '10224060',
+                 'PKB_group', 'LTP', 'Homo sapiens',
+                 '2004-12-31 00:00:00+01']
+
+raw_data = [columns, non_human_no_kinase, human_no_kinase, human_kinase1,
+            human_kinase2]
+
+
+def test_download_from_s3():
+    s3 = get_s3_client(False)
```
bgyori: This test needs to be tagged as `nonpublic` if it uses AWS credentials.

kkaris:
All the tests related to the processor are passing now.
@johnbachman and @bgyori, could any of you check that the special case mappings (in phosphoelm_mapping.py) make sense?

bgyori:
I think this is good now and extracts 3k+ phosphorylation relations. However, it does not actually address #177 which is specifically about getting information on binding between phosphorylated proteins and specific domains of other proteins (e.g., as shown on this page http://phospho.elm.eu.org/cgimodel.py?fun=printSubstratePage1&substrateId=460&substrate=EGFR&organisms=instances_vertebrates&MS=). This information doesn't appear to be part of the standard downloadable table so a distinct approach from this PR would be needed to get any of that information.

PR #1022, opened 2019-12-31
Update MeSH resources and fix MeSH test 

bgyori:
This PR updates the MeSH name/ID mapping files to their latest versions. In an attempt to fix the failing test_mesh_term_lookups test, I investigated both the downloadable dumps and the web service to see why  'Colorectal Cancer' does not map to  'D015179' (Colorectal Neoplasms) anymore. I found that 'Colorectal Cancer' has been removed as a term from D015179's descriptor and appears as a narrower synonym to it under the corresponding M heading (rather than the D descriptor). Similarly, the current web service does not return this mapping anymore.

Two confusing facts remain, however: (1) On the MeSh website, Colorectal Cancer still appears as a term for D015179 (https://meshb.nlm.nih.gov/record/ui?ui=D015179), and (2) Similar relationships, such as 'Prostate Cancer' as a term of D011471 (Prostatic Neoplasms) are still there, which seems to be somewhat inconsistent.


bgyori:
The MeSH tests now pass, the failure is an unrelated Eidos issue reported on that repo.

PR #1023, opened 2019-12-31
Updated TAS extractions

bgyori:
This PR updates the Target Affinity Spectrum data set to the latest version which is substantially larger than the previous version. The old data set is removed from version control in the resources folder, and the new one is made accessible as a public file via S3 (can be downloaded simply using requests).  

The structure of the data is similar to the previous one but has multiple specific issues that need to be resolved/discussed before merging this PR:
1. The data normalizes drug-target relationships across multiple distinct ChEBML IDs that are considered to be essentially the same chemical. The difficulty is that often these entries correspond to distinct ChEBI IDs, and INDRA's assembly considers then to be different. So one might see groups of statements like
```
 Inhibition(bambuterol(), BCHE()),
 Inhibition((R)-bambuterol(), BCHE()),
 Inhibition(bambuterol hydrochloride(), BCHE()),
```
that TAS considers equivalent but INDRA would consider as three distinct inhibitors of BCHE. One possible solution is to somehow make use of the LSPCI which is an internal ID in TAS that is shared by groups of chemicals that TAS considers equivalent.

2. The naming of chemicals from TAS falls into multiple categories that require custom treatment. a) Currently only those chemicals that have ChEBI mappings are assigned a standard name. b) There may be a subset of chemicals that aren't mapped to ChEBI but still have a name available. c) The majority of chemicals that ChEBML reports target data for do not have any sort of displayable name in ChEBML, they are just compounds with a structure and ID. Currently the processor simply assigns the CHEBML ID as the Agent name in case of b) and c). e.g., 
```
Inhibition(CHEMBL504301(), PSMB5())
```

3. The size of the new data set and especially the large set of unnamed chemicals may compromise the usability of certain downstream applications. For instance, the new TAS data set reports 2588 chemical inhibitors of BRAF (2559 of which have just CHEMBL IDs with no name), which would certainly overwhelm any kind of visual or natural language representation of a "network" around BRAF. Some downstream apps would need to be adapted with targeted filters or a more sophisticated separation of statements in their output to deal with this. On the other hand, for some large scale, automated analysis applications, the large quantity of such relationships could be very valuable.

4. TAS provides provenance for many of the relationships (e.g., DOIs), however, the current export doesn't contain this information. It may make sense to implement this before merging.


bgyori:
The current state of the PR addresses all several of the above issues as follows:
- Where available, CHEMBL names are assigned to drugs, even if a CHEBI ID is not available. CHEBML names are all caps, therefore not always appealing, but still better than no name at all.
- Provenance information in evidences is now extracted with multiple sources including PMIDs and source IDs (e.g., entries in HMS LINCS or US patents).
- The API adds additional parameters: `named_only` and `standardized only` to control what drug information is extracted. For `named_only`, we only extract statements in which the drug has a proper name, and for `standardized_only`, we extract only statements in which the drug is fully ID-and-name standardized. This allows easily customizing the amount and quality of information for various downstream use cases.

bgyori:
Alright, I think I'll merge this now. Thanks again @steppi!

PR #1025, opened 2020-01-07
Add some new flexibility to the HTML assembler.

pagreene:
This PR allows the dict-and-list structure that is sent to Jinja to be returned directly for use in new applications. This PR also makes changes fixing a prior bug where the same content could sometimes appear twice, and allows headings to have labeled named entities at every level of the drop-down.

(As of PR, some testing needed before merge)


bgyori:
Looks good

PR #1035, opened 2020-01-30
Wm updates

kkaris:
This PR updates the LiveCurator and its associated service to be able to load corpora during runtime based on its ID.

Other improvements:
- Local caching of corpora components
- Save the state of curations for given corpus ID
- The app can start without loading any corpus immediately
- A new meta data attribute is added to the Corpus class


bgyori:
One more small thing: when running the service without any arguments, it tries to load "1" from S3, i.e.
```
INFO: [2020-01-30 19:25:10] live_curation - Loading corpus: indra_models/1/
WARNING: [2020-01-30 19:25:10] live_curation - Failed to get from s3: An error occurred (AccessDenied) when calling the GetObject operation: Access Denied
```

PR #1043, opened 2020-02-13
Update docs and examples

kkaris:
This PR updates parts of the documentation, with focus on examples and tutorials.

Main changes:
- More references to different modules and outside sources have been added
- Code in the changed files and appropriate functions have been updated to be able to run with the latest version
- Some documentation have been removed to focus on fewer examples for clarity
- More examples of output assembly have been added with IndraNet and PySB


bgyori:
I think these changes are great! I wonder if we could make all examples that appear in the documentation (or at least scaled-down versions of them) Travis-tested to make sure they continue to be operational?

kkaris:
I'm looking into why the test stalled. I believe it's one of the tests that I added that made it too slow, I'm tracking down which one it might be

bgyori:
I made some updates to the examples, in particular the way REACH is used, and edited some of the text. Along the way, I also made a number of updates in the REACH API/processor and tests too.

PR #1055, opened 2020-03-24
Refactor LiveCuration and add download curations endpoint

kkaris:
This PR updates the live curation service with the ability to download curations with their corresponding statements per corpus id with the option to narrow down the results to only contain curation and statements from a given reader.


bgyori:
@kkaris I took this opportunity to restructure the code into a separate module. I also changed Corpus such that its first constructor argument is the corpus_id, which becomes one of its attributes. I think this is convenient since then it doesn't need to be managed separately. Could you take another look and see if you can spot any issues with the restructuring? Thanks!

kkaris:
I only saw a small thing I fixed in `__main__` in `live_curation.py`, but otherwise it looked ok. The only test that didn't pass locally for me was the `LiveGroundingTestCase.test_add_ontology_node()`, but the error was related tonot having the edidos service running so I thinks it's ok.

PR #1058, opened 2020-03-30
Add evidence source hashes to edges in PyBEL assembler

cthoyt:
This PR isn't done yet, but it will make some improvements for the rational enrichment package.

- [x] Update export functions (5bc73624)
- [x] Add source hash of evidence to PyBEL edge data dictionary
- [x] Add ability to output ungrounded entities (this might be useful in a curation setting; b0cdba58)

### DIFF
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -431,11 +431,11 @@ def _combine_edge_data(relation, subj_edge, obj_edge, stmt_hash, evidences):
     for ev in evidences:
         edge_data_one = copy(edge_data)
         citation, evidence, annotations = _get_evidence(ev)
-        edge_data.update({
+        edge_data_one.update({
             pc.CITATION: citation,
             pc.EVIDENCE: evidence,
         })
-        edge_data[pc.ANNOTATIONS].update(annotations)
+        edge_data_one[pc.ANNOTATIONS].update(annotations)
         edge_data_list.append(edge_data_one)
     return edge_data_list
 
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -218,11 +218,11 @@ class PybelAssembler(object):
         else:
             with open(path, 'w') as fh:
                 if output_format == 'json':
-                    pybel.to_json_file(self.model, fh)
+                    pybel.to_nodelink_file(self.model, fh)
                 elif output_format == 'cx':
                     pybel.to_cx_file(self.model, fh)
                 else: # output_format == 'bel':
-                    pybel.to_bel(self.model, fh)
+                    pybel.to_bel_script(self.model, fh)
 
     def _add_nodes_edges(self, subj_agent, obj_agent, relation, stmt_hash,
                          evidences):
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -440,6 +440,7 @@ def _update_edge_data_from_evidence(evidence, edge_data):
 def _get_annotations_from_stmt(stmt):
     return {
         'stmt_hash': stmt.get_hash(refresh=True),
+        'uuid': stmt.uuid
     }
 
 
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -234,7 +234,11 @@ class PybelAssembler(object):
         self.model.add_node_from_data(subj_data)
         self.model.add_node_from_data(obj_data)
         edge_data_list = _combine_edge_data(
-            relation, subj_edge, obj_edge, stmt)
+            relation=relation,
+            subj_edge=subj_edge,
+            obj_edge=obj_edge,
+            stmt=stmt,
+        )
         for edge_data in edge_data_list:
             self.model.add_edge(subj_data, obj_data, **edge_data)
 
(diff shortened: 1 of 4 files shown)


cthoyt:
Between the INDRA statement UUID and the evidence source hash, it should be enough information to send curated information (at least correct/incorrect) back to the INDRA DB, right?

indra/assemblers/pybel/assembler.py
```diff
@@ -589,7 +589,9 @@ def _get_evidence(evidence):
                     pc.CITATION_IDENTIFIER: cit_ref_str}
     pybel_ev[pc.CITATION] = citation
 
-    annotations = {}
+    annotations = {
+        'source_hash': evidence.get_source_hash(),
```
cthoyt: All evidences have source hashes, regardless of if there's a source API or source identifier

indra/assemblers/pybel/assembler.py
```diff
@@ -414,20 +415,27 @@ def belgraph_to_signed_graph(
...
+        pc.RELATION: relation,
+        pc.ANNOTATIONS: {},
+    }
     if subj_edge:
         edge_data[pc.SUBJECT] = subj_edge
     if obj_edge:
         edge_data[pc.OBJECT] = obj_edge
     if stmt_hash:
-        edge_data['stmt_hash'] = stmt_hash
+        edge_data[pc.ANNOTATIONS]['stmt_hash'] = stmt_hash
```
cthoyt, line 427: @bgyori this is where we should put more statement-level metadata from INDRA

indra/assemblers/pybel/assembler.py
```diff
@@ -574,22 +584,27 @@ def _get_agent_activity(agent):
...
-                    pc.CITATION_IDENTIFIER: cit_ref_str}
-    pybel_ev[pc.CITATION] = citation
+        citation = citation_dict(
+            db=pc.CITATION_TYPE_OTHER,
+            db_id=cit_ref_str,
+        )
 
-    annotations = {}
+    annotations = {
+        'source_hash': evidence.get_source_hash(),
```
cthoyt, line 606: @bgyori this is where we should put more evidence-level metadata from INDRA

indra/assemblers/pybel/assembler.py
bgyori: Yes, the UUID could be useful but it would take a bit of refactoring to expose it here since we only propagate the stmt_hash down to function.

bgyori:
@dianakolusheva please take a look and see if these changes affect how path finding and the processing of results is affected, since the position of the statement hash in the PyBEL edge structure is moved by this PR.

indra/assemblers/pybel/assembler.py
cthoyt: Refactored to propagate the statement instance in dee6f1f0

indra/tests/test_pybel_assembler.py
```diff
@@ -115,14 +112,15 @@ def test_modification_with_evidences():
     edge_data = get_edge_data(belgraph, braf_dsl, map2k1_mod_dsl)
     assert edge_data.get(pc.SUBJECT) == activity('kin')
     assert edge_data[pc.RELATION] == pc.INCREASES
-    assert edge_data[pc.EVIDENCE] == 'evidence text'
```
cthoyt, line 118: this should still always be available, so I would leave it with `[]` instead of `.get()`

indra/assemblers/pybel/assembler.py
```diff
@@ -247,72 +247,62 @@ def _assemble_regulate_activity(self, stmt):
...
-            self._add_nodes_edges(stmt.agent, act_agent, relation,
-                                  stmt_hash, stmt.evidence)
+            self._add_nodes_edges(stmt.agent, act_agent, relation, stmt)
         else:
             for mod in stmt.agent.mods:
                 mod_agent = Agent(
                     stmt.agent.name, db_refs=stmt.agent.db_refs, mods=[mod])
-                self._add_nodes_edges(mod_agent, act_agent, relation,
-                                      stmt_hash, stmt.evidence)
+                self._add_nodes_edges(mod_agent, act_agent, relation, stmt)
```
cthoyt, line 295: were these in the wrong order? shouldn't relation be first?

cthoyt: nevermind

indra/tests/test_pybel_assembler.py
bgyori: It's just to make the test FAIL instead of ERROR if the key is missing (which it was  before I updated the test), it's fine either way though.

cthoyt: ah okay

bgyori:
I fixed an issue with Travis tests getting stuck, and now it shows that a few model checking tests related to PyBEL paths are broken, probably because of having moved the stmt_hash under annotations.

dianakolusheva:
> @dianakolusheva please take a look and see if these changes affect how path finding and the processing of results is done, since the position of the statement hash in the PyBEL edge structure is moved by this PR.

It affects the reporting part, I added a change to update this.

indra/explanation/reporting.py
```diff
@@ -168,7 +168,7 @@ def stmts_from_pybel_path(path, model, from_db=True, stmts=None):
         hashes = set()
         for j in range(len(edges)):
```
cthoyt, line 169: can we switch this to use the pythonic generator instead of indexing?

```python
for edge in edges:
    try:
        hashes.add(edge['annotations']['stmt_hash']
        # partOf and hasVariant edges don't have hashes
    except KeyError:
        continue
```

cthoyt: wait, it seems like edges is a dictionary where keys are consecutive integers... wouldn't a list be a better data structure?

dianakolusheva: I think it's an artifact of an earlier version when all_edges have been used directly instead of filtering to edges, it mimics the data structure of all_edges (which is I think a standard NetworkX EdgeView structure)

bgyori: I can't recall why it's implemented this way, we'll review and think about it as a follow up but I'll merge this PR first.

PR #1062, opened 2020-04-07
Handle NDEx client errors

dianakolusheva:
This PR adds handling of exception happening at style setting stage and increases sleep time for updating the profile (it requires more time for larger models)

### DIFF
--- a/indra/databases/ndex_client.py
+++ b/indra/databases/ndex_client.py
@@ -222,15 +222,7 @@ def update_network(cx_str, network_id, ndex_cred=None):
             logger.error('Could not update NDEx network profile.')
             logger.error(e)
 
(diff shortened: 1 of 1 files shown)


bgyori:
Thanks, this looks good! Do you think it would make sense to put the second loop inside the set_style function? It gets called from multiple places so then the retries would be handled in one place inside the function.

dianakolusheva:
> Thanks, this looks good! Do you think it would make sense to put the second loop inside the set_style function? It gets called from multiple places so then the retries would be handled in one place inside the function.

One strange behavior I noticed is that the function `set_style` works fine if called directly (from ipython) and only fails when `update_network` is called without the time.sleep (and only for covid19 model, probably because of the size).

PR #1068, opened 2020-04-14
Use internal NXML processing in ISI preprocessor

bgyori:


bgyori:
The tests are failing locally, and there are some extensions to the NXML processing code that @steppi will look into before we merge this.

PR #1076, opened 2020-04-20
Overhaul GO integration via OBO client

bgyori:
This PR removes all the GO resource files and replaces them with a single JSON file derived from go.obo and exposed via the OboClient class. It also makes multiple changes and improvements to the OboClient, for instance, better processing of synonyms, and representation of isa relationships. This also improves the integration of EFO, HP, and DOID, and extends the corresponding resource files.

The following resource files are removed and replaced by go.json:
- cellular_components.tsv
- cellular_components_patch.tsv
- go_id_label_mappings.tsv
- go_secondary_mappings.tsv
The cellular component hierarchy and the uniprot subcellular location mappings are updated.

Further, the PR removes cellular location resources and manipulation from the `statements` module, and changes processors and assemblers to use the `go_client` module to work with cellular locations instead.


bgyori:
The GO ontology has changes, for instance, "cell part" is now deprecated, resulting in a situation where e.g., "plasma membrane" is no longer part of "cell". I am looking into this to see what can be done.

bgyori:
All the tests are passing, the failures are due to Travis issues on tests that pass locally.

PR #1079, opened 2020-04-21
Assembly pipeline

dianakolusheva:
This PR introduces a new AssemblyPipeline class which can simplify creating and replicating pipelines for statements assembly. AssemblyPipeline can be initiated manually or from a json file. In case the json file is used, all functions referenced there must be registered with register_pipeline decorator (manually appended functions will be registered automatically). Many of the functions defined within Indra package and commonly used for assembly are already registered.


indra/pipeline/pipeline.py
```diff
@@ -0,0 +1,322 @@
...
+            return pipeline_functions[name]
+        raise NotRegisteredFunctionError('%s is not registered' % name)
+
+    def run_simple_function(self, func, *args, **kwargs):
+        """Run a simple function - simple here means a function all arguments
+        of which are simple values (do not require extra function calls).
+        """
+        if 'statements' in kwargs:
+            statements = kwargs['statements']
+            del kwargs['statements']
```
bgyori, line 197: We can use `statements = kwargs.pop('statements', None)`

PR #1086, opened 2020-05-01
Agent coordinates in English and HTML assemblers

dianakolusheva:
This PR updates EnglishAssembler and HtmlAssembler code. AgentWithCoordinates and SentenceBuilder classes are added in English assembler to store the (relative) location of agent in text. The result of EnglishAssembler.make_model() is unchanged which ensures that no changes are needed in any downstream applications, but the EnglishAssembler object contains the AgentWithCoordinates objects to keep track of agents' coordinates. HTML assembler is using the new structure of EnglishAssembler to tag the agents in statement text (previous version of matching the agent names with regex is also supported for statements that could not be assembled with EnglishAssembler).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,7 @@ indra/benchmarks/assembly_eval/batch4/trips
 indra/benchmarks/assembly_eval/batch4/reach
 models/fallahi_eval/output
 indra/sources/eidos/cache/geonames/index/
+indra/tests/tempfile.html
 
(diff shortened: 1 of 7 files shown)


dianakolusheva:
Additional changes:
- _format_evidence_text function in html.assembler is extended to provide counts for correct/incorrect curations
- html assembler tests are extended and more tests are added

PR #1105, opened 2020-06-03
Bfs update

kkaris:
This PR changes the default behavior of `bfs_search()` when the `terminal_ns` argument is provided. Now, the generator _only_ yield paths that terminate on nodes from the provided namespaces. 

Previously any path would be yielded.

Another change in the PR is to add a memory limit for the sum of the memory allocated for two variables in the function as a safeguard.

Tests have been updated and added.


kkaris:
The failing tests are in `test_pysb_assembler`. The one's in `test_pathfinding` for this PR are passing.

PR #1107, opened 2020-06-05
FLASK REST API

dianakolusheva:
This PR reimplements INDRA REST API using Flask RestX framework. Endpoints based on assembly_corpus functions are dynamically generated, while others are added manually. Because RestX generates Swagger documentation automatically, `doc` submodule is removed. Two previous test files (`Indra.tests.test_rest_api.py` and `rest_api.test_api.py`) are merged, the tests are updated, and new tests are added. Testing is done using local app (not calling web service).

Documentation improvements are in process.


bgyori:
This is really great work @dianakolusheva! The only issue I see is that docstring-parser requires at least Python 3.6 so this would effectively make INDRA 3.6+ only as well which will require some additional migration work.

bgyori:
All the tests are passing now 🎉

PR #1112, opened 2020-06-18
Expose multiprocessing traceback

kkaris:
This PR adds the WrapException class, which is used to expose exceptions occurring in child processes when running multiprocessing. The doc of the file containing the class contains a minimum working example.


kkaris:
I could not get a test together, I kept getting the error `ValueError: Pool is still running`

PR #1118, opened 2020-06-24
Propagate statement information to BEL signed edge graph

kkaris:
This PR allows annotations in a belgraph to be propagated to the corresponding signed edge graph. Statement belief was also added as one of the annotation to add from statements.

A test is also added to `test_pybel_assembler.py`.


kkaris:
Some tests are failing, I have to look into it

PR #1120, opened 2020-06-26
Add Full Text search links for PMC in HTML Assembler

PritiShaw:
Adds [Scroll To Text Fragment](https://chromestatus.com/feature/4733392803332096) feature for links to PMC journals, when 
-  `add_full_text_search_link` is `True` in `make_model()`
-  *PMCID* is available for the sentence

The links will automatically search for the sentence, scroll to it, and highlight, without using any plugin.
Highlighted Sentence

#### Sample Output
HTML Assembler [output](https://gistcdn.githack.com/PritiShaw/840ad0a586d5f43da5b83f59080e6313/raw/83e9189d67a233f62ceff0da952ba6f45d689a8c/output.html), [source](https://gist.github.com/PritiShaw/840ad0a586d5f43da5b83f59080e6313#file-output-html)
[Script used to generate above output](https://gist.github.com/PritiShaw/840ad0a586d5f43da5b83f59080e6313#file-script-py)
Sample Output from HTML Assembler

#### Supported Browsers
- Chrome v80+for both desktop and android
- Chromium-based Edge, Opera

### DIFF
--- a/indra/assemblers/html/assembler.py
+++ b/indra/assemblers/html/assembler.py
@@ -325,7 +325,7 @@ class HtmlAssembler(object):
         add_full_text_search_link : bool
             If True, link with Text fragment search in PMC journal will be added
             for the statements
-
+            
         All other keyword arguments are passed along to the template. If you
         are using a custom template with args that are not passed below, this
         is how you pass them.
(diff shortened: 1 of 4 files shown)


bgyori:
Thanks a lot @PritiShaw, looks good! I have a question about this condition:
```python
if not set(['PMCID', 'pmcid']) \
    .issubset(evidence["text_refs"].keys()):
```
It looks like here you have the `{'PMCID', 'pmcid'}` set and compare this against `evidence["text_refs"].keys()` which will typically be something like `dict_keys(['PMID', 'TRID', 'PMCID', 'DOI', 'TCID', ...])`. So there will never be an `issubset` relation and the if condition will thus always be satisfied. So this condition should be replaced by
```python
if 'PMCID' not in evidence.get('text_refs', {})
```
There are a few other obscure corner cases that we'll fix (like PMID missing, etc.) once you made this change. Thanks!

PritiShaw:
> ```python
> if 'PMCID' not in evidence.get('text_refs', {})
> ```

Thanks for your review, I have modified the if the condition. I tested the code with the same script ([output](https://gistcdn.githack.com/PritiShaw/5673cc7d996a49063f003c4eb978df1d/raw/472c4437b69162e3b14361bf881748d2a81cfd11/output.html))
I was trying to take care the condition where PMID is missing, but could not understand what will be the fallback to get PMCID.

Thanks

PR #1131, opened 2020-07-12
Handle location change of citation_dict

djinnome:
According to https://github.com/pybel/pybel/pull/453
@cthoyt  moved `citation_dict` from `pybel.utils` to `pybel.language`
This code tries to import  `citation_dict` from the new `pybel.language` location and if it results in an `ImportError` it tries to import from the old `pybel.utils` location.

### DIFF
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -8,7 +8,10 @@ import pybel
 import pybel.constants as pc
 from pybel.dsl import *
 from pybel.language import pmod_namespace
-from pybel.utils import citation_dict
+try:  # this works after pybel pull request #453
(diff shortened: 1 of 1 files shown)


bgyori:
Thanks, @djinnome, the try/import looks like a good way to handle this since it will work with both versions. We are planning to go through the rest of the code and make it compatible with pybel 0.15+ soon, and then change the version constraints in INDRA's setup.py (which currently requires <0.15.0).

cthoyt:
I will take care of making a PR to update the PyBEL I/O in INDRA once v0.15.0 is finalized, so @bgyori please don't start working on this until I've finalized the feature set (to save yourself the hassle of doing it twice)

PR #1137, opened 2020-07-17
Allow for import of HTML assembler without id_lookup.

pagreene:
This allows the HTML assembler to be imported without committing the the extensive bloat that comes with `id_lookup`. When `id_lookup` is not available, it simply does not add the PMCID.

### DIFF
--- a/indra/assemblers/html/assembler.py
+++ b/indra/assemblers/html/assembler.py
@@ -17,7 +17,10 @@ from indra.assemblers.english import EnglishAssembler, AgentWithCoordinates
 from indra.databases import get_identifiers_url
 from indra.util.statement_presentation import group_and_sort_statements, \
     make_top_level_label_from_names_key, make_stmt_from_sort_key
-from indra.literature import id_lookup
+try:
+    from indra.literature import id_lookup
+except:
+    id_lookup = None
 
 logger = logging.getLogger(__name__)
 HERE = dirname(abspath(__file__))
@@ -363,11 +366,12 @@ class HtmlAssembler(object):
                         for evidence in stmt_info["evidence"]:
                             if 'PMCID' not in evidence.get('text_refs', {}):
                                 if evidence.get('pmid'):
-                                    ev_pmcid = id_lookup(
-                                        evidence['pmid'], 'pmid') \
-                                        .get('pmcid', None)
-                                    if ev_pmcid:
-                                        evidence['pmcid'] = ev_pmcid
+                                    if id_lookup:
+                                        ev_pmcid = id_lookup(
+                                            evidence['pmid'], 'pmid') \
+                                            .get('pmcid', None)
+                                        if ev_pmcid:
+                                            evidence['pmcid'] = ev_pmcid
                             else:
                                 evidence['pmcid'] = \
                                     evidence['text_refs']['PMCID']


pagreene:
This PR fixes an issue that was breaking the DB REST API.

bgyori:
I wonder if a more general solution to this issue would be to use mock objects in the REST API (https://realpython.com/python-mock-library/). That would make what is loaded directly controlled by the API code without INDRA's code having to be adapted.

pagreene:
I would actually think of that as a more specific, not more general solution. In general I think we are a bit careless about our dependencies, and make extensive webs of dependencies "necessary" when they have no real need to be. The use of `id_lookup` is extremely minor, but the web of dependencies it requires is quite large. I don't think the REST API is the only place this might come up as an issue.

Also it looks like mock is mainly intended for testing, and furthermore I'm not sure how I could inject my mock of `id_lookup` into the scope of HTMLAssembler from the scope of the REST API code. That said, if there is a strong objection to this change in INDRA, I could invest some more time to try to figure out how else to get around this.

bgyori:
Well, `id_lookup` is a fully integrated module of INDRA which is installed in any standard INDRA version so it shouldn't normally be controversial to import it. The REST API is an idiosyncratic application in that it removes parts of INDRA due to technical constraints. While one such change is probably okay, I don't think INDRA's code should be subjected to try/except blocks for this purpose . I can only imagine that the same kind of thing will happen again in other contexts (what if for some reason we need to import id_lookup in another module for some other reason later?) and so it is better to control this from the API directly than from all the different parts of INDRA.

Mock is indeed originally intended for testing but it is more generally applicable for this purpose (whenever the import of an actual module is undesirable, for instance, in readthedocs where there are resource constraints) - so I don't think that is an issue. Scoping is also solved as we implemented here for instance: https://github.com/sorgerlab/indra/blob/master/doc/conf.py#L304-L317. In my view the REST API could declare a similar list of modules to protect against undesirable loading of modules that it doesn't want to load.

pagreene:
The link you give uses MOCK in the testing aspect of Sphinx docs, so I am still not sure if that would apply to the REST API. I could dig into the Sphinx code to figure out how they apply the mocking though, and see if it could be adapted.

That said, as someone that has shifted largely to using INDRA rather than developing it, I still believe that the more general approach is that we should start trying to design INDRA in a way that "full integration" is not necessary to use any part of the system. As you say, this isn't likely to be the only time we run into issues with the sprawling dependency trees, and I also don't think the REST API is the only place that won't want to import the world just to use a minute subset of INDRA functionality. We _often_ struggle with RAM and disk limitations, and INDRA has gotten to be quite large in both.

So my belief is that starting to better compartmentalize INDRA, and giving users non-complicated-workaround ability to limit the scope of the resources they need as a potentially valuable feature of INDRA. But I am also looking into how to mock the import in this context.

johnbachman:
In general I agree with Ben. While we have refactored pieces of INDRA out in the past (e.g., ProtMapper, INDRA DB) it was largely motivated by diverging development paths and publication considerations. In general IMO it's to our advantage to keep the framework unified because it provides a conceptual structure for how the pieces fit together. Let's discuss in the meeting.

bgyori:
```python
import sys                                                                                         
import mock                                                                                        
sys.modules['indra.literature'] = mock.MagicMock()                                                 
sys.modules['indra.literature'].id_lookup.return_value = None                                      

# This can be in any code deeper in the import stack
from indra.literature import id_lookup                                                             
id_lookup('xxxx')                                                                                  
```

PR #1151, opened 2020-08-03
Bfs mesh ids filtering

kolusask:
This PR adds mesh-ids-based edge filtering to BFS search used in finding open paths

### DIFF
--- a/indra/explanation/pathfinding/pathfinding.py
+++ b/indra/explanation/pathfinding/pathfinding.py
@@ -286,6 +286,8 @@ def bfs_search(g, source_node, reverse=False, depth_limit=2, path_limit=None,
             allowed_edges = [(u, v) for u, v in g.edges() if allow_edge(u, v)]
         else:
             return []
+    else:
+        allowed_edges = []
 
     queue = deque([(source_node,)])
     visited = ({source_node}).union(node_blacklist) \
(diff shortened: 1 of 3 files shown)


kkaris (commented):
Nice work, just some minor changes needed.

indra/explanation/pathfinding/pathfinding.py
```diff
@@ -263,7 +354,8 @@ def bfs_search(g, source_node, reverse=False, depth_limit=2, path_limit=None,
                             ign_vals = None
                             pass
                         else:
-                            ign_vals = yield new_path
+                            ign_vals = yield new_path, \
+                                       ['N/A'] * (len(new_path) - 1)
```
dianakolusheva, line 358: It seems like the function is changed to generate a tuple of a path + ignore values instead of just a path? Does it do that always or just in some condition? Can you update the documentation? Also this function is called in `bfs_search_multiple_nodes` which currently only expects paths be generated, function call has to be updated there.

dianakolusheva (changes requested):

dianakolusheva (approved):
Code causing issues in EMMAA is fixed

kkaris (approved):
Good job everyone, and @kolusask in particular

johnbachman (approved):

PR #1154, opened 2020-08-06
Update inheritance of CTD processor

cthoyt:
This reduces some redundant code and makes it easier for the IDE to introspect on the object that's been returned


bgyori:
Thanks!

PR #1156, opened 2020-08-12
Mesh supc

steppi:
This PR updates the function `get_ids_for_mesh` in `indra.literature.pubmed_client` to allow for querying pmids annotated with MeSH supplementary concepts. The new behavior is that if major_topic is set to False and the input MeSH id is a supplementary concept starting with `'C'`, then in addition to the query being made for regular MeSH terms, a query is also made based on supplementary concepts. The returned set of pmids is then the union of the pmids returned for the query based on regular MeSH terms and that based on supplementary concepts.


bgyori:
Thanks!

PR #1158, opened 2020-08-18
Omnipath processor

kkaris:
This PR adds the Omnipath processor that queries the omnipath web api. Currently two datasets are handled, the [ptms dataset](https://github.com/saezlab/pypath/blob/master/webservice.rst#enzyme-substrate-interactions) containing enzyme-substrate interactions and the [ligand-receptor interactions dataset](https://github.com/saezlab/pypath/blob/master/webservice.rst#interaction-datasets).


doc/modules/sources/index.rst
```diff
@@ -45,6 +45,7 @@ Standard Molecular Pathway Databases
    virhostnet/index
    ctd/index
    drugbank/index
+   omnipath/index
```
bgyori, line 48: The actual index file for omnipath is not in version control currently

indra/databases/hgnc_client.py
```diff
@@ -207,7 +207,6 @@ def get_hgnc_from_rat(rgd_id):
         rgd_id = rgd_id[4:]
     return rat_map.get(rgd_id)
 
-
 def get_rat_id(hgnc_id):
```
bgyori, line 210: We could add the line back here that was removed

indra/sources/omnipath/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .api import process_from_web
```
bgyori, line 1: We could add a brief triplequoted description above the first import of the module and approach to accessing/processing information from Omnipath that will then show up in the documentation.

indra/sources/omnipath/api.py
```diff
@@ -0,0 +1,65 @@
+import logging
+import requests
+from .processor import OmniPathProcessor
+
+logger = logging.getLogger("omnipath")
```
bgyori, line 5: We can use `__name__` here to construct the logger.

indra/sources/omnipath/processor.py
```diff
@@ -0,0 +1,175 @@
+from __future__ import unicode_literals
+import logging
+from collections import Counter
+from indra.preassembler.grounding_mapper import GroundingMapper
+from indra.statements import modtype_to_modclass, Agent, Evidence, Complex, \
+    get_statement_by_name as stmt_by_name, BoundCondition
+
+logger = logging.getLogger(__file__)
```
bgyori, line 8: We should use `__name__` here to construct the logger.

indra/sources/omnipath/processor.py
```diff
@@ -0,0 +1,175 @@
+from __future__ import unicode_literals
+import logging
+from collections import Counter
+from indra.preassembler.grounding_mapper import GroundingMapper
```
bgyori, line 4: It would be better to use `from indra.ontology.standardize import standardize_agent_name` and use that below.

indra/sources/omnipath/processor.py
```diff
@@ -0,0 +1,175 @@
...
+    get_statement_by_name as stmt_by_name, BoundCondition
+
+logger = logging.getLogger(__file__)
+
+
+class OmniPathProcessor(object):
+    def __init__(self, ptm_json=None, ligrec_json=None):
+        self.statements = []
+        self.statements.extend(self._stmts_from_op_mods(ptm_json))
+        self.statements.extend(self._stmt_from_op_lr(ligrec_json))
```
bgyori, line 15: In many other processors, these two calls are actually done in the API so that the processor can be constructed without immediately starting processing. However, we can leave it like this too, not a big issue.

indra/sources/omnipath/processor.py
```diff
@@ -0,0 +1,175 @@
...
+                           f'regulation were skipped')
+
+        return ligrec_stmts
+
+    @staticmethod
+    def _agent_from_up_id(up_id):
+        """Build an Agent object from a Uniprot ID. Adds db_refs for both
+        Uniprot and HGNC where available."""
+        db_refs = {'UP': up_id}
+        ag = Agent(None, db_refs=db_refs)
```
bgyori, line 124: We could set the `up_id` as an interim name here to avoid the corner case possibility of ending up with a None Agent name.

indra/statements/statements.py
```diff
@@ -195,6 +207,9 @@
     'Complex', 'Translocation', 'RegulateAmount', 'DecreaseAmount',
     'IncreaseAmount', 'Influence', 'Conversion', 'Unresolved',
     'Association', 'Event', 'Migration',
+    'Carboxylation', 'Decarboxylation', 'Amidation', 'Deamidation', 'Sulfation',
```
bgyori, line 210: I'm trying to think of any other places we need to account for these statement types. One would be the INDRA Statements JSON specification but there may be more.

indra/tests/test_omnipath.py
```diff
@@ -0,0 +1,60 @@
+import requests
+from indra.sources.omnipath import OmniPathProcessor
+from indra.sources.omnipath.api import op_url
+from indra.statements import Agent
+from indra.preassembler.grounding_mapper import GroundingMapper
```
bgyori, line 5: Same as in the processor, we should now use `from indra.ontology.standardize` import standardize_agent_name. Given that the ontology is mocked on Travis, we may need to add some more entries to the mock ontology to make tests pass.

models/paper2/stmts_vs_pmids.py
```diff
@@ -0,0 +1,92 @@
+import pickle
```
bgyori, line 1: I think this file may have been added by accident, we could remove it from the PR.

setup.py
```diff
@@ -78,9 +78,9 @@ def main():
                     'indra.sources.hume', 'indra.sources.hypothesis',
                     'indra.sources.index_cards',
                     'indra.sources.indra_db_rest', 'indra.sources.isi',
-                    'indra.sources.lincs_drug',
-                    'indra.sources.ndex_cx', 'indra.sources.phosphoelm',
-                    'indra.sources.reach',
+                    'indra.sources.lincs_drug', 'indra.sources.ndex_cx',
```
bgyori, line 81: In addition to setup.py, we should also add an entry in the README for the new source.

indra/statements/statements.py
kkaris: The new statement types are now in a separate branch, PRd here: #1161

PR #1161, opened 2020-08-19
New modification statements

kkaris:
This PR adds new Modification statement types:

- Carboxylation/Decarboxylation
- Amidation/Deamidation
- Sulfation/Desulfation
- Neddylation/Deneddylation
- Alkylation/Dealkylation
- Prenylation/Deprenylation

### DIFF
--- a/indra/statements/statements.py
+++ b/indra/statements/statements.py
@@ -1024,49 +1024,61 @@ class Carboxylation(AddModification):
     """Carboxylation modification."""
     pass
 
+
 class Decarboxylation(RemoveModification):
     """Decarboxylation modification."""
     pass
+
 
 class Amidation(AddModification):
     """Amidation modification."""
     pass
 
+
 class Deamidation(RemoveModification):
     """Deamidation modification."""
     pass
+
 
 class Sulfation(AddModification):
     """Sulfation modification."""
     pass
 
+
 class Desulfation(RemoveModification):
     """Desulfation modification."""
     pass
+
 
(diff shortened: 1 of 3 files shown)


kkaris:
I found some places that potentially need to be updated if the new types are integrated:
- https://github.com/sorgerlab/indra/blob/master/indra/assemblers/pysb/sites.py
- https://github.com/sorgerlab/indra/blob/master/indra/tests/test_english_assembler.py
- https://github.com/sorgerlab/indra/blob/master/indra/sources/ctd/processor.py

indra/statements/statements.py
```diff
@@ -1005,6 +1020,66 @@ class Demethylation(RemoveModification):
     """Demethylation modification."""
 
 
+class Carboxylation(AddModification):
+    """Carboxylation modification."""
+    pass
```
cthoyt, line 1025: if you have a docstring, you actually don't need a `pass`

cthoyt:
Maybe it makes sense to add a bit of metadata along with each statement type to align them with related entries in other controlled vocabularies. For example, neddylation corresponds to the GO biological process "protein neddylation" (GO:0045116).

I did this for PyBEL and you could copy relevant parts if you think this is a good idea. No single controlled vocabulary is covering all types of modifications, though. See: https://github.com/pybel/pybel/blob/57b24a9ddf6d5d73d82c92294ed2186ddf4443c5/src/pybel/language.py#L320-L524

PR #1171, opened 2020-10-01
Covid grounding

steppi:
This PR adds entries to grounding resource files. A list of potentially ambiguous and incorrectly grounded terms relevant to the COVID-19 EMMAA model were identified through machine learned anomaly detection models. Entries included are based upon a curation of (agent text, grounding) pairs which either were involved in at least one statement used in an explanatory path, or for which there were at least 50 statements involving this (agent text, grounding) pair in COVID-19 EMMAA model. All such (agent_text, grounding) pairs with an anomaly detection score over a given threshold were curated. 

Unambiguous terms which were being incorrectly grounded have been added to the existing "covid_grounding.csv". Terms for which a statement may potentially be of interest were added to "misgrounding_map.csv", but I may have been too loose in this consideration and some of these may be better suited for inclusion in "ignore.csv" (e.g. "por" overwhelmingly appears in the sense of the spanish preposition, but there is a bacterial protein called por different from the existing incorrect grounding so "por" was added to the misgrounding_map).

At this moment I'm only ~90% confident in the correctness and utility of all of the new entries. Some things may be worthy of further discussion and it's possible some IDs may have been miskeyed.


bgyori:
Looks great, thanks @steppi. The only change we could make is ground `mice` to MeSH instead of adding it to the misgrounding map. Also, once finalized, please run `update_grounding_map()` in `indra.resources.update_resources` to get the actual merged grounding map file updated and then check that in.

indra/resources/grounding/covid_grounding.csv
```diff
@@ -22,3 +22,54 @@ spike protein,UP,P0DTC2
...
+pDC,MESH,D003713
+bat,MESH,D002685
+P-gp,HGNC,ABCB1
+PRRs,MESH,D051192
+fish,D005399
+MRI,MESH,D008279
+sham,MESH,D010919
+NoV,MESH,D029322
+NETs,MESH,D065206
+G 1,MESH,D059585
```
bgyori, line 42: We could make this D016193 "G1 Phase" instead

indra/resources/grounding/covid_grounding.csv
```diff
@@ -22,3 +22,54 @@ spike protein,UP,P0DTC2
...
+nsp1,PF,PF11501
+PGE2,CHEBI,CHEBI:15551
+sex,MESH,D012723
+miR,MESH,D035683
+catL,HGNC,CTSL
+app,MESH,D063731
+apps,MESH,D063731
+Xa,MESH,D015951
+Stx,MESH,D022603
+Ang II,CHEBI,CHEBI:48432
```
bgyori, line 58: Let's make this and the one below `UPPRO,PRO_0000032458`

indra/resources/grounding/covid_grounding.csv
```diff
@@ -22,3 +22,54 @@ spike protein,UP,P0DTC2
...
+ANG II,CHEBI,CHEBI:48432
+fat,MESH,D008055
+Ach,CHEBI,CHEBI:15355
+SAg,MESH,D018089
+itch,MESH,D011537
+Ad5,Taxonomy,28285
+NOx,CHEBI,CHEBI:35196
+male,MESH,D008297
+CaR,HGNC,CASR
+Ang-1,HGNC,ANGPT1
```
bgyori, line 68: I think `UPPRO,PRO_0000032457` would be better for `Ang-1`

PR #1173, opened 2020-10-06
Update HTML look

pagreene:
Update the look of the HTML produced by the HTML assembler, and improve the ability for outside utilities to customize the sources used in the display. This PR also adds `KNOWLEDGE_SOURCE_INFO`, where "biographic" information about our sources can be stored in a JSON-like (dict-of-dicts) format for use in various applications.


indra/config.py
```diff
@@ -136,3 +134,179 @@ def has_config(key):
     """
     return get_config(key, failure_ok=True) is not None
 
+
+KNOWLEDGE_SOURCE_INFO = {
```
johnbachman, line 138: Can we move this info to a separate resource file?

pagreene: I suppose we could add a file such as `source_info.json`. I would note this is also where we could add descriptions for each of the sources and the database API is already set up to incorporate them once we add them.

bgyori: Yes, source_info.json in resources would work, and I'd actually suggest putting the KNOWLEDGE_SOURCE_INFO variable into `indra/sources/__init__.py` since `config` is meant more for user-specific environmental configurations.

pagreene: :+1:

pagreene:
In addition to putting `source_info.json` into INDRA resources, I also gave INDRA resources and `__init__.py` file with function that returns a file handle to a given resource file, specified by name. Basically, it means you don't need to do complicated relative path shenanigans, just import the function, name the file, and don't worry about installation locations and stuff like that. It does also provide an importable global variable called `RESOURCES_PATH` if that is needed somewhere.

bgyori:
Oh okay, I actually added something similar in https://github.com/sorgerlab/indra/pull/1172/files#diff-a5abbb4c432c6c0c633014973e38d125

bgyori:
I now see that we haven't merged https://github.com/sorgerlab/indra/pull/1170 yet, I can merge that first and we can rebase/resolve any conflicts in this PR.

pagreene:
I suppose great minds think alike. :sweat_smile:

bgyori (approved):

PR #1174, opened 2020-10-07
Adeft connections

steppi:
This PR updates the preassembly process to allow a single database connection to be used for gathering text content during a preassembly run. The separate functions `run_adeft_disambiguation` and `run_gilda_disambiguation` have now become members of the class `DisambManager`. This class has ownership of its own database connection. This will require [PR134](https://github.com/indralab/indra_db/pull/134) to be merged into indra_db.

The following is a snapshot from PGAdmin of the number of database connections used while grounding mapping an old version of the EMMMA COVID-19 model using this PR branch.
adeft_connections
compare this with a snapshot while using the current master branch
master

Performance has also improved substantially. Grounding mapping a sample of 20,000 statements from this model took 3min3s on a t3.2 ec2 instance using the master branch compared to 34.6 seconds using the PR branch. The output when mapping these 20,000 statements was identical (up to matches keys) between the two branches. Before merging we should verify that there are no edge cases that have not been handled.


bgyori:
This looks great! I think we should do a test run on an EMMAA model to see if there are any corner cases that come up... I'll report back if I find anything.

PR #1177, opened 2020-10-12
Reimplementation / significant speedup of preassembly algorithm

bgyori:
This PR reimplements the refinement finding step of preassembly from scratch. Rather than focus on connected components in the ontology as a way to group statements, here, we explicitly exploit the graph structure of isa/partof relationships in the ontology to rapidly find possible refinement relationships for each statement before doing actual comparisons. We first separate statements by type, and then create indices of each agent's position in the ontology, stratified by agent role (e.g., subj, obj, etc.) while also handling None agents and agents without grounding corresponding to a node in the ontology graph. Using another index to link from agents to statements, using BFS and set operations, we can determine, for each statement, a set of statements it can possibly refine. We then run these comparisons and save the ones that are valid.

The improvement to assembly performance can be several orders of magnitude, in particular, when it comes to statements in which all agents are grounded to fully connected and large ontologies (e.g., CHEBI/GO/MESH/WM, etc) where the connected component-based algorithm required an all-by-all comparison among statements. In the World Modelers setting, where every entity is grounded to a single, connected ontology, this new algorithm opens up refinement finding to use cases that could previously not be considered.

- Example 1 in the biology setting: For a 1 million statement sample of de-duplicated statements from the INDRA DB, the old algorithm required ~218 million comparisons and took 45 minutes overall to find 29,135 support links. The algorithm in this PR requires just ~160 thousand comparisons (1,000x+ decrease) and takes 3 minutes overall (**~15x speedup**). Interestingly, it also finds one additional Conversion refinement that the old method incorrectly missed. The speedup is "just" 15x because currently these statements are enriched for proteins but as we get more statements whose agents are grounded to e.g., CHEBI and MeSH, the new approach here will make a very big difference.
- Example 2 in the biology setting: The EMMAA COVID19 model in its last assembly run took around 5 hours to find statement refinements. The new approach finds all refinements in less than 4 minutes, which is a **speedup of 75x**.
- Example 1 in the world modelers setting: the number of refinement comparisons for a ~500 document-based (fairly small) corpus of Events and Influences goes from 27,430,902 (20 minutes) to 1,152 (< 1 second), i.e., more than a **1,000x speedup**.
- Example 2 in the world modelers setting: Refinement finding for extractions from a 39k document set would have  required 1,895,992,849 comparisons (estimated to take ~1day), and was therefore disabled in a recent assembly run. With this new setup, the refinement step takes less than 3 seconds, for an estimated **speedup of 30,000x**.

Another advantage here is that the ontology no longer requires connected component labels to be constructed (the slowest part of ontology constructions) which means that the BioOntology graph can be constructed in ~40s, down from ~2 minutes.


bgyori:
Few more remarks:
- I didn't mention above that one change in the PR is an extra argument to `refinement_of` called `entities_refined` (bool), which, when set to True, tells the `refinement_of` function that the agents of the statements being checked are guaranteed to be in a refinement relationship (either matching or hierarchically related) and therefore, we don't need to again invoke any code to check this relationship inside the `refinement_of` function. Given the transitive closure optimization in IndraOntology, this doesn't have a huge effect in practice and so it it causes complications, could be removed.
- Parallelization is not implemented anymore internally in the Preassembler. It will be more productive and robust to implement any necessary parallelization external to the Preassembler, for instance, by separating statements by type or construct batches of statements in surrounding code for any given application.
- Currently, the function that maps agents to nodes in the ontology graph is fixed, this could be made a parameter of the Preassembler (similar to `refinement_fun` and `matches_fun`) to allow non-standard, custom ways of mapping agents and their grounding to an instance of IndraOntology.
- Given that the input to refinement finding is a set of de-duplicated statements that have **unique** matches keys / matches hashes with respect to a given `matches_fun`, this PR uses hashes in all the internal data structures to identify statements and refinements. For backwards compatibility, these hashes are then mapped to lined indices in the input statement list. There is only one application that uses this list directly, the INDRA DB, which, as soon as it gets the linear indices, maps these back to hashes. This back and forth conversion is redundant, so in a subsequent PR, the Preassembler could be changed to return refinements as tuples of statement hashes that the DB code can use directly without further conversions needed.

PR #1178, opened 2020-10-13
Implement split index for new preassembly

bgyori:
This is a follow-up to #1177, which reinstates the concept of a `split_idx` in preassembly, which allows the INDRA DB assembly to compare inter-group (but not intra-group) refinements between two group of statements. Things would have worked without this but this will, in fact, further speed up assembly for this special case.


pagreene:
I was actually wondering about this when I read the previous PR. This work looks amazing!

PR #1180, opened 2020-10-19
Covid grounding2

steppi:
This PR adds new entries to the covid_grounding.csv resource file and a single entry to misgrounding.csv. The gene CD2 has been added as a misgrounding for the text "SRBC". "SRBC" could be grounding mapped to "sheep red blood cells" but it appears there is no entry for this in any identifiers.org ontology.


johnbachman:
I wonder, do we actually want to keep a MESH grounding for ANOVA? I have trouble imagining it being relevant in most use cases and it might be hard to filter out of paths (for example) using semantic/ontological constraints given what we've discussed about the structure of the MESH hierarchy.

steppi:
Good point. I agree

bgyori:
Yes, though for instance `cytometry` is also related to a method, and we added a grounding for it before, so to standardize, we could move that to ignore too since it's not really viable as an argument to any INDRA Statement.

steppi:
> Yes, though for instance `cytometry` is also related to a method, and we added a grounding for it before, so to standardize, we could move that to ignore too since it's not really viable as an argument to any INDRA Statement.

Yes, this should also be in ignore. I was going to mention this one.

Issue #1184, opened 2020-10-27
Error on BioPAX Parsing of Reactome content


cthoyt:
It appears that there are some issues with none-types when processing content from Reactome. It can be reproduced with (pybiopax==0.0.4, indra==1.18.0):

```python
import pybiopax
import indra.sources.biopax
biopax_model = pybiopax.model_from_owl_url('https://reactome.org/ReactomeRESTfulAPI/RESTfulWS/biopaxExporter/Level3/1643685')
bp = indra.sources.biopax.process_model(model)
INFO: [2020-10-27 18:01:19] indra.ontology.bio.ontology - Loading INDRA bio ontology from cache at /Users/cthoyt/.indra/bio_ontology/1.2/bio_ontology.pkl
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/cthoyt/dev/bel/indra/indra/sources/biopax/api.py", line 208, in process_model
    bp.process_all()
  File "/Users/cthoyt/dev/bel/indra/indra/sources/biopax/processor.py", line 59, in process_all
    self.get_modifications()
  File "/Users/cthoyt/dev/bel/indra/indra/sources/biopax/processor.py", line 258, in get_modifications
    for enz, sub, gained_mods, lost_mods, \
  File "/Users/cthoyt/dev/bel/indra/indra/sources/biopax/processor.py", line 239, in _conversion_state_iter
    for inp, outp in self.find_matching_left_right(conversion):
  File "/Users/cthoyt/dev/bel/indra/indra/sources/biopax/processor.py", line 149, in find_matching_left_right
    return BiopaxProcessor.find_matching_entities(left_simple, right_simple)
  File "/Users/cthoyt/dev/bel/indra/indra/sources/biopax/processor.py", line 183, in find_matching_entities
    matches += BiopaxProcessor.find_matching_entities(inp_members,
  File "/Users/cthoyt/dev/bel/indra/indra/sources/biopax/processor.py", line 183, in find_matching_entities
    matches += BiopaxProcessor.find_matching_entities(inp_members,
  File "/Users/cthoyt/dev/bel/indra/indra/sources/biopax/processor.py", line 161, in find_matching_entities
    input_ers = {mpe.entity_reference.uid
  File "/Users/cthoyt/dev/bel/indra/indra/sources/biopax/processor.py", line 161, in <setcomp>
    input_ers = {mpe.entity_reference.uid
AttributeError: 'NoneType' object has no attribute 'uid'
[...]

PR #1185, opened 2020-10-28
Make sure PMID is converted to string in PyBEL Assembler

cthoyt:
This should address https://github.com/sorgerlab/indra/issues/1184#issuecomment-717783584

### DIFF
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -612,7 +612,7 @@ def _get_evidence(evidence):
     if evidence.pmid:
         citation = citation_dict(
             db=pc.CITATION_TYPE_PUBMED,
-            db_id=evidence.pmid,
+            db_id=str(evidence.pmid),
         )
     # If no PMID, include the interface and source_api for now--
     # in general this should probably be in the annotations for all evidence


johnbachman:
I wonder if we should handle this in the processor instead, because the PMIDs in the evidence objects should always be strings, not ints.

bgyori:
Yes, this is a Reacome/BioPAX-specific issue that is better handled in `indra.sources.biopax`. The new Statement validator (https://github.com/sorgerlab/indra/pull/1172) can be used to ensure that the statements ingested from a given source are valid, and we can make changes in the processor until these issues are addressed. I'll finish this up today.

bgyori:
Based on the things I looked into, this may not be related to how content is obtained from BioPAX, see my comment on #1184.

bgyori:
Given what we found in #1184 and #1187, is this PR still needed? I don't think any sources produce int PMIDs currently, especially with the validation module in place, so this may not be needed.

cthoyt:
All gucci! Not sure exactly what fixed it, but I'll take it.

Running on:

```python
>>> import indra.util.get_version
>>> indra.util.get_version.get_version()
'1.18.0-5a1ec8a80ea756adc6bcc2b65dbf5779631e8327'
```

Ran this code:

```python
import pybel
import pybiopax
import indra.sources.biopax

biopax_model = pybiopax.model_from_owl_url('https://reactome.org/ReactomeRESTfulAPI/RESTfulWS/biopaxExporter/Level3/1643685')
bp = indra.sources.biopax.process_model(biopax_model)
print(len(bp.statements))

graph = pybel.from_indra_statements(bp.statements)
graph.summarize()
indra vbf0d420b-1169-452d-9584-42da61cdd94b
Number of Nodes: 1370
Number of Edges: 4274
Number of Citations: 312
Number of Authors: 0
Network Density: 2.28E-03
Number of Components: 47
Number of Warnings: 0
```

PR #1187, opened 2020-10-28
Fix data schema of annotations in PyBEL assembler

cthoyt:
In PyBEL v14, the annotations data structure should be a `Mapping[str, Mapping[str, bool]]` where the first key is the name of the annotation, the nested key is the value of the annotation, and the bool is always just True (unless you have some negative annotations, then set it to false!). Even for annotations that only have a singleton value and the positive/negative dichotomy is irrelevant, it should still be done this way. Typically, even developers won't be accessing the networkx Graph class's `add_edge()` function and the PyBEL wrappers on it should handle of a variety of data structures, such as this one, but since we're accessing it directly in the PyBEL assembler, this is a problem.

Addresses https://github.com/sorgerlab/indra/issues/1184#issuecomment-718036895

### DIFF
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -635,8 +635,10 @@ def _get_evidence(evidence):
     for key, value in evidence.epistemics.items():
         if key == 'direct' or value is None:
             continue
-        annotations[key] = {v: True for v in value}
-
+        if isinstance(value, (list, set, tuple)):
+            annotations[key] = {v: True for v in value}
+        else:
+            annotations[key] = {value: True}
     return citation, text, annotations
 
 
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -626,16 +626,16 @@ def _get_evidence(evidence):
         )
 
     annotations = {
-        'source_hash': evidence.get_source_hash(),
+        'source_hash': {evidence.get_source_hash(): True},
     }
     if evidence.source_api:
-        annotations['source_api'] = evidence.source_api
+        annotations['source_api'] = {evidence.source_api: True}
     if evidence.source_id:
-        annotations['source_id'] = evidence.source_id
+        annotations['source_id'] = {evidence.source_id: True}
     for key, value in evidence.epistemics.items():
         if key == 'direct':
             continue
-        annotations[key] = value
+        annotations[key] = {value: True}
 
     return citation, text, annotations
 
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -400,7 +400,7 @@ def belgraph_to_signed_graph(
         rel = edge_data.get('relation')
         pos_edge = \
             (u, v, ('sign', 0)) + \
-            tuple((k, (tuple(v) if isinstance(v, list) else v))
+            tuple((k, tuple(v))
                   for k, v in edge_data.get('annotations', {}).items()) \
             if propagate_annotations else (u, v, ('sign', 0))
         # Unpack tuple pairs at indices >1 or they'll be in nested tuples
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -633,9 +633,9 @@ def _get_evidence(evidence):
     if evidence.source_id:
         annotations['source_id'] = {evidence.source_id: True}
     for key, value in evidence.epistemics.items():
-        if key == 'direct':
+        if key == 'direct' or value is None:
             continue
-        annotations[key] = {value: True}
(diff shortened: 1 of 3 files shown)


bgyori:
Thanks! This will require adapting some code in `indra.explanation.model_checker`, we will look into it.

cthoyt:
> Thanks! This will require adapting some code in `indra.explanation.model_checker`, we will look into it.

The last few commits should address some of this. Will wait to see what Travis thinks. However, it seems like there are other unrelated tests failing

indra/assemblers/pybel/assembler.py
```diff
@@ -400,7 +400,7 @@ def get_ns(n):
         rel = edge_data.get('relation')
         pos_edge = \
             (u, v, ('sign', 0)) + \
-            tuple((k, (tuple(v) if isinstance(v, list) else v))
+            tuple((k, tuple(v))
```
cthoyt, line 403: We should assume that all `v` from this annotations dictionary are now of type `Mapping[str,bool]` where the keys are the actual values of the annotation, and the "values" are booleans saying whether they're true or not. Just assume always true.

indra/explanation/reporting.py
```diff
@@ -173,7 +173,7 @@ def stmts_from_pybel_path(path, model, from_db=True, stmts=None):
         hashes = set()
         for j in range(len(edges)):
             try:
-                hashes.add(edges[j]['annotations']['stmt_hash'])
+                hashes.add(list(edges[j]['annotations']['stmt_hash'])[0])
```
cthoyt, line 176: We're assuming that `edges[j]['annotations']['stmt_hash']` is of type `Mapping[str,bool]` and even further that it has exactly one key. This change extracts that key

cthoyt:
It looks like all of the PyBEL related tests are passing now

bgyori:
Looks good to me! @dianakolusheva do you want to take a quick look to see if there's anything related to EMMAA tests that this could affect and isn't covered by tests, just in case?

indra/assemblers/pybel/assembler.py
```diff
@@ -626,16 +626,16 @@ def _get_evidence(evidence):
...
-        annotations['source_api'] = evidence.source_api
+        annotations['source_api'] = {evidence.source_api: True}
     if evidence.source_id:
-        annotations['source_id'] = evidence.source_id
+        annotations['source_id'] = {evidence.source_id: True}
     for key, value in evidence.epistemics.items():
         if key == 'direct':
             continue
-        annotations[key] = value
+        annotations[key] = {value: True}
```
dianakolusheva, line 638: The value here is None or a list. As list is unhashable, we can replace it with a tuple. I am not sure how we'd want to handle None. {None: True}?

dianakolusheva:
Other than the small issue with annotating epistemics everything seems to be working fine!

indra/assemblers/pybel/assembler.py
cthoyt: so the type of `evidence.epistemics` is `Mapping[str, Union[None, List[str]]]`?

dianakolusheva: I believe the type is `Mapping[str, Union[None, bool, List[str]]]`, so it can be a boolean as well. @bgyori can you confirm this?

bgyori: Yes, that's right, values in the epistemics dict are sometimes lists (e.g., from Eidos), which would error here, we just don't have a test to cover that yet. Just to make this more concrete, an example entry is
```
{'hedgings': ['might', 'consider']}
```
Should we skip these annotations (given that they aren't critical for PyBEL analysis) or should we make a sorted tuple out of the value?

bgyori: Actually, I think `annotations[key] = {v: True for v in value}` as @cthoyt suggested could work. I'm not sure we need special handling for None and bool (i.e., `if key == 'direct' or value is None`), those are valid dict keys, or am I missing something?

cthoyt: Adding `None`'s will be an issue, so `if value is None: continue` is important. The `key == 'direct'` check is skipped because that part of the epistemics dictionary is used elsewhere to determine the edge type, so it would be redundant here.

cthoyt: I made the update in https://github.com/sorgerlab/indra/pull/1187/files#diff-01513c4320c875fa208bd896b195470920e0bf2415486994cacbb3077eff7ace, but I think it would be good to include a test that covers this

cthoyt: Full solution (that's robust to lists/scalars) and corresponding tests added in 615774d63 and 47183edd1

bgyori: Thanks @cthoyt, will merge once Travis is done!

PR #1191, opened 2020-11-02
Fix Statements View

pagreene:
Fix some minor issues in the display of statements using the old interface. (This PR was created using PyCharm)


johnbachman:
This is already deployed on the latest DB endpoint, correct?

pagreene:
Correct, this is the code currently deployed on latest (which I just fixed a crippling bug resulting from an unrelated issue).

PR #1192, opened 2020-11-05
Handle curation JSON in assemble_corpus.

pagreene:
This PR goes hand-in-hand with https://github.com/indralab/indra_db/pull/143 in `indra_db`, in which the results from `get_curations` are returned as JSON rather then objects which leak database connections.


bgyori:
Thanks!

bgyori:
Oops, I forgot that REST API tests are now separated out and one failed, I'll push a fix.

PR #1193, opened 2020-11-06
Improvements in reader processors

bgyori:
This PR makes small improvements to reader processors as follows:
- Handle new EKB tags for locations that changed in the latest version of TRIPS/DRUM
- Handle ActiveForm extraction from the latest version of TRIPS/DRUM, which produces an extra HAVE-PROPERTY event
- Remove `UNKNOWN` agent modification conditions when processing statements from Reach
- Update Reach documentation to update instructions for running the web service on the latest Reach master


bgyori:
The test failures are due to an unrelated web service

PR #1196, opened 2020-11-12
Fix pbseg edges

kkaris:
This PR fixes a bug were edge data in signed graphs assembled from Bel graphs were represented as 1-tuples instead of their values.

### DIFF
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -400,8 +400,8 @@ def belgraph_to_signed_graph(
         rel = edge_data.get('relation')
         pos_edge = \
             (u, v, ('sign', 0)) + \
-            tuple((k, *tuple(v))
-                  for k, v in edge_data.get('annotations', {}).items()) \
+            tuple((key, *tuple(entry))
+                  for key, entry in edge_data.get('annotations', {}).items()) \
             if propagate_annotations else (u, v, ('sign', 0))
         # Unpack tuple pairs at indices >1 or they'll be in nested tuples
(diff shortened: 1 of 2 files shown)


bgyori:
@cthoyt, when you standardized the annotation values to tuples in this commit: https://github.com/sorgerlab/indra/commit/206496f06c8aa60609aff4ffa20d55f83a1e8b16, was that something required by PyBEL? Or is it something that we can change like @kkaris did in this PR?

cthoyt:
Yes, this dict format is required by PyBEL for I/O to work. Normally, there's the high level API (e.g., `BELGraph.add_directly_increases()`) that protects users and automatically updates the data structure from some common ones (like a simple dict or list), but the PyBEL processor currently uses the low-level API provided by NetworkX (i.e., `Graph.add_edge()`), so these checks aren't done.

Sorry this is causing confusion - when I made this update, I was pretty sure that I updated the code related to other components. Maybe this was something not covered by a unit test, and I wasn't aware?

To be honest, I just got back to working on the PyBEL 15 PR for INDRA and I might have mixed up a few things. I wasn't happy with this data structure in PyBEL and for PyBEL 15 opted to have a list of `pybel.language.Entity` objects instead

bgyori:
Sorry, I was a bit confused, this change is actually inside a function called `belgraph_to_signed_graph` which is an _export_ of the PyBEL graph, so changes in here can't cause any potential problems with the PyBEL graph data structure.

bgyori (approved):

dianakolusheva (approved):

cthoyt:
All's well that ends well! Thanks for checking in either way :)

PR #1199, opened 2020-11-19
Enable specification of BioPAX OWL file encoding

cthoyt:
As a follow-up to https://github.com/indralab/pybiopax/pull/7, this uses the new PyBioPAX v0.0.5 functionality to specify the file encoding when opening an OWL file. This is useful for Windows users, whose default encoding is wack (cp1252 instead of utf-8).


bgyori:
Thanks!

PR #1200, opened 2020-11-19
Migrate testing from Travis to GitHub actions

bgyori:
This PR implements the essential parts of the automated test environment in GitHub actions. Some non-essential parts of the build have not yet been implemented (e.g., setting up scheduled tests that also run tests tagged as `slow`) so these can be revisited later (some of the Travis code is left commented out in the new workflow file for reference).


bgyori:
Looks like the action wasn't triggered on this repo yet but the tests passed for the branch at https://github.com/bgyori/indra/runs/1426196034?check_suite_focus=true.

PR #1209, opened 2020-12-04
Improve BioGRID processing

### LINKED ISSUE #1207: Biogrid processor should get latest release
The Biogrid processor currently has a hard-coded URL for a particular release of the data. On the website, there is a hyperlink "Current-Release" that maps to a (webpage) directory with the latest release. Ideally we would make use of this information to get the latest release automatically.

bgyori:
This PR improves the processing of BioGRID content in several ways:
- Getting content directly from the latest link without having to hard-code the version. Fixes #1207.
- Migrating to the current standard format which is tab3 (https://wiki.thebiogrid.org/doku.php/biogrid_tab_version_3.0) and has important additional information.
- Agent names are taken from the symbol field (always available, typically a gene name) instead of the systematic name (usually not a canonical name and not always available). Since statements with missing Agent names were skipped before, this change adds ~10% new statements.
- Agent db_refs now make use of Entrez IDs as well as reviewed or unreviewed UniProt IDs (important for non-human genes).
- Text refs can represent PMIDs as well as DOIs, either of which can be present in each row


cthoyt:
You might also take inspiration from the BioGRID -> BEL converter in Bio2BEL https://github.com/bio2bel/bio2bel/blob/master/src/bio2bel/sources/biogrid.py

bgyori:
Thanks @cthoyt, interesting, looks like you built on PSI-MI 2.5, whereas in INDRA we first used TAB 2.0 and now migrated to TAB 3.0 (these are BioGRID's own formats).

cthoyt:
In general I was hoping to reuse a PSI-MI importer for other databases as well. I had discussed some of the other standards with @vtoure in the summer and how they could be interconverted with BEL/related formats, but she was busy finishing her PhD (all done now!)

PR #1210, opened 2020-12-05
Add Entrez mappings to ontology graph

bgyori:
This PR adds mappings between Entrez and HGNC to the ontology graph so that these can be handled by the standardization process instead of having to be looked up and applied manually via the hgnc_client.

### DIFF
--- a/indra/tests/make_mock_ontology.py
+++ b/indra/tests/make_mock_ontology.py
@@ -2,6 +2,7 @@
 which can be put in the appropriate cache location in place of the
 real bio ontology for testing purposes"""
 import os
+import boto3
 import pickle
 from indra.ontology.bio.ontology import BioOntology, CACHE_DIR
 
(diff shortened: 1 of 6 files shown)


indra/tests/make_mock_ontology.py
```diff
@@ -75,4 +76,10 @@ def keep_node(node):
     bio_ontology._build_transitive_closure()
     fname = os.path.join(CACHE_DIR, 'mock_ontology.pkl')
     with open(fname, 'wb') as fh:
-        pickle.dump(bio_ontology, fh, protocol=pickle.HIGHEST_PROTOCOL)
+        pickle.dump(bio_ontology, fh, protocol=4)
```
cthoyt, line 79: sad, pickle protocol 5 ftw

bgyori: Everyone loves protocol 5 except python 3.7 or anything below :)

cthoyt: Check out https://pypi.org/project/pickle5/

PR #1214, opened 2020-12-11
Belief sorting, sort by function, generalize sorting algorithm.

pagreene:
This PR completely re-tools the `statement_presentation` module to enable sorting by arbitrary parameters, or functions of parameters, that are associated with statements and groupings of statements. The key new features are:
- Allow arbitrary data to be aggregated for groups of statements by arbitrary methods (e.g. sum, max, ave, custom model)
- Allow for sorting by any of the given parameters, e.g. 'ev_count' or 'belief'.
- Allow for sorting by an arbitrary function of the given parameters.
- Allow for arbitrary levels of statement grouping, at least at the `statement_presentation` level.

EDIT: Since the original PR, some more features have been added:
- Fix a bug in the isolated statement display's expand/collapse all feature
- Allow the collapse of statement evidence
- Allow `sort_by=None` option to maintain input order of statements. No guarantees made about the results when using this feature along with grouping by relation or agent-pair.

### DIFF
--- a/indra/assemblers/html/assembler.py
+++ b/indra/assemblers/html/assembler.py
@@ -194,14 +194,17 @@ class HtmlAssembler(object):
                                               metric_dict=metric_dict,
                                               sort_by=self.sort_by)
 
-        # Do some extra formatting.
+        # Set up some data structures to gather results.
         stmts = OrderedDict()
         agents = {}
         previous_stmt_set = set()
         all_previous_stmts = set()
         source_count_keys = set() if not self.source_counts \
             else {k for k in next(iter(self.source_counts.values())).keys()}
-        for key, inps, verb, stmts_group, agp_m, rel_m in stmt_rows:
+
+        # Loop through all the rows, gathering as we go.
+        for _, agp_key, rel_key, stmts_group, agp_m, rel_m in stmt_rows:
+
             # Distinguish between the cases with source counts and without.
             if self.source_counts:
                 rel_src_counts = {k: rel_m[k] for k in source_count_keys}
@@ -272,12 +275,12 @@ class HtmlAssembler(object):
                         del ag.db_refs[dbn]
 
             # Update the top level grouping.
-            if isinstance(stmt, (ActiveForm, HasActivity)):
-                agent_pair_names = [inps[0]]
-            elif isinstance(stmt, Conversion) and isinstance(inps[1], tuple):
-                agent_pair_names = [inps[0]] + [*inps[1]] + [*inps[2]]
+            if rel_key[0] == 'Conversion' and isinstance(agp_key[1], tuple):
+                agent_pair_names = [agp_key[0]] + [*agp_key[1]] + [*agp_key[2]]
             else:
-                agent_pair_names = inps[:]
+                agent_pair_names = agp_key[:]
+
+            # Form group key label
             if with_grouping:
                 agp_key = '-'.join([str(name) for name in agent_pair_names])
                 agent_pair_agents = {name: Agent(name)
@@ -294,6 +297,7 @@ class HtmlAssembler(object):
                 agp_label = 'All Statements'
                 agent_pair_agents = None
 
+            # Select where to put these statements.
             if agp_key not in stmts.keys():
                 agents[agp_key] = agent_pair_agents
                 stmts[agp_key] = {'html_key': str(uuid.uuid4()),
--- a/indra/assemblers/html/assembler.py
+++ b/indra/assemblers/html/assembler.py
@@ -20,7 +20,7 @@ from indra.statements.validate import validate_id
 from indra.databases.identifiers import get_identifiers_url, ensure_prefix
 from indra.assemblers.english import EnglishAssembler, AgentWithCoordinates
 from indra.util.statement_presentation import group_and_sort_statements, \
-    make_top_level_label_from_names_key, make_stmt_from_sort_key, \
+    make_top_level_label_from_names_key, make_stmt_from_relation_key, \
     reader_sources, db_sources, all_sources, get_available_source_counts, \
     get_available_ev_counts, standardize_counts, get_available_beliefs, \
     merge_to_metric_dict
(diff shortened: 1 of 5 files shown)


pagreene:
I still want to write some tests, especially of the various new features. However it appears to at least be supporting past functionalities.

indra/util/statement_presentation.py
```diff
@@ -91,100 +517,137 @@ def group_and_sort_statements(stmt_list, ev_totals=None, source_counts=None):
...
     StmtClass = get_statement_by_name(verb)
-    inps = list(key[1])
     if agents is None:
         agents = []
     if verb == 'Complex':
         agents.extend([make_agent(name) for name in inps])
         stmt = StmtClass(agents[:])
     elif verb == 'Conversion':
+        if isinstance(inps[1], str):
+            pass
```
pagreene, line 650: Note to self, figure out what you were thinking here and finish it if necessary.

johnbachman:
Is this ready for review or should it be considered WIP?

pagreene:
I would say it is still in WIP.

pagreene:
I think it is ready for review now. Check to see that you like the results of sorting in various cases. It should now be pretty easy to tweak the ordering of output (far more than it used to be).

I have also fixed a bug I came across in the expand/collapse all feature that led to extremely bad, sporadic behavior.

indra/assemblers/html/assembler.py
```diff
@@ -106,6 +111,27 @@ class HtmlAssembler(object):
         the configuration entry indra.config.get_config('INDRA_DB_REST_URL').
         If None, the URLs are constructed as relative links.
         Default: None
+    sort_by : str or function or None
+        If str, it indicates which parameter in metric_dict to sort by, or
```
johnbachman, line 115: It's not clear when reading this what metric_dict is as it is not a parameter or an attribute.

pagreene: Good catch, I think this bit of doc string got out of date.

indra/util/statement_presentation.py
```diff
@@ -1,184 +1,671 @@
...
 internal_source_mappings = {
     'bel': 'bel_lc'
 }
 
-
 all_sources = db_sources + reader_sources
 
 
-def _get_keyed_stmts(stmt_list):
+def _get_relation_keyed_stmts(stmt_list, grouping_level='agent-pair'):
```
johnbachman, line 29: Can you add a docstring for this function?

indra/util/statement_presentation.py
```diff
@@ -1,184 +1,671 @@
...
+    """
+
+    def __init__(self, name, data, data_type, agg_class):
+        self.name = name
+        self.data = data
+        self.data_type = data_type
+        self.agg_class = agg_class
+
+    @classmethod
+    def from_dicts(cls, dict_data, data_type, agg_class):
```
johnbachman, line 104: Needs documentation of the return type

indra/util/statement_presentation.py
```diff
@@ -1,184 +1,671 @@
...
+
+
+class StmtStatGather:
+    """Gather metrics for items that are derived from statements.
+
+    Example usage:
+    >>> # Get ev_count, belief, and ag_count from a list of statements.
+    >>> stmt_stats = StmtStat.from_stmts(stmt_list)
+    >>>
+    >>> # Add add another stat for a measure of relevance
```
johnbachman, line 209: Typo, "Add add"

indra/assemblers/html/assembler.py
```diff
@@ -106,6 +109,26 @@ class HtmlAssembler(object):
         the configuration entry indra.config.get_config('INDRA_DB_REST_URL').
         If None, the URLs are constructed as relative links.
         Default: None
+    sort_by : str or function or None
+        If str, it indicates which parameter to sort by, such as 'belief' or
+        'ev_count', or 'ag_count'. Those are the default options because they
+        can be derived from a list of statements, however if you give a custom
+        stmt_sort_gather, you may use any of the parameters used to build it.
```
johnbachman, line 116: Presumably this should be updated to read "if you provide customized attributes via the `custom_stats` argument"? (i.e., because stmt_sort_gather is not an arg of the assembler anymore)

indra/assemblers/html/assembler.py
```diff
@@ -174,163 +202,222 @@ def make_json_model(self, with_grouping=True, no_redundancy=False):
...
-                continue
-            elif no_redundancy and curr_stmt_set <= all_previous_stmts:
-                continue
+        source_count_keys = set() if not self.source_counts \
+            else {k for k in next(iter(self.source_counts.values())).keys()}
+
+        # Loop through the sorted and grouped statements.
+        all_hashes = set()
+
+        def handle_rows(rows, level, **kwargs):
```
johnbachman, line 222: Instead of kwargs, probably better to just add the explicit optional arguments `agent_key` and `agp_agents`, because those are the only ones that are ever provided to the function.

pagreene: They aren't the only arguments; they change depending on which level you are calling. I think putting in the rather arbitrary other arguments would actually serve to obfuscate the meaning of the function. `handle_rows` takes arguments of "rows", "level", and "maybe some other things". I really think that is the best way to think of this function.

indra/assemblers/html/assembler.py
pagreene: Good catch, I thought I got all of those updated references but obviously one slipped through.

pagreene: Ok, I did make these fixes. I must have forgotten to push before I re-loaded my OS. Oh well.

indra/util/statement_presentation.py
```diff
@@ -1,184 +1,822 @@
...
-        evidence as the values. Including this will allow statements to be
-        better sorted.
-    source_counts : dict{int: OrderedDict}
-        A dictionary, keyed by statement hash, with an OrderedDict of
-        counts per source (ordered).
+    sort_by : str or function or None
+        If str, it indicates which parameter to sort by, such as 'belief' or
+        'ev_count', or 'ag_count'. Those are the default options because they
+        can be derived from a list of statements, however if you give a custom
+        `stmt_metrics`, you may use any of the parameters used to build it.
```
johnbachman, line 671: Should this be "custom stats"?

johnbachman (approved):

PR #1221, opened 2020-12-19
Organism prioritization in Reach processor

bgyori:
This PR implements handling user-supplied organism prioritization when processing Reach output. Reach supplies both an `xrefs` (typically a single element) and an `alt-xrefs` list of groundings for entities it extracts. In the case of proteins, `xrefs` is prioritized to be a human protein and `alt-xrefs` is a (sometimes long) list of alternative matches to proteins that can sometimes include other human proteins or non-human proteins. Before this PR, we didn't make use of the `alt-xrefs` list, and instead took what was given in `xrefs`.

This PR adds an optional `organism_priority` argument to the REACH API functions and the Processor class which allows re-prioritizing organisms when extracting the UniProt ID for proteins in the given reading/processing call. The `organism_priority` is a list of Taxonomy IDs which https://github.com/indralab/protmapper/pull/32 allows checking for each protein in the grounding list to choose the prioritized one.

### DIFF
--- a/indra/sources/reach/api.py
+++ b/indra/sources/reach/api.py
@@ -33,7 +33,8 @@ default_output_fname = 'reach_output.json'
 
 
 def process_pmc(pmc_id, offline=False, url=None,
-                output_fname=default_output_fname):
+                output_fname=default_output_fname,
+                organism_priority=None):
     """Return a ReachProcessor by processing a paper with a given PMC id.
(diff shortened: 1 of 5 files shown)


bgyori:
It is now implemented in a way that it generalizes to protein chain/fragment grounding as well.

PR #1222, opened 2020-12-20
Update make df 

kkaris:
This PR adds two columns to the DataFrame produced by `IndraNetAssembler.make_df`: `residue` and `position`. Unknown/unspecified values are treated as NaN's. The update allows for these two attributes to be propagated to the sif dump from the indra db.

### DIFF
--- a/indra/assemblers/indranet/assembler.py
+++ b/indra/assemblers/indranet/assembler.py
@@ -168,6 +168,12 @@ class IndraNetAssembler():
                 The belief score associated with the statement.
             *source_counts*
                 The number of evidences per input source for the statement.
+            *residue*
+                If applicable, the amino acid residue being modified. NaN if
+                if it is unknown or unspecified/not applicable.
+            *position*
+                If applicable, the position of the modified amino acid. NaN
+                if it is unknown or unspecified/not applicable.
             *initial_sign*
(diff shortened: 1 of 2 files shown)


johnbachman:
Just tested these on the signor statements (returned by indra.sources.signor.process_from_web()). Residue and position look good, but I noticed that the initial_sign column was None.

indra/assemblers/indranet/assembler.py
```diff
@@ -236,13 +242,23 @@ def make_df(self, exclude_stmts=None, complex_members=3):
                 agA_ns, agA_id = get_ag_ns_id(agA)
```
johnbachman, line 242: On lines 239-240 above, anything that is not an Influence, Association, Complex or Conversion is given a sign `None`. Presumably there should also be an if block handling Activation/Inhibition and Increase/DecreaseAmount to assign correct polarity. We should also add appropriate tests for these statements and others for which sign should properly be None (e.g. Modifications).

johnbachman: Given that the existing sif dump pipeline (that doesn't use the IndraNet) doesn't include sign information, changing this is probably out of the scope of this PR and we can address later.

PR #1225, opened 2021-01-20
Extend HTML statement header width

bgyori:
This PR changes the column class of headers in the HTML assembler template from `col` to `col-auto` to allow wider headings to extend beyond the center.

Before:
image

After:
image

I am not sure, however, if this has unintended consequences so have to think about that before merging.


pagreene:
I feel like I have made this change in several places already (I had thought it was in all places), so it should be a good change. I think the only thing to check is that I didn't already make the change in an outstanding PR.

pagreene:
Can confirm this is is new to the code. For my part, I think we should merge it as soon as the tests finish running.

PR #1226, opened 2021-01-25
Get titles from PMCID

dianakolusheva:
This PR adds a function to pmc_client to extract title only given a PMCID


bgyori:
Thanks @dianakolusheva, we could also add a test for this, for instance with the `PMC4322985` ID which appears in some of the other PMC client tests.

PR #1227, opened 2021-01-25
Get more complex titles

dianakolusheva:
This PR fixes an edge case in pubmed_client when we need to iterate to get full title even when the tag.text contains some text.

### DIFF
--- a/indra/literature/pubmed_client.py
+++ b/indra/literature/pubmed_client.py
@@ -264,7 +264,7 @@ def _get_title_from_article_element(article):
     if title_tag is not None:
         title = title_tag.text
         if hasattr(title_tag, 'itertext'):
-            title = ' '.join(list(title_tag.itertext()))
+            title = ''.join(list(title_tag.itertext()))
     return title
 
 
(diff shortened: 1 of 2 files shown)


bgyori:
This looks good, thanks! Looking at the text example for reference, it might be the case that don't need to add spaces between parts of the title, i.e.,
```
Atomic structures of respiratory complex III2, complex IV, and supercomplex III2-IV from vascular plants.
```
instead of
```
Atomic structures of respiratory complex III 2 , complex IV, and supercomplex III 2 -IV from vascular plants.
```
at least in this example. But I'm not sure what some other examples are that we could look at to see which joining is better.

dianakolusheva:
> This looks good, thanks! Looking at the test example for reference, it might be the case that we don't need to add spaces between parts of the title, i.e.,
> 
> ```
> Atomic structures of respiratory complex III2, complex IV, and supercomplex III2-IV from vascular plants.
> ```
> 
> instead of
> 
> ```
> Atomic structures of respiratory complex III 2 , complex IV, and supercomplex III 2 -IV from vascular plants.
> ```
> 
> at least in this example. But I'm not sure what some other examples are that we could look at to see which joining is better.

I have several examples, will check

dianakolusheva:
Confirmed that joining on space adds a space inside the entity name when there's a subscript or superscript, removed the space. Examples of such entities: `Ca2+`, `T1AM-TAAR1`, `KrasG12D` (vs `Ca 2+`, `T 1 AM-TAAR1`, `Kras G12D` if the space is used)

PR #1228, opened 2021-01-27
Update the INDRA client to the INDRA Database

pagreene:
This PR re-imagines what is possible with the INDRA Client, making use of the new constructable query class system developed in the INDRA Database, this PR enables the user to submit arbitrary queries to the database. Furthermore, any query will now be able to be paged, in much the same way that subject-verb-object queries were. In fact, this PR is able to make use of improvements in the back end to significantly simplify the paging process.

This PR also does away with the old `simple_response` option, which has been on a deprecation path for a long time now. All API calls will now return a processor, just as other source API's do.

This PR goes along with https://github.com/indralab/indra_db/pull/161


indra/assemblers/html/assembler.py
```diff
@@ -634,6 +634,7 @@ def get_role(ag_ix):
                         'text_refs': text_refs,
                         'text': format_text,
                         'source_hash': str(ev.source_hash),
+                        'original_json': ev.to_json(),
```
dianakolusheva, line 637: This creates OrderedDict that cannot be rendered by Vue

johnbachman (approved):

Issue #1230, opened 2021-02-09
UnicodeDecodeError: 'charmap' codec with REACH


zebulon2:
Hello,

I have an issue when reading a PMC article with REACH using this simple code (using indra 1.19.0 and 1.18.0):
```
from indra.sources import reach
reach_processor = reach.process_pmc('3717945')
```
returns:
```
WARNING: [2021-02-09 10:30:52] indra.sources.reach.api - Could not import jnius, offline reading option will not be available.
INFO: [2021-02-09 10:30:52] indra.sources.reach.api - Loading 3717945 from PMC
INFO: [2021-02-09 10:30:54] indra.sources.reach.api - Looking up PMID for 3717945
INFO: [2021-02-09 10:30:55] indra.sources.reach.api - Processing 3717945 with REACH
Traceback (most recent call last):
  File "C:/Users/xxx/PycharmProjects/indra18/reach.py", line 14, in <module>
    reach_processor = reach.process_pmc('3717945')
  File "C:\Users\xxx\PycharmProjects\indra18\venv\lib\site-packages\indra\sources\reach\api.py", line 85, in process_pmc
    output_fname=output_fname)
  File "C:\Users\xxx\PycharmProjects\indra18\venv\lib\site-packages\indra\sources\reach\api.py", line 287, in process_nxml_file
    json_str = _read_nxml_file_service_old(file_name, url=reach_nxml_url)
  File "C:\Users\xxx\PycharmProjects\indra18\venv\lib\site-packages\indra\sources\reach\api.py", line 451, in _read_nxml_file_service_old
    nxml_str = fh.read()
  File "C:\Program Files\Python36\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 8667: character maps to <undefined>
```
I was able to fix it simply by editing `indra\sources\reach\api.py` (lines 494-495) and specifiying the encoding as utf8 in the open fct:
```
def _read_nxml_file_service_old(nxml_file, url=reach_nxml_url):
    with open(nxml_file, 'r', encoding='utf8') as fh:
```
Is it the "proper" way to fix this issue?

bgyori:
Hi @zebulon2, yes, that change makes sense. I think this happens because utf-8 isn't set as the default encoding in your environment so it needs to be added here explicitly.

zebulon2:
[...]

PR #1233, opened 2021-02-11
Explicitly specify utf-8 for reach read service

zebulon2:
In case utf-8 is not the system default encoding, a reach read file service would throw a UnicodeDecodeError


zebulon2:
This is to resolve issue #1230

bgyori:
Thanks!

PR #1234, opened 2021-02-11
Map db_refs

dianakolusheva:
This PR adds a new assemble corpus function to overwrite IDs in db_refs with new provided ones. It also adds a resource file that can be loaded as default db_refs_map and a test for a new function.


bgyori:
Looks, good, thanks! At some point, I might make the default map part of the ontology graph since such mappings could be applied more systematically to replace deprecated groundings with their current equivalents.

PR #1236, opened 2021-02-16
Fix bug in preassembly splitting.

pagreene:
Fix a minor bug in the preassembly code in which `<=` was used in place of `<`, causing a few statements at the edge to not be compared appropriately. This bug was discovered over the course of resurrecting database test code,

### DIFF
--- a/indra/preassembler/__init__.py
+++ b/indra/preassembler/__init__.py
@@ -395,7 +395,7 @@ class Preassembler(object):
         if split_idx:
             # This dict maps statement hashes to a bool value based on which
             # of the two groups the statement belongs to.
-            hash_to_split_group = {sh: (idx <= split_idx) for sh, idx
+            hash_to_split_group = {sh: (idx < split_idx) for sh, idx
                                    in stmt_to_idx.items()}
         else:
             hash_to_split_group = None
--- a/indra/tests/test_preassembler.py
+++ b/indra/tests/test_preassembler.py
@@ -1020,7 +1020,7 @@ def test_split_idx():
     assert (2, 0) in maps, maps
     assert pa._comparison_counter == 2
     pa = Preassembler(bio_ontology)
-    maps = pa._generate_id_maps([st1, st2, st3], split_idx=1)
+    maps = pa._generate_id_maps([st1, st2, st3], split_idx=2)
     assert (2, 0) in maps, maps
     assert (1, 0) not in maps, maps
     assert pa._comparison_counter == 1


bgyori:
Interesting, the old code before #1177 for this was
```
        for idx, stmt in stmt_tuples:	
            if idx <= split_idx:	
                stmt_group_a.append((idx, stmt))	
            else:	
                stmt_group_b.append((idx, stmt))	
```
so I'm surprised this came up as an issue but if this is the way it works, we can change it.

bgyori:
I guess if `<=` used to be the behavior here all along and not something added in #1177, an alternative would be to change the calling code at https://github.com/indralab/indra_db/blob/master/indra_db/preassembly/preassemble_db.py#L452 to match how split_idx is implemented here.

pagreene:
I think the behavior used to just be a list split, in which case "<" is
correct. Also i implemented this originally for the database, so if it
doesn't work with the database code it must have been changed. Indeed this
code is currently only a few months old.

bgyori:
I was looking at this commit as the original split_idx implementation https://github.com/sorgerlab/indra/commit/e253d9559d7ceb3b1397632c88705317f1fc1b23

bgyori:
I confirmed that the issue stems from the calling code: as of this commit, https://github.com/sorgerlab/indra/blame/e247be88a3638df511041ebcc16ff6745b7d80ec/indra/db/preassembly_script.py, split_idx was set correctly:

```python
split_idx = len(only_old_stmts) + 1
merge_stmts = only_old_stmts + only_new_stmts

# Find the support links between the new corpora
merge_support_links = self._get_support_links(merge_stmts,
                                                      split_idx=split_idx,
                                                      **kwargs)
```

but then this commit https://github.com/sorgerlab/indra/blame/e6a88e8da1b4038c7f426abbb62e3125697c7983/indra/db/preassembly_script.py didn't preserve the `+1` added to split_idx
```
split_idx = len(inner_batch)
support_links |= self._get_support_links(outer_batch+inner_batch,
                                                         split_idx=split_idx)
```
This code was then later moved to the separate INDRA DB repo.

Independently, I also confirmed that the behavior of `split_idx` in the Preassembler has not changed since its original implementation in   https://github.com/sorgerlab/indra/commit/e253d9559d7ceb3b1397632c88705317f1fc1b23, as confirmed by checking out that commit and running

```python
> from indra.preassembler.hierarchy_manager import hierarchies
> from indra.preassembler import Preassembler
> from indra.statements import *
> pa = Preassembler(hierarchies)
> ras = Agent('RAS', db_refs={'FPLX': 'RAS'})
> kras = Agent('KRAS', db_refs={'HGNC': '6407'})
> hras = Agent('HRAS', db_refs={'HGNC': '5173'})
> st1 = Phosphorylation(Agent('x'), ras)
> st2 = Phosphorylation(Agent('x'), kras)
> st3 = Phosphorylation(Agent('x'), hras)
> pa._generate_id_maps([st1, st2, st3], split_idx=0)
{(1, 0), (2, 0)}
> pa._generate_id_maps([st1, st2, st3], split_idx=1)
{(2, 0)}
> pa._generate_id_maps([st1, st2, st3], split_idx=2)
set()
```
Overall, this suggests that the calling code should be updated.

pagreene:
I still think using the <= is a goofy implementation requiring an inconsistent use of list indices. This was likely the result of a less experienced me being confused and haphazardly debugging things. Regardless of whether that was the original implementation, I think it is better as <, and I still believe the change should be made to make a more 0-indexed API. Given how otherwise inconsequential the change is, I see little harm in it. Unless of course there is other code besides the INDRA Database that uses this split index feature, in which case I agree changing the `indra_db` code makes more sense.

PR #1250, opened 2021-03-20
Improvements to grounding Minerva-derived statements

bgyori:
This PR makes improvements to grounding agents processed from Minerva models. It builds on new curated mappings I added to Biomappings between NCBI protein entries and UniProt chains to be able to standardize the names and groundings for SARS-CoV-2 protein fragments. It also more generally extends the set of name spaces handled.


bgyori:
There are still some issues, namely, there are a number of SARS-CoV-1 proteins in the models that aren't covered by the curated mappings for SARS-CoV-2. Also, there are some entities that represent multiple proteins, e.g., "Nsp3-4" and have multiple NCBI protein groundings available for the single entity. When we iterate over these, an arbitrary one is chosen and mapped, for instance, only Nsp3 when creating an Agent.

PR #1251, opened 2021-03-20
Grounding curation

steppi:
This PR adds a number of entries to the grounding resources that were found by manually reviewing HGNC groundings that were found to be highly anomalous by our anomaly detection framework and for which the agent texts were not all capital letter shortforms. A script has been written that allows such entries to be added directly from a curation spreadsheet. A new file,
"indra/resources/grounding/extra_gmap_locations.json" has been created which contains the filenames of all non-famplex grounding maps that live in "indra/resources/grounding/". `update_grounding_map` in `indra.resources.update_resources` reads these filenames when merging the different grounding maps. Previously the different grounding maps were hardcoded into the `update_grounding_map` function.


indra/resources/grounding/covid_grounding.csv
```diff
@@ -78,4 +78,4 @@ CpG,NCIT,C67101
 ACh,CHEBI,CHEBI:15355
 age,MESH,D000375
 Age,MESH,D000375
-CatL,HGNC,CTSL
```
cthoyt, line 81: maybe would be a good idea to have a unit test that validates the entries in these CSVs as having valid prefix/identifier pairs -> I just noticed that HGNC:CTSL is definitly not a valid HGNC identifier

steppi: I agree, although it's possible the tests are already in place and will soon fail. Also you've just reminded me of something important. For historical reasons, the HGNC entries in our grounding maps use HGNC symbols instead of HGNC IDs. I actually had forgotten about his when adding all of the new entries and used the IDs instead. The script I wrote to automatically add entries from a curation sheet does validate the pairs, but the validation expects IDs instead of symbols. I think having these seemingly arbitrary exceptions to standard rules makes the need for testing even more important.

steppi: I don't plan to work on this further over the weekend. On Monday, I will change the grounding map entries I've added with HGNC IDs to use HGNC symbols, change the update script to accept either ID or symbol in a curation spreadsheet, validate appropriately, and add entries to the grounding map files always using symbols. We can all discuss the testing situation during the standup.

bgyori: In this file the correct convention - for better or worse - is to use HGNC symbols. So this is okay as is.

steppi:
I think this is ready to merge

indra/resources/grounding/gene_grounding.csv
```diff
@@ -0,0 +1,223 @@
...
+GLUT3,HGNC,SLC2A3
+TG2,HGNC,TGM2
+PsA,MESH,D015535
+Atrial natriuretic factor,MESH,D009320
+SpA,UP,P02976
+lipid phosphatases,ECCODE,3.1.3.76
+rotenone,CHEBI,CHEBI:28201
+cyclin D1,HGNC,CCND1
+neoadjuvant chemotherapy,NCIT,C15665
+CT26,CLO,CLO:0002631
```
bgyori, line 223: Let's change this to: CT26,CVCL,CVCL_7254. This validates correctly, i.e.,
```python
> validate_ns('CVCL')
True

> validate_db_refs({'CVCL': 'CVCL_7254'})
True

> get_identifiers_url('CVCL', 'CVCL_7254')
'https://identifiers.org/cellosaurus:CVCL_7254'
```

indra/tests/test_grounding_resources.py
```diff
@@ -0,0 +1,44 @@
+import os
+import csv
+from indra.statements.validate import validate_db_refs, validate_ns
+from indra.preassembler.grounding_mapper import default_grounding_map
+from indra.preassembler.grounding_mapper import default_misgrounding_map
+
+
+# Namespaces that are not currently handled but still appear in statements
+exceptions = ['CLO']
```
bgyori, line 9: With the modification to the grounding to CLO I proposed above, this won't be necessary.

PR #1252, opened 2021-03-21
⚗️ Update ChEBI relations in the bioontology

### LINKED ISSUE #1249: Check if we should add more CHEBI relationships in the ontology graph
Here are the current relations from ChEBI mapped to `isa`:
https://github.com/sorgerlab/indra/blob/bad6f8767d0142cf988f11ec3382fe24d19401c8/indra/ontology/bio/ontology.py#L209-L214
are there any more we should add?

cthoyt:
Closes #1249 

This PR does the the following:

1. Adds the `is_conjugate_base_of` relationship from ChEBI
2. I double checked the other relationships used in ChEBI, and none of them make sense in this context. There are two relationships for tautomers and enantiomers, which are stereochemical relationships that imply structural similarity, but not necessarily functional similarity (for example, this is why [thalidomide](https://en.wikipedia.org/wiki/Thalidomide) was such an issue with causing birth defects)
3. Adds the `has part` relationship from ChEBI, but since this has the opposite polarity, I had to add a bit of extra information that tells the program when to reverse the source/target


bgyori:
The test that is failing now is actually of some consequence to how PathwayCommons content is processed... will have to think about what the right solution is for that case.

bgyori:
I'm having second thoughts about the `is_conjugate_base_of` and `is_conjugate_acid_of` edges turned into `isa` edges... looking at some examples, it's not such a good fit. I'm actually not sure why I added `is_conjugate_acid_of` in the first place, possibly, that was what solved a specific problem in processing PathwaysCommons content (see the failing test).

cthoyt:
I'm not against removing them

cthoyt:
@bgyori anything else we want to do here?

bgyori:
No, this is good, I am just working on another branch to avoid the BioPAX processing issues that this change would cause and will probably merge that first.

cthoyt:
> No, this is good, I am just working on another branch to avoid the BioPAX processing issues that this change would cause and will probably merge that first.

Cool, CC me on PR if you want me to take a look

bgyori (approved):

PR #1255, opened 2021-03-25
Improve small molecule grounding extraction from BioPAX

bgyori:
This PR makes two changes to how small molecule groundings are extracted from BioPAX content. First, if an entity reference happens to be an identifiers.org URL for a specific CHEBI ID, then that CHEBI ID is used even of there are xrefs to further CHEBI IDs. Second, more generally, any xref that is provided as a RelationshipXref with the "see also" relationship is skipped which often provides links to non-exact matches / higher level terms. Since I didn't notice this before, the `get_specific_chebi_id` was necessary to find the most specific ID based on all the xrefs provided for small molecules.

### DIFF
--- a/indra/sources/biopax/processor.py
+++ b/indra/sources/biopax/processor.py
@@ -630,8 +630,7 @@ class BiopaxProcessor(object):
             # This is the "see also" relation which points to related but
             # not exact xrefs that we can skip here
             if isinstance(xref, bp.RelationshipXref) and \
-                xref.relationship_type and xref.relationship_type.uid == \
-                    'http://identifiers.org/psimi/MI:0361':
+                    xref.relationship_type in psi_mi_see_also:
                 continue
             xrefs[xref_db_ns].add(xref.id)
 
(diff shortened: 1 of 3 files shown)


indra/sources/biopax/processor.py
```diff
@@ -615,13 +615,24 @@ def _get_processed_xrefs(bpe: bp.PhysicalEntity):
...
             if not xref.db:
                 continue
             xref_db_ns = xref_ns_map.get(xref.db.lower())
             if not xref_db_ns:
                 continue
+            # This is the "see also" relation which points to related but
+            # not exact xrefs that we can skip here
+            if isinstance(xref, bp.RelationshipXref) and \
+                xref.relationship_type and xref.relationship_type.uid == \
+                    'http://identifiers.org/psimi/MI:0361':
```
cthoyt, line 634: are they always written this way? this is a goofy identifiers.org URL since it uses both the `namespace/id` syntax in combination with the namespace embedded in LUI. Maybe also test for `xref.relationship_type.uid in {'http://identifiers.org/psimi/MI:0361', 'http://identifiers.org/MI:0361'}`

bgyori: Yes, that's how it appears in the PC12 OWL. It's also weird becuase `psimi` isn't a valid identifiers.org prefix. The correct "old style" URL would actually be http://identifiers.org/mi/MI:0361, and the correct "new style" URL would be the one you added, http://identifiers.org/MI:0361, but neither of those are actually used!

bgyori: I added a more general solution

PR #1260, opened 2021-03-30
Use PyStow for INDRA ontology location

cthoyt:
This PR uses PyStow for automating the placement of the INDRA ontology. It doesn't add any docs like when we did this for ProtMapper or Adeft because it seems like the user shouldn't be interacting with this anyway


cthoyt:
I don't think the build failure is related to this PR. It's giving exit code 137, which I think is a docker code for out of memory

bgyori:
The failure is related to the PR in the sense that the tests make use of the expected ontology location to put in place small mock ontology used for testing.

cthoyt:
Oh, I must have missed that :/ Maybe we can change around the output of the tests to only show information on failures

cthoyt:
@bgyori ahh thanks for that fix.

PR #1261, opened 2021-03-30
Refactor instantiation of agents

cthoyt:
The code for standardizing the db_refs and name showed up in many places, so i refactored it into a classmethod for the Agent class. The only caveat is that the standardize_name_db_refs function is imported inside since it would cause circular dependencies.

Maybe this function needs a different name though, since it has the same signature as `Agent.__init__()` and I don't want to cause confusion

### DIFF
--- a/indra/ontology/standardize.py
+++ b/indra/ontology/standardize.py
@@ -1,5 +1,5 @@
 __all__ = ['standardize_agent_name', 'standardize_db_refs', 'get_standard_name',
-           'standardize_name_db_refs']
+           'standardize_name_db_refs', 'get_standard_agent']
 
 import logging
 from copy import deepcopy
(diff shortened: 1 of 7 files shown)


bgyori:
I think the test failures show that there is still a circular import. Would it make more sense to implement this as a standalone function rather than a class method? For instance, in `indra.ontology.standardize`, we could have
```
def get_standard_agent(name, db_refs, ontology=None, **kwargs):
    ontology = bio_ontology if not ontology else ontology
    standard_name, db_refs = standardize_name_db_refs(db_refs)
    if standard_name:
        name = standard_name
    assert_valid_db_refs(db_refs)
    return Agent(name, db_refs=db_refs, **kwargs)
```

bgyori (approved):
I think this is good, we can wait for the tests to pass again before merging, though the failures are probably unrelated.

PR #1262, opened 2021-03-30
Add DGI Processor

### LINKED ISSUE #1259: Add Drug Gene Interaction Database source
http://www.dgidb.org

cthoyt:
Closes #1259

This PR adds a processor for the DGI database. It automates acquisition of the data and stores via PyStow. It applies a very opinionated curation of the interaction types annotated in the database to decide which INDRA statement type each drug-target interaction should be, with several conflicts to the philosophy of the DrugBank converter (notably, agonist, antagonist, and inverse agonist can't be used to determine the polarity of the relationship, only that it has the same/opposite effect as the native ligand).

It produces 11534 statement, unfortunately throwing away many apolar statements we could keep if it were possible to instantiate the RegulatesActivity class.

This PR also brings over some of the functionality introduced in #1261 for instantiation of agents, so it would be nice to review that one first.

## Tasks

- [x] Implement parser
- [x] Add test data and unit tests (f9d2d6a7f)

## Summary

```
Number skipped: 44872

Statement Type      Count
----------------  -------
Inhibition           8067
Complex              5254
Activation            798
DecreaseAmount          4

Source                         Count
---------------------------  -------
ChemblInteractions              7041
GuideToPharmacology             5440
MyCancerGenome                   786
TALC                             430
MyCancerGenomeClinicalTrial      238
DTC                               95
CancerCommons                     94

Interaction                        Count
-------------------------------  -------
inhibitor                           6929
antagonist                          2328
agonist                             2232
blocker                             1134
positive modulator                   602
modulator                            222
antibody                             222
activator                            195
allosteric modulator                 170
partial agonist                       74
inverse agonist                       24
negative modulator                     5
antisense oligonucleotide              4
inducer                                1
inhibitory allosteric modulator        1
suppressor                             1
```


cthoyt:
@bgyori actually this is blocked by #1261 still - I want to reuse the code for standardizing the agent in this PR

cthoyt:
@bgyori ready for review

indra/sources/dgi/__main__.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+
+"""Command line interface for DGI-DB."""
+
+from collections import Counter
+
+from tabulate import tabulate
```
bgyori, line 7: Is tabulate an indirect dependency through some other dependency of INDRA (maybe pandas)?

indra/sources/dgi/processor.py
```diff
@@ -0,0 +1,243 @@
...
+        if df is None:
+            self.version, df = get_version_df(version=version)
+        else:
+            self.version = version
+        self.df = process_df(df)
+        self.statements = []
+        self.skip_databases = set(['DrugBank'] if skip_databases is None else skip_databases)
+        self.skipped = 0
+
+    def extract_statements(self) -> List[Statement]:
```
bgyori, line 48: There should be a corresponding function in the `api` file along the lines of
```
def process_df(df):
   dp = DGIProcessor(df)
   dp.extract_statements()
   return dp
```
Another endpoint in the `api` could be
```
def process_version(version=None):
    version, df  = get_version_df(version)
    return process_df(df)
````

File: indra/tests/test_sources/__init__.py [hunk 3]
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+
+"""A submodule for organizing tests for sources."""
```
bgyori, line 3: I think this is good though I'm not sure whether we should now move all the existing tests in here.

indra/tests/test_sources/test_dgi.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+
+import pandas as pd
+
+from indra.sources.dgi.api import USECOLS
+from indra.sources.dgi.processor import DGIProcessor
+from indra.statements import Inhibition
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+TEST_FILE = os.path.join(HERE, 'resources', 'dgi_sample_interactions.tsv')
```
bgyori, line 14: I think this test file needs to be committed.

cthoyt: I didn't realize TSV files were ignored by default. Done in 4038d2667

indra/tests/test_sources/__init__.py
cthoyt: For future PRs for sources, this might make organization easier. I wouldn't say we need to move other tests over at the moment though

indra/sources/dgi/processor.py
cthoyt: Done in 367d4ad3a with type annotations + docs

doc/modules/sources/index.rst
```diff
@@ -44,9 +44,16 @@ Standard Molecular Pathway Databases
...
-   drugbank/index
    omnipath/index
 
+Chemical Information Databases
+------------------------------
+.. toctree::
+   :maxdepth: 3
+
+    drugbank/index
+    dgi/index
```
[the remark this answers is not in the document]
cthoyt: Done in 9ce7b060f. CTD does contain gene-disease interactions, but it's more focused on chemistry so I think this is okay. I had thought about that before and was unsure as well.

indra/sources/dgi/__main__.py
cthoyt: No, unfortunately it's not. This is a one-off summary script, so I could also delete it from the PR. Otherwise, if someone wants to run it, I think it would be obvious enough to `pip install tabulate` from the import warning

indra/tests/make_mock_ontology.py
```diff
@@ -42,7 +42,7 @@
     'UP:P15056', 'UP:O60674', 'UP:P0DP23', 'UP:Q13507', 'UP:P36507',
     'DRUGBANK:DB00305', 'HGNC:23077', 'HGNC:17347', 'EGID:27113',
     'EGID:10320', 'EGID:116986', 'EFO:0004542', 'UP:P04377',
-    'CHEBI:CHEBI:136692'
+    'CHEBI:CHEBI:136692', 'EGID:1022', 'HGNC:1778',
```
cthoyt, line 45: @bgyori the test was failing becuase it wasn't able to get `HGNC:1778` when standardizing `EGID:1022`. Will this fix it? Do we need to make a new upload each time we add to this list?

bgyori: You have to run the `indra/tests/make_mock_ontology.py` script to get it up on S3 where the tests pull from.

cthoyt: @bgyori okay done (though the script didn't give me any feedback about the upload being successful)! I saw this existed, but was hesitant to try it without checking. I will re-run the tests now to see if it worked.

PR #1265, opened 2021-04-01
Update interpretation of DrugBank actions

### LINKED ISSUE #1263: Review DrugBank statement polarities
@cthoyt brought up an interesting point in https://github.com/sorgerlab/indra/pull/1262#issue-603695782 about assumptions of polarity for statements derived from agonist and antagonist relations. We should review the extracted statements to see if the current assumptions result in any incorrect polarities.

cthoyt:
This PR closes #1263 by updating some of the interpretations made in DrugBank's action field. 

It's most likely the case that all curated interactions in DrugBank imply binding between an active molecule and the target. For example, some actions like agonist, partial agonist, antagonist, partial antagonist, and inverse agonist do not give enough information to infer activates or inhibits, because they are with respect to the native ligand. Therefore, the best we can do is say that these chemicals bind to their targets (in the absence a more generic INDRA Statement for apolar regulates activity).

This PR also does the following:
- adds several interpretations that were missing 
- reorganizes the order of priority of mapping to statement types
- small style updates in the DrugBank processor code

### Before

| Action     | Count |
|----------|-------:|
| Inhibition | 16,026 |
| Activation | 2,569 |
| DecreaseAmount | 69 |
| IncreaseAmount | 8 |

### After

| Statement Type   |   Count |
|------------------|--------:|
| Inhibition       |    5,012 |
| Activation       |    2,088 |
| Complex          |    1,325 |
| DecreaseAmount   |      73 |
| IncreaseAmount   |      27 |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,3 +122,4 @@ indra/tests/tempfile.html
 .pytest_cache
 .DS_Store
 .noseids
+.venv
--- a/indra/sources/drugbank/api.py
+++ b/indra/sources/drugbank/api.py
@@ -1,6 +1,7 @@
 import logging
 from typing import Optional, Sequence, Union
 from xml.etree import ElementTree
+
 from .processor import DrugbankProcessor
 
 logger = logging.getLogger(__name__)
--- a/indra/sources/drugbank/api.py
+++ b/indra/sources/drugbank/api.py
@@ -1,7 +1,6 @@
 import logging
 from typing import Optional, Sequence, Union
 from xml.etree import ElementTree
-
 from .processor import DrugbankProcessor
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


bgyori (commented):
There are a few questionable cases still, for instance, we now map `agonist` and `antagonist` somewhat conservatively to `Complex`, but `binder`, `binding`, and `antibody` are still mapped to `Inhibition`. Do those interpretations make sense?

cthoyt:
Binder and binding both are neutral, but most antibodies that I know of exert their action by binding to a protein, which either tags it for degradation or inhibits it. I think leaving as inhibits is fine.

bgyori:
It turns out that there is a large number of entries that neither specify an action, nor declare that there is a direct interaction between the drug and the target. Since it's not clear what statement type these can be mapped to if we want to be precise, we skip them now. The statement count drops significantly, with the overall statement stats before this PR:
```
('Inhibition', 16026),
('Activation', 2569),
('DecreaseAmount', 69),
('IncreaseAmount', 8)
```
and after this PR
```
('Complex', 4320),
('Inhibition', 3121),
('Activation', 724),
('DecreaseAmount', 60),
('IncreaseAmount', 28)
```

PR #1266, opened 2021-04-01
Update documentation and type hints

cthoyt:
This PR does not introduce any new code functionality, but rather is a pilot on a few improvements to code and document maintenance. It does the following:

- introduces a MyPy static typing test configuration in `tox.ini` that can be run with `tox -e py`. Right now, it's only running on the one file, `/indra/ontology/ontology_graph.py`, that's used as an example in this PR.
- turns on the intersphinx Sphinx extension for mappings in the documentation (this makes links to other packages' docs when possible, such as PyBEL)
- introduces the `sphinx_autodoc_typehints` Sphinx extension, which takes the type annotations from a function and uses them while generating the docs (so they don't need to be written in free text to get parsed by Napoleon)
- demonstration can be found in in newly added code in `indra.sources.dgi`
- a few random cleanups in the docs

Considerations:
- the `ClassVar`-style type annotation changes the order compared to using the Napoleon `Attributes` header in the class docstring
- I haven't exactly figured out how to get it to play nice with functions that annotate their return types with an explanation (napoleon seems to break when there's no name, and if there's a name, it gets interpreted as the type). Update: after reading the Sphinx guide on Napoleon (both the Google and Numpy variants), I don't think there's a way around this.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -150,7 +150,7 @@ html_theme = 'sphinx_rtd_theme'
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+# html_static_path = ['_static']
 
 # Add any extra paths that contain custom files (such as robots.txt or
 # .htaccess) here, relative to this directory. These files are copied
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -36,6 +36,7 @@ extensions = [
     'sphinx.ext.coverage',
     'sphinx.ext.viewcode',
     'sphinx.ext.napoleon',
+    'sphinx.ext.intersphinx',
     'sphinxarg.ext',
     'sphinx.ext.autosummary',
     'IPython.sphinxext.ipython_directive',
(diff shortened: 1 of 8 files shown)


doc/conf.py
```diff
@@ -320,3 +322,11 @@
 # from our code that uses it.
 jnius_config = sys.modules['jnius_config']
 jnius_config.vm_running = False
+
+# Example configuration for intersphinx: refer to the Python standard library.
+intersphinx_mapping = {
+    'https://docs.python.org/3/': None,
+    'pybel': ('https://pybel.readthedocs.io/en/latest/', None),
+    'boto3': ('https://boto3.amazonaws.com/v1/documentation/api/latest/', None),
+    'pysb': ('https://docs.pysb.org/en/latest/', None),
```
cthoyt, line 331: we can add more entries here for linking to other docs

cthoyt:
@bgyori I removed the updates to the ontology graph and instead point to the recently added DGI source, where I made use of type annotations. This PR is ready for review now

bgyori:
Okay, thanks. I'm not sure including strict linting into Github actions is productive (looking at some of the errors it raises), I can see how this is useful for documentation though.

cthoyt:
@bgyori good point - mypy does some recursive checking and until we can clean the dependent code up to also pass mypy, this is best not to go in GHA for now

bgyori (approved):
Great, thanks!

PR #1268, opened 2021-04-01
Add section on new knowledge sources

bgyori:
This PR adds a small extension to the contribution guidelines to list some extra steps when adding a new knowledge source.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -158,6 +158,19 @@ If a new submodule is added, that submodule needs to be
 - listed in setup.py under packages to make sure it is included in installs
 - referred to in the documentation explicitly to be included
 
+New knowledge sources
+---------------------
+If a new knowledge source is added in the `indra.sources` module, an additional
+set of steps need to be taken in addition to the general instructions for
+adding a new module (see above):
+- Structured information about the new source should be added to the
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -158,6 +158,19 @@ If a new submodule is added, that submodule needs to be
...
+---------------------
+If a new knowledge source is added in the `indra.sources` module, an additional
+set of steps need to be taken in addition to the general instructions for
+adding a new module (see above):
+- Structured information about the new source should be added to the
+`indra/resources/source_info.json` file.
+- The new source's name, link to the documentation, and link to an outside
+reference should be added to the README.md in the appropriate table in
+the Knowledge sources section.
+- A prior estimate of random and systematic belief scores for the new source
```
cthoyt, line 171: could some of these things get stored as class variables inside the processors themselves, then imported / propagated into these files? I can imagine lots of reasons that might not be such a great idea but if it's possible, it would be great!

bgyori: There are sometimes multiple processors per input module, however, in principle, these could be defined as module-level variables. One thing that would be inconvenient about doing this at the code level is that it would require loading every module explicitly (and some of them have weird, complicated non-default dependencies) to aggregate the source info.

PR #1269, opened 2021-04-02
Add single BEL statement processor and integrate with Hypothes.is processor

cthoyt:
This PR adds a function to `indra.sources.bel` that can turn a single BEL string into an INDRA statement. Example usage:

```python
>>> from indra.sources.bel import process_bel_statement
>>> bel_s = 'kin(p(FPLX:MEK)) -> kin(p(FPLX:ERK))'
>>> indra_s = process_bel_stmt(bel_s)
Activation(MEK(kinase), ERK(), kinase)
```

This PR also updates the Hypothes.is processor to read BEL (and gives some string-based defaults for readers). Example usage after associating your hypothesis API key with [this group](https://hypothes.is/groups/Z8RNqokY/cthoyt-bel):

```python
>>> from indra.sources import hypothesis
>>> processor = hypothesis.process_annotations(group='Z8RNqokY', reader='bel')
>>> processor.statements
[Phosphorylation(AKT(), PCGF2(), T, 334)]
```

## Tasks

- [x] wrap pre-existing functionality to quickly generate INDRA statements
- [ ] consider adding calls to Gilda to handle any entities whose namespaces are `TEXT`

### DIFF
--- a/indra/sources/bel/api.py
+++ b/indra/sources/bel/api.py
@@ -161,7 +161,7 @@ def process_text(bel: str, squeeze: bool = False):
     >>> from indra.sources.bel import process_text
     >>> bel_s = 'kin(p(FPLX:MEK)) -> kin(p(FPLX:ERK))'
     >>> indra_s = process_text(bel_s, squeeze=True)
-    IncreaseAmount(MEK(), ERK())
+    Activation(MEK(kinase), ERK(), kinase)
     """
     r = pybel.parse(bel)
     graph = pybel.BELGraph()
--- a/indra/sources/bel/api.py
+++ b/indra/sources/bel/api.py
@@ -9,6 +9,7 @@ import logging
 import requests
 from functools import lru_cache
 
+import pybel.constants as pc
 from pybel.io.sbel import add_sbel_row
 
 from .processor import PybelProcessor
@@ -164,6 +165,12 @@ def process_text(bel: str, squeeze: bool = False):
     Activation(MEK(kinase), ERK(), kinase)
     """
     r = pybel.parse(bel)
+    # make sure activations in the right place
+    for a, b in [(pc.SOURCE, pc.SOURCE_MODIFIER), (pc.TARGET, pc.TARGET_MODIFIER)]:
+        side = r[a]
+        for c in [pc.MODIFIER, pc.EFFECT, pc.FROM_LOC, pc.TO_LOC, pc.LOCATION]:
+            if c in side:
+                r.setdefault(b, {})[c] = side.pop(c)
     graph = pybel.BELGraph()
     add_sbel_row(graph, r)
     bp = process_pybel_graph(graph)
--- a/indra/sources/bel/api.py
+++ b/indra/sources/bel/api.py
@@ -136,7 +136,7 @@ def process_pybel_neighborhood(entity_names, network_type='graph_jsongz_url',
     return bp
 
 
-def process_text(bel: str, squeeze: False = True):
+def process_text(bel: str, squeeze: bool = False):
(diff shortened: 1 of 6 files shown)


indra/sources/bel/processor.py
```diff
@@ -334,12 +334,12 @@ def _get_evidence(self, u_data, v_data, k, edge_data):
         ev_text = edge_data.get(pc.EVIDENCE)
         ev_citation = edge_data.get(pc.CITATION)
         ev_pmid = None
+        ev_ref = None
```
cthoyt, line 337: this change was necessary to deal with BELGraph edges that are programmatically generated, that could potentially be missing a citation

cthoyt: There was actually a comment about this below that I deleted in 08879ef

indra/sources/hypothesis/api.py
```diff
@@ -199,6 +201,18 @@ def process_annotations(group=None, reader=None, grounder=None):
         A HypothesisProcessor object which contains a list of extracted
         INDRA Statements in its statements attribute, and a list of extracted
         grounding curations in its groundings attribute.
+
+    Example
+    -------
+    Process all annotations that have been written in BEL with:
+
+    >>> from indra.sources import hypothesis
+    >>> processor = hypothesis.process_annotations(group='Z8RNqokY', reader='bel')
```
cthoyt, line 210: any idea how to make this doctest work?

indra/sources/bel/api.py
```diff
@@ -133,6 +136,42 @@ def process_pybel_neighborhood(entity_names, network_type='graph_jsongz_url',
...
+        A list of INDRA statments derived from the BEL statement.
+        If squeeze is true and there was only one statement, the
+        unpacked INDRA statement will be returned.
+
+    Examples
+    --------
+    >>> from indra.sources.bel import process_text
+    >>> bel_s = 'kin(p(FPLX:MEK)) -> kin(p(FPLX:ERK))'
+    >>> indra_s = process_text(bel_s, squeeze=True)
+    IncreaseAmount(MEK(), ERK())
```
bgyori, line 164: This part here is odd since we'd expect an Activation statement. I found that inside `process_text`, `pybel.parse` produces the expected intermediate result, i.e., 
```
{'source': {'modifier': 'Activity', 'function': 'Protein', 'concept': {'namespace': 'FPLX', 'name': 'MEK'},
 'effect': {'namespace': 'go', 'name': 'kinase activity', 'identifier': '0016301'}},
 'relation': 'directlyIncreases',
 'target': {'modifier': 'Activity', 'function': 'Protein', 'concept': {'namespace': 'FPLX', 'name': 'ERK'},
 'effect': {'namespace': 'go', 'name': 'kinase activity', 'identifier': '0016301'}}}
```
but then in `add_sbel_row`, the `effect` terms are ignored, so this is something that probably needs to be fixed in PyBEL.

bgyori (commented):
Great, this will be a nice addition to the hypothes.is workflow! Probably worth looking into the issue I commented on before merging.

indra/sources/bel/api.py
cthoyt: I was a bit hasty - the sbel JSON format isn't exactly the same as what comes out of `pybel.parse`. I added a bit of code to massage the JSON in 6d3f20497. Unfortunately this functionality isn't directly available in PyBEL - I think it would need a big restructure to make this elegant so I'd opt for the simple/fast solution

indra/sources/hypothesis/api.py
cthoyt: skipped it.

bgyori (approved):
Awesome!

PR #1270, opened 2021-04-02
Add Chemical Roles Graph

### LINKED ISSUE #1253: Add Chemical Roles Graph as new source
The Chemical Roles Graph (CrOG) has manually curated relationships between ChEBI roles and the actual agents that they module (information which is not present in ChEBI itself nor any other data source)

Data at: https://github.com/chemical-roles/chemical-roles
Manuscript with further description at: https://chemical-roles.github.io/manuscript/

cthoyt:
Closes #1253.

This PR has no `api.py` information since the Chemical Roles Graph repo is already generating INDRA JSON.

 < look at how cute this frog is

## Summary

| Statement Type   |   Count |
|------------------|---------|
| Inhibition       |     685 |
| Activation       |      83 |
| DecreaseAmount   |       3 |

### DIFF
--- a/indra/resources/default_belief_probs.json
+++ b/indra/resources/default_belief_probs.json
@@ -31,7 +31,8 @@
     "omnipath": 0.01,
     "biofactoid": 0.01,
     "minerva": 0.01,
-    "dgi": 0.01
+    "dgi": 0.01,
+    "crog": 0.01
   },
   "rand": {
     "eidos": 0.3,
@@ -65,6 +66,7 @@
     "omnipath": 0.1,
     "biofactoid": 0.1,
     "minerva": 0.1,
-    "dgi": 0.1
+    "dgi": 0.1,
+    "crog": 0.01
   }
 }
--- a/indra/sources/crog/processor.py
+++ b/indra/sources/crog/processor.py
@@ -15,7 +15,7 @@ CROG_URL = 'https://raw.githubusercontent.com/chemical-roles/' \
 
 
 class CrogProcessor(RemoteProcessor):
-    """A processor for INDRA JSON file to be retrieved by URL."""
+    """A processor for the Chemical Roles Graph."""
 
     def __init__(self):
         super().__init__(url=CROG_URL)
--- a/indra/sources/crog/processor.py
+++ b/indra/sources/crog/processor.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 
-"""Processor for the `Chemical Roles Graph (CRoG) <https://github.com/chemical-roles/chemical-roles>`_."""
+"""Processor for the `Chemical Roles Graph (CRoG)
+<https://github.com/chemical-roles/chemical-roles>`_.
+"""
 
 from ..utils import RemoteProcessor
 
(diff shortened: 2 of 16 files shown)


indra/resources/default_belief_probs.json
```diff
@@ -65,6 +66,7 @@
     "omnipath": 0.1,
     "biofactoid": 0.1,
     "minerva": 0.1,
-    "dgi": 0.1
+    "dgi": 0.1,
+    "crog": 0.01
```
cthoyt, line 70: maybe i'm tooting my own horn, but I'm very confident in the curation I did here

indra/sources/crog/processor.py
```diff
@@ -0,0 +1,21 @@
...
+
+__all__ = [
+    'CrogProcessor',
+]
+
+CROG_URL = 'https://raw.githubusercontent.com/chemical-roles/' \
+           'chemical-roles/master/docs/_data/crog.indra.json'
+
+
+class CrogProcessor(RemoteProcessor):
```
bgyori, line 17: I would actually still add an `api.py` that has an edpoint like `process_from_web`. That would fit with the behavior of other sources.

cthoyt: okay will do, I will make it possible to give an optional URL to override the internal one (perhaps for testing purposes)

indra/sources/crog/processor.py
```diff
@@ -0,0 +1,21 @@
...
+    """A processor for the Chemical Roles Graph."""
+
+    def __init__(self):
+        super().__init__(url=CROG_URL)
```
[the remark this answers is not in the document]
cthoyt: True, but my opinion is that Identifiers.org is wrong. The ec-code:1.5.1 is part of the hierarchy and gives us a lot of really useful information. The bioregistry validation overrides the identifiers.org pattern (see https://github.com/cthoyt/bioregistry/blob/1270930a4b199801cc2e0ca53de4939ae7c4f724/src/bioregistry/data/bioregistry.json#L6050).

Maybe we need to update the PybelProcessor since it's possible for PyBEL citations to have DOI prefixes instead of PMIDs. I'll look into that

indra/sources/crog/processor.py
cthoyt: Addressed in 4a247d1da

indra/sources/crog/processor.py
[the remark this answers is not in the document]
cthoyt: I made the PyBEL processor more generally able to handle different kinds of citations in d8c4d0068

indra/tests/test_pybel_api.py
```diff
@@ -409,6 +409,27 @@ def test_phosphorylation_one_site_with_evidence():
     assert ev.epistemics == {'direct': True, 'section_type': 'abstract'}
 
 
+def test_doi_evidence():
```
bgyori, line 412: This test errors on Github actions, is this due to a PyBEL version issue (like this code assumes a newer version than what is installed there)?
```
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pybel/struct/graph.py", line 918, in <dictcomp>
    for key, values in annotations_dict.items()
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pybel/struct/graph.py", line 954, in _clean_value
    raise NotImplementedError(f'where is key {key}?')
NotImplementedError: where is key TextLocation?
```

cthoyt: Fixed - this happened because there was missing metadata when I created the BELGraph class

bgyori (approved):

PR #1272, opened 2021-04-08
Implement a pretty print function for stmt lists.

pagreene:
This PR adds a function to `indra.statements.io` that prints Statements and their evidence to the terminal prettily. This can be extremely useful when browsing Statements in the command line. In future this could in principal be integrated with all the ordering and sorting tools developed for the HTML Assembler. There are also numerous ways in which the display could be tweaked.

Some key features:
- The line width is adaptive to the current terminal width (should work on most OSs)
- The evidence text is wrapped neatly
- The metadata for each evidence is displayed along side the text using `tabular`.

Some good features that could be added:
- The ability to select what metadata is shown
- The ability to use the English form of the statements in the headings
- The ability to sort and group Statements a la the HTML Assembler (although I think there is a compelling argument that perhaps keeping it simpler is better).

### DIFF
--- a/indra/statements/io.py
+++ b/indra/statements/io.py
@@ -4,7 +4,7 @@ __all__ = ['stmts_from_json', 'stmts_from_json_file', 'stmts_to_json',
 
 import json
 import logging
-from typing import List
+from typing import List, Optional
 
 from indra.statements.statements import Statement, Unresolved
 
@@ -221,8 +221,9 @@ def draw_stmt_graph(stmts):
     plt.show()
 
 
-def pretty_print_stmts(stmt_list: List[Statement], stmt_limit: int = None,
-                       ev_limit: int = 5) -> None:
+def pretty_print_stmts(stmt_list: List[Statement],
+                       stmt_limit: Optional[int] = None,
+                       ev_limit: Optional[int] = 5) -> None:
     """Print a list of statements to the commandline in a neatly formatted way.
 
     Parameters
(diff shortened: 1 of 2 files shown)


indra/statements/io.py
```diff
@@ -222,6 +221,70 @@ def draw_stmt_graph(stmts):
     plt.show()
 
 
+def pretty_print_stmts(stmt_list: List[Statement], stmt_limit: int = None,
```
cthoyt, line 224:
```suggestion
def pretty_print_stmts(stmt_list: List[Statement], stmt_limit: Optional[int] = None,
```

indra/statements/io.py
```diff
@@ -235,32 +246,38 @@ def pretty_print_stmts(stmt_list: List[Statement],
     ev_limit :
         The maximum number of Evidence to print for each Statement.
         (Default is 5)
+    width :
+        Manually set the width of the table. Default is None, in which case
+        the function will try to match the current terminal width, and if that
+        is not possible it will use `pretty_print_default_width`, which can be
```
cthoyt, line 252: you can trigger sphinx to document it if you do 
```suggestion
        is not possible it will use :data:`pretty_print_default_width`, which can be
```

indra/statements/io.py
```diff
@@ -235,32 +246,38 @@ def pretty_print_stmts(stmt_list: List[Statement],
...
+        adjusted using the `set_pretty_print_default_width` function.
```
cthoyt, line 253: Also for sphinx linking

```suggestion
        adjusted using the :func:`set_pretty_print_default_width` function.
```

johnbachman (approved):

PR #1273, opened 2021-04-12
Filter edge

dianakolusheva:
This PR introduces a functionality to filter a graph to a subgraph based on any provided criteria to filter edges. This functionality is propagated to ModelChecker (currently for Signed and Unsiged graph model checkers) `get_graph` method to allow filtering the graph before searching for paths. In addition, this PR extends IndraNet assembler to allow adding any extra edge data and adds and extends relevant tests.

### DIFF
--- a/indra/explanation/pathfinding/util.py
+++ b/indra/explanation/pathfinding/util.py
@@ -1,6 +1,6 @@
 __all__ = ['path_sign_to_signed_nodes', 'signed_nodes_to_signed_edge',
            'get_sorted_neighbors', 'get_subgraph', 'register_edge_filter',
-           'filter_to_internal_edges']
+           'filter_to_internal_edges', 'edge_filter_functions']
 import logging
 import networkx as nx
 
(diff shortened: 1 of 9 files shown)


bgyori:
This looks good! But I want to understand better why `edge_filter_func_name` is implemented as a string argument rather than as a function argument - my understanding is that it has to do with the configurability of the filter as JSON in EMMAA. I wonder if it would be cleaner to perform the string-based function loading in EMMAA (or some other external code), and use a function-based interface in all the code here.

indra/explanation/pathfinding/util.py
```diff
@@ -113,3 +115,47 @@ def get_sorted_neighbors(G, node, reverse, force_edges=None):
...
+
+
+def get_subgraph(g, filter_func_name):
+    """Get a subgraph of original graph filtered by a provided function."""
+    # Updating global variable here to handle NetworkX implementation of
+    # closures used in subgraph_view
+    global G
+    G = g
+    filter_edge = edge_filter_functions.get(filter_func_name)
+    view = nx.subgraph_view(g, filter_edge=filter_edge)
```
cthoyt, line 143: ben's comment was it would be better to do something like:

```suggestion
    view = nx.subgraph_view(g, filter_edge=functools.partial(filter_edge, g))
```

And change the interface of `filter_edge` to also have an argument for the graph in the first position.

indra/tests/test_indranet_assembler.py
```diff
@@ -75,6 +75,14 @@ def test_make_df():
         'initial_sign', 'residue', 'position'}
     assert df.residue.isna().sum() == 9  # Check that all but one row is NaN
     assert df.position.isna().sum() == 9  # Check that all but one row is NaN
+    # Extra column
+    df2 = ia.make_df(extra_columns=[
+        ('stmt_type_upper',
+         lambda stmt: type(stmt).__name__.upper())])
```
cthoyt, line 81: `stmt.__class__.__name__` is probably safer

dianakolusheva:
I moved the internal filter and registration into EMMAA and used functools to handle graph variable. Thanks for your help @bgyori  and @cthoyt!

bgyori (approved):

PR #1278, opened 2021-04-20
Move World Modelers-specific modules into the INDRA World package

bgyori:
This PR moves a number of modules or parts of modules into the INDRA World package (https://github.com/indralab/indra_world), see the corresponding PR at https://github.com/indralab/indra_world/pull/20. The changes can be summarized as follows:
- Input sources: 3 input sources, `sofia`, `hume` and `cwms` are moved out, while `eidos` remains since it has general availability and applicability.
- Core modules
  - Belief engine: World Modelers-specific belief models (e.g., what is in the `wm_scorer` module) are moved out
  - Preassembler: Word Modelers-specific custom matching/refinements functions from the `custom_preassembly` file are moved out
  - Ontology: The WorldOntology and associated loading code are moved out
  - Assembly pipeline: World Modelers-specific functions from `indra.tools.assemble_corpus` are moved out
- REST API: some reading and assembly functionalities are removed

Notably, the statement classes aren't changed, though, in principle, at least some of those could be moved too since there's relatively little interdependency at that level.

One tricky part is what to do about testing, since we implemented lots of tests of core modules using World Modelers-specific constructs. If we moved these, then useful testing for core code would become indirect. So for the time being, I left a lot of these tests as is, and included cloning `indra_world` in the test setup.


bgyori:
To address the inconvenience of `indra_world` being imported in tests, I added handling to skip those tests if `indra_world` is not available locally. It is included in the Github Actions setup though so these tests will run regularly.

PR #1280, opened 2021-04-22
Pathfinding edge filter

kkaris:
This PR patches `bfs_search` with an edge filter that is applied during the neighbor search. The filter is implemented as a function passed in to `bfs_search` with the new argument `edge_filter` that must take the graph, `g`, and the nodes `u`, `v` of the edge between `u` and `v`.

Other updates in this PR:
- Type hints are added for `bfs_search` and `get_sorted_neighbors` in `indra.explanation.pathfinding.util`
- A new section of `test_bfs` in `indra.tests.test_pathfinding` was added that tests the edge filter


indra/explanation/pathfinding/pathfinding.py
```diff
@@ -212,10 +219,23 @@ def shortest_path_func(G, source, target, weight, ignore_nodes,
...
+               node_blacklist: Optional[Set[Node]] = None,
+               terminal_ns: Optional[List[str]] = None,
+               sign: Optional[int] = None,
+               max_memory: int = int(2**29),
+               hashes: Optional[List[int]] = None,
+               allow_edge: Optional[Callable[[Node, Node], bool]] = None,
+               strict_mesh_id_filtering: bool = False,
+               edge_filter: Optional[Callable[[nx.DiGraph, Node, Node],
+                                              bool]] = None,
+               **kwargs) -> Generator[Tuple[Node], SendType, None]:
```
cthoyt, line 238: Is this really implemented as a generator (that both accepts and sends data from `yield`?) I think you probably want `Iterable` here

indra/explanation/pathfinding/util.py
```diff
@@ -76,45 +83,59 @@ def signed_nodes_to_signed_edge(source, target):
...
+    Returns
+    -------
+    List[Node]
+        A list of nodes representing the filtered and sorted neighbors
     """
+    if force_edges:
+        neigh_edges = G.in_edges if reverse else G.out_edges
+        ix = 0 if reverse else 1
+        neighbors = (e[ix] for e in
+                     set(neigh_edges(node)).intersection(set(force_edges)))
```
cthoyt, line 122: you might be able to get away with `set.intersection(neigh_edges(node), force_edges`

indra/explanation/pathfinding/util.py
```diff
@@ -76,45 +83,59 @@ def signed_nodes_to_signed_edge(source, target):
...
-                node)).intersection(set(force_edges)))
-        else:
-            neighbors = G.successors(node)
-        return sorted(
-            neighbors,
-            key=lambda n:
-                G.edges[(node, n)].get('belief', 0),
-            reverse=True)
+        return sorted(neighbors,
+                      key=lambda n: G.edges[(node, n)].get('belief', 0),
```
cthoyt, line 137:
```suggestion
                      key=lambda n: G.edges[node, n].get('belief', 0),
```

indra/explanation/pathfinding/pathfinding.py
```diff
@@ -212,10 +219,23 @@ def shortest_path_func(G, source, target, weight, ignore_nodes,
...
+               max_per_node: int = 5,
+               node_filter: Optional[List[str]] = None,
+               node_blacklist: Optional[Set[Node]] = None,
+               terminal_ns: Optional[List[str]] = None,
+               sign: Optional[int] = None,
+               max_memory: int = int(2**29),
+               hashes: Optional[List[int]] = None,
+               allow_edge: Optional[Callable[[Node, Node], bool]] = None,
+               strict_mesh_id_filtering: bool = False,
+               edge_filter: Optional[Callable[[nx.DiGraph, Node, Node],
```
cthoyt, line 236: saving the type for edge filters is probably generally reusable

indra/explanation/pathfinding/util.py
```diff
@@ -76,45 +83,59 @@ def signed_nodes_to_signed_edge(source, target):
...
+) -> List[Node]:
+    """Filter and sort neighbors of a node in descending order by belief
 
     Parameters
     ----------
     G : nx.DiGraph
         A networkx DiGraph
-    node : str|int
-        A valid networkx node name
+    node : Node
```
cthoyt, line 99: You can just skip all of these type annotations in text if they're done as type annotations in the definition (everything should work as expected except for the Returns, that needs to be duplicated for napoleon style docs)

kkaris:
There is more cleanup that can be done in `bfs_search`:
- The context search parameters are unnecessarily complicated: there are three arguments (`strict_mesh_id_filtering`, `hashes`, `allow_edge`) that amounts to one variable, `allowed_edges`, being checked, so we could probably reduce those three to one.
- We could think about renaming the function to `breadth_first_search` as **search** in `bfs_search` is redundant.
- While being at it on implementing filters, I was thinking a node filter function similar to `edge_filter` could be implemented.

Considering this PR is/was meant to specifically fix one thing, adding on the above listed updates is probably out of scope.

indra/explanation/pathfinding/pathfinding.py
```diff
@@ -224,44 +244,60 @@ def bfs_search(g, source_node, reverse=False, depth_limit=2, path_limit=None,
...
         If true, exclude all edges not relevant to provided hashes
+    edge_filter : Optional[Callable[[nx.DiGraph, Node, Node], bool]]
+        If provided, must be a function that takes three arguments: a graph
+        g, and the nodes u, v of the edge between u and v. The function must
+        return a boolean. The function must return True if the edge is
+        allowed, otherwise False. Example of function that only allows edges
+        that have an edge belief above 0.75:
+
+        >>> def filter_example(g, u, v):
+        ...    return g.edges[(u, v)].get('belief', 0) > 0.75
```
cthoyt, line 288:
```suggestion
        ...    return g.edges[u, v].get('belief', 0) > 0.75
```

indra/explanation/pathfinding/util.py
kkaris: That is one of the things I tagged you for, I wasn't sure exactly how it was working :) Will fix

indra/tests/test_pathfinding.py
```diff
@@ -181,6 +181,23 @@ def test_bfs():
...
+    # ('A2', 'B1'), ('A3', 'B2'), ('A4', 'B2'), ('B2', 'C1'), ('B3', 'C1')
+    def _filter_func(g, u, v):
+        return g.edges[(u, v)]['belief'] > 0.75
+
+    expected_paths = {('D1', 'C1'), ('D1', 'C1', 'B1'),
+                      ('D1', 'C1', 'B1', 'A1'),
+                      ('D1', 'C1', 'B1', 'A1', 'Z1')}
+    gen = bfs_search(g=dg, source_node='D1', depth_limit=5, reverse=True,
+                     edge_filter=_filter_func)
+    paths = [p for p in gen]
```
cthoyt, line 195:
```suggestion
    paths = list(gen)
```

indra/tests/test_pathfinding.py
```diff
@@ -181,6 +181,23 @@ def test_bfs():
         error = True
     assert error
 
+    # Test edge filter
+    # With belief cutoff at 0.75, we should remove:
+    # ('A2', 'B1'), ('A3', 'B2'), ('A4', 'B2'), ('B2', 'C1'), ('B3', 'C1')
+    def _filter_func(g, u, v):
+        return g.edges[(u, v)]['belief'] > 0.75
```
cthoyt, line 188:
```suggestion
        return g.edges[u, v]['belief'] > 0.75
```

indra/explanation/pathfinding/pathfinding.py
kkaris: Yes, it receives send values, e.g. here: https://github.com/sorgerlab/indra/blob/master/indra/explanation/pathfinding/pathfinding.py#L336

cthoyt: Amazing!!!! I have literally never seen this feature used before by anyone except David Beazley. Super cool.

indra/tests/test_pathfinding.py
kkaris: Wait, I always use `G.edges[(u, v)]` and I see now the documentation does it your way: https://networkx.org/documentation/stable/reference/introduction.html#graph-reporting

I'm guessing they provide both ways since an edge is a tuple, e.g. `[e for e in g.edges]`

kkaris:
RE: @dianakolusheva 
> Do you think it could make sense to use a subgraph (`indra.explanation.pathfinding.util.get_subgraph`) to filter the graph using filter edge function in the beginning of `bfs_search` instead of filtering it every time the `get_sorted_neighbors` is called?

I should probably check timing between filtering the whole graph and filtering on the go, and use the graphs used in production.

indra/tests/test_pathfinding.py
cthoyt: Yes, it will also work if you do:
```python
e = (u,v)
g.edges[e]
```

This is not a specific feature of networkx, but more generally how the square bracket syntax functions.

indra/explanation/pathfinding/pathfinding.py
kkaris: It is modeled after the way we do it on `shortest_simple_paths`

indra/explanation/pathfinding/pathfinding.py
kkaris: We could save it in `pathfinding.util` perhaps

dianakolusheva:
> RE: @dianakolusheva
> 
> > Do you think it could make sense to use a subgraph (`indra.explanation.pathfinding.util.get_subgraph`) to filter the graph using filter edge function in the beginning of `bfs_search` instead of filtering it every time the `get_sorted_neighbors` is called?
> 
> I should probably check timing between filtering the whole graph and filtering on the go, and use the graphs used in production.

I actually think they are still filtered on the go but it's done internally in networkx

kkaris:
> > RE: @dianakolusheva
> > > Do you think it could make sense to use a subgraph (`indra.explanation.pathfinding.util.get_subgraph`) to filter the graph using filter edge function in the beginning of `bfs_search` instead of filtering it every time the `get_sorted_neighbors` is called?
> > 
> > 
> > I should probably check timing between filtering the whole graph and filtering on the go, and use the graphs used in production.
> 
> I actually think they are still filtered on the go but it's done internally in networkx

That's cool, I will definitely look into that then!

kkaris:
There are two tests failing:
- `indra.tests.test_uniprot_client.test_get_family_members` failed with `ConnectionResetError: [Errno 104] Connection reset by peer`. This error occurred only for Py3.6, Py3.9 seems to have no problem. When I ran it locally just now it worked as well.
- The other test failing was the code I wrote as an example in the docstring for `bfs_search`:
```
File "/home/runner/work/indra/indra/indra/explanation/pathfinding/pathfinding.py", line 282, in indra.explanation.pathfinding.pathfinding.bfs_search
Failed example:
    path_generator = bfs_search(edge_filter=filter_example, ...)
```
which is here:
https://github.com/kkaris/indra/blob/pathfinding-edge-filter/indra/explanation/pathfinding/pathfinding.py#L273-L282
(didn't know it was possible to test code examples in documentation!)

bgyori:
The UniProt error is probably just a network issue, the example code, when formatted this way is interpreted as a doctest and executed as part of the test suite - it can be quite useful but if you don't want that to happen, it might be better to change the format for it not be to interpreted as a doctest.

kkaris:
I restarted the tests and got the same connection error again, `ConnectionResetError: [Errno 104] Connection reset by peer`, for `test_uniprot_client.test_get_family_members` this time. It looks like that's the only test failing.

indra/explanation/pathfinding/util.py
```diff
@@ -118,19 +118,25 @@ def get_sorted_neighbors(
...
     """
-    if force_edges:
+    # Check for edge filtering
+    if force_edges or edge_filter:
         neigh_edges = G.in_edges if reverse else G.out_edges
         ix = 0 if reverse else 1
-        neighbors = (e[ix] for e in
-                     set(neigh_edges(node)).intersection(set(force_edges)))
+        if force_edges:
+            edges = set(neigh_edges(node)).intersection(set(force_edges))
```
cthoyt, line 126: you can save one line of code if you do 
```python
edges = set(neigh_edges(node)
if force_edges:
    edges.intersection_update(force_edges)
```

bgyori (approved):

PR #1287, opened 2021-04-28
Updates to CRoG statements

### LINKED ISSUE #1286: Add naming info for CRoG targets
I noticed that some of the statements coming from `indra.sources.crog` use EC codes as names for targets. It would be better to either get names during processing using web service calls or integrate a naming resource that covers these to get better identifiable display names (for instance, for 3.6.1.3 it would be adenosinetriphosphatase). For this particular example, it's also worth noting that 3.6.1.3 has been deleted (https://www.ebi.ac.uk/intenz/query?cmd=SearchEC&ec=3.6.1.3) and split into multiple new entries so the source data could be updated as well. Finally, it would be better to use `crog` as the `source_api` in Evidences to facilitate appropriate assembly.

```
Inhibition(EC 3.6.1.3 (adenosinetriphosphatase) inhibitor(), 3.6.1.3())
[Evidence(source_api='bel',
          source_id='2e6c5ac2d6b7e8844fd047dc2abd9a25',
          annotations={
                       "bel": "a(chebi:\"CHEBI:78928\" ! \"EC 3.6.1.3 (adenosinetriphosphatase) inhibitor\") directlyDecreases act(p(eccode:3.6.1.3 ! 3.6.1.3), ma(go:0003674 ! \"molecular function\"))",
                       "citation_ref": "doi: 10.26434/chemrxiv.12591221"
                      },
          epistemics={
                      "direct": true
                     }
          )
 ]
```

bgyori:
This PR sets the `source_api` to `crog` and adds `EC` prefixes to target Agents that don't have a proper name beyond an EC code. Fixes #1286.

### DIFF
--- a/indra/sources/crog/processor.py
+++ b/indra/sources/crog/processor.py
@@ -39,4 +39,4 @@ class CrogProcessor(RemoteProcessor):
             # have the EC prefix in their name
             for agent in stmt.real_agent_list():
                 if agent.name == agent.db_refs.get('ECCODE'):
-                    agent.name = 'EC%s' % agent.name
+                    agent.name = 'EC %s' % agent.name
--- a/indra/sources/crog/processor.py
+++ b/indra/sources/crog/processor.py
@@ -29,7 +29,8 @@ class CrogProcessor(RemoteProcessor):
     def __init__(self, url: Optional[str] = None):
         super().__init__(url=url or CROG_URL)
 
-        new_stmts = []
+    def extract_statements(self):
+        super().extract_statements()
         for stmt in self.statements:
             # We remap the source API to crog to align with the belief model
             for ev in stmt.evidence:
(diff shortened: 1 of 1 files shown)


indra/sources/crog/processor.py
```diff
@@ -28,3 +28,15 @@ class CrogProcessor(RemoteProcessor):
...
+        super().extract_statements()
+        for stmt in self.statements:
+            # We remap the source API to crog to align with the belief model
+            for ev in stmt.evidence:
+                ev.source_api = 'crog'
+            # We also change the name of targets whose names are ECCODEs to
+            # have the EC prefix in their name
+            for agent in stmt.real_agent_list():
+                if agent.name == agent.db_refs.get('ECCODE'):
+                    agent.name = 'EC%s' % agent.name
```
cthoyt, line 42: Might be easier to read with a space

```suggestion
                    agent.name = 'EC %s' % agent.name
```

bgyori: Sure, we can do that

PR #1289, opened 2021-04-29
Add functions to assemble corpus

dianakolusheva:
This PR adds several functions to assemble_corpus in the format required to work with AssemblyPipeline.


bgyori:
Thanks, looks good! Could you add some tests to cover these functions?

bgyori (approved):

PR #1290, opened 2021-05-07
Belief prediction using sklearn models

johnbachman:
This PR adds a set of classes to allow the training and deployment of sklearn-based classifiers for statement belief prediction. It contains a base class `indra.belief.skl.SklearnScorer` which combines the functionality of the `BeliefScorer` interface (i.e., via the function `score_statements`) as well as basic features of an sklearn model, by composition (via the functions `fit`, `predict`, `predict_proba`, and `predict_log_proba`).

The role of this class is to implement the preprocessing of statement properties into a feature matrix in a standard way, so that a classifier trained on one corpus of statement data will still work when used on another corpus. Implementing subclasses must implement at least one of the methods for building the feature matrix from statement data, `stmts_to_matrix` or `df_to_matrix`. 

As an implementation, the PR includes `CountsScorer`, a subclass of `SklearnScorer` that can be used to build classifiers that can use combinations of the following statement properties for belief prediction:

* evidence source counts
* statement type
* number of statement agents/members
* number of unique supporting PMIDs.

Further details in the docs. The module was named `indra.belief.skl` rather than `indra.belief.sklearn` to prevent clashing with the `sklearn` module itself.

Some changes were made to the internal implementation of the belief engine API to work with matrix-based classifiers. In particular, the BeliefScorer method `score_statement` was replaced with `score_statements`, which is now expected to process a list of Statements in batch and return a corresponding list of beliefs. The BeliefEngine methods `set_prior_probs` and `set_hierarchy_probs` were adapted to call this method in the scorer for setting statement beliefs.

One thing that also came up in refactoring was the implicit assumption that the list of statements given to the BeliefEngine included *all* statements used in preassembly, including both top-level and non-top-level statements. To adapt the BeliefEngine to correctly set hierarchical beliefs for any set of statements based on the statements in their `supports` (more specific statement) attributes, the construction of the refinements graph was changed to take the original set of statements rather than a set of hashes including both original and related statements.  However, the structure of the graph itself is the same, so any code that extends or otherwise uses the graph should be unaffected. Incidentally, this change means that, as would presumably be expected, only the statements given to the BeliefEngine for scoring will have their belief set, not any supports/supported_by statements pointed to by those statements.

There are two desirable features that are not included in this initial implementation. The first is the handling of belief prediction of statements that include evidences from sources not explicitly included in the `CountsScorer.source_list`.  Currently such "out of sample" evidences are ignored, however, a more robust approach would be to fall back on a prior model like that of the `SimpleScorer` to integrate this information.

A more straightforward extension is the ability to train a `CountsScorer` on statement data with evidence counts from refiner (supports) statements. Currently the process of collecting supporting evidence is built in to the BeliefEngine functions downstream of `set_hierarchy_probs`, external to the `CountsScorer`. If the process of collecting `extra_evidence` information was refactored out then it could be collected and e.g. passed to the `CountsScorer.fit` method as an additional argument, as can currently be done for the `predict_*` methods. Note that in practice this would assume that correctness curations were made while accounting for additional evidence in refiner statements which has not been the case until now.


bgyori (commented):
Cool! I think we could add a heading for the new module in the documentation, and also deal with the new dependency. I think `scikit-learn` could be added as an extra dependency in setup.py (interestingly, it's installed in Github actions due to being an indirect dependency via some other package but it would be better to make it explicit so the tests don't error) and would have to be added in doc/requirements.txt as well.

steppi:
@bgyori, scikit-learn is currently being installed as a dependency of adeft.

bgyori:
Would it make sense to add back `score_statement`, at least as a trivial pass-through to `score_statements` to keep the old API? Some tests actually error due to this change and one approach to fixing that would be to just enable that method again.

johnbachman:
Yes, I noticed that as well and added the implementation along with a test.

johnbachman:
As for the docs, I split the docs for `belief.__init__.py` and `skl.py` into two separate files under the heading "Belief prediction".

johnbachman:
Another thing I just realized is that there is a third pickle of 100 test sttaements that is missing from version control but it's way bigger than I expected (100mb!). I'm going to resample/repickle and update.

cthoyt:
@johnbachman you can always gzip your pickles

johnbachman:
> @johnbachman you can always gzip your pickles

It's OK, it turned out that some of the stmts in the sample had supports or supported_by statements with large amounts of evidence. I tweaked the sample to exclude some of these.

indra/tests/test_belief_sklearn.py
```diff
@@ -0,0 +1,345 @@
...
+    with open(test_stmt_path, 'wb') as f:
+        pickle.dump((stmt_sample, y_arr), f)
+    return stmt_sample
+
+
+def test_counts_wrapper():
+    """Instantiate counts wrapper and make stmt matrix"""
+    lr = LogisticRegression()
+    source_list = ['reach', 'sparser']
+    cw = CountsScorer(lr, source_list)
```
cthoyt, line 71: perhaps make `lr` optional in CountsScorer and internally instantiate a LogisticRegression by default

johnbachman: Since the best approach to modeling/fitting statement data is still in research mode, I think it's probably better to keep this explicit at least for now

PR #1293, opened 2021-05-17
Add more functions to handle namespaces embedded in IDs

bgyori:
This PR adds two more functions to help add namespaces to IDs when needed.


indra/databases/identifiers.py
```diff
@@ -278,11 +277,59 @@ def parse_identifiers_url(url):
...
+    ----------
+    db_ns :
+        The namespace associated with the identifier.
+    db_id :
+        The original identifier.
+
+    Returns
+    -------
+    :
+        The identifier with namespace embedded if needed.
```
bgyori, line 320: @cthoyt I found that this way of annotating the return type renders correctly and doesn't require naming a return variable or duplicating the already specified return type.

PR #1297, opened 2021-05-25
Stylesheet assembler

kkaris:
This PR reorganizes how the styling of the html templates in indra are generated. The CSS rules for each of the templates are moved to jinja macros. The CSS rules that govern the color styling for the source count badges are put in their own macro for separate generation.

A new set of assemblers are added in a new submodule: `indra.assemblers.html.css_assembler`. These assemblers can be used to generate standalone CSS files.

Tests for the CSS rule generation has been added in `indra.tests.test_css_assembler`.


bgyori (commented):
@kkaris I have an initial question without having looked at the code in detail: given that we use "assembler" to refer to modules that take Statements and produce models, maybe we could choose a different nomenclature for naming things associated with this feature, for instance "CSS generator". What do you think?

kkaris:
> @kkaris I have an initial question without having looked at the code in detail: given that we use "assembler" to refer to modules that take Statements and produce models, maybe we could choose a different nomenclature for naming things associated with this feature, for instance "CSS generator". What do you think?

That sounds good! I'll change it

indra/assemblers/html/css_generator.py
```diff
@@ -0,0 +1,89 @@
...
+                           join(HERE, 'templates', 'macros')])
+env = Environment(loader=loader)
+
+stylesheet_template_file = 'standalone_stylesheet_template.css'
+
+
+class CSSGenerator:
+    """Base class to create a stylesheet from a jinja2 template"""
+    template_file = stylesheet_template_file
+
```
cthoyt, line 30: You can annotate the instance variables inside the class body - this lets other code introspect by asking for `CSSGenerator.__annotations__`:

```suggestion

template: Template
model: Optional[str]
template_kwargs: Dict

```

cthoyt: Then you don't have to do the annotations inside the executed code

kkaris:
This PR is now superseded by #1300, so I'm closing it.

PR #1300, opened 2021-06-09
Stylesheet generator

kkaris:
This PR is focused on making it possible to generate a standalone CSS file for the styling of the source badges in order to display them outside the scope of Jinja assembled html documents.

Main updates:
- A function that regenerates colors in a specific order and adds them to `source_info.json` is added
- The ground truth of the _default coloring_ now lives in `source_info.json`: `DEFAULT_SOURCE_COLORS` is now generated from loading the colors from it instead of being regenerated on each import
- Two new color schemes have been added: `hues` and `qualitative`, with the qualitative scheme containing 29 colors
- Tests were added to check the new functions added
- A bug that occurred when rendering templates with `simple=False` with specific sources provided was fixed

This PR replaces #1297.


indra/assemblers/html/assembler.py
```diff
@@ -23,25 +26,78 @@
...
 env = Environment(loader=loader)
 
 default_template = env.get_template('indra/statements_view.html')
 
+reader_order = ['reach', 'sparser', 'medscan', 'trips', 'eidos']
+
+
+def _sort_readers(reader_list: List[str]):
+    """Sort readers according to reader_order"""
+    reader_list.sort(key=lambda r: reader_order.index(r)
```
cthoyt, line 51: `key=lambda r: reader_order.index(r)` is the same as just `key=reader_order.index`

indra/assemblers/html/assembler.py
```diff
@@ -50,18 +106,274 @@ def color_gen(scheme):
...
+    ----------
+    fname :
+        Where to save the stylesheet
+    source_colors :
+        Colors defining the styles
+    """
+    if source_colors is None:
+        source_colors = DEFAULT_SOURCE_COLORS
+
+    rule_string = '.source-{src} {{\n    background-color: {src_bg};\n    ' \
```
cthoyt, line 364: might be overkill, but you could use a jinja template for this

indra/assemblers/html/assembler.py
kkaris: `reader_order` sometimes is only a subset of the provided readers in `reader_list` and errors in that case:
```
In [8]: lst = [8, 23, 99, 7, 9, 0]

In [9]: sub_lst = [8, 0,  9]

In [10]: lst.sort(key=sub_lst.index)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-14-7f03eae0e9ed> in <module>()
----> 1 lst.sort(key=sub_lst.index)

ValueError: 23 is not in list
```

indra/assemblers/html/assembler.py
kkaris: I think that would re-open the rabbit hole that I was in in the previous attempt (#1297) :sweat_smile:

indra/assemblers/html/assembler.py
cthoyt: Wouldn't you get the same error when wrapping it with a lambda?

dianakolusheva: The following if statement handles the error

kkaris: :point_up:

cthoyt: bahhh I'm really sorry I totally didn't see that

pagreene (approved):
Looks good!

indra/assemblers/html/assembler.py
```diff
@@ -23,25 +26,78 @@
...
+SourceInfo = Dict[str, Dict[str, Union[str, Dict[str, str]]]]
+SourceColors = List[Tuple[str, Dict[str, Union[str, Dict[str, str]]]]]
+
 
 loader = FileSystemLoader(join(HERE, 'templates'))
 env = Environment(loader=loader)
 
 default_template = env.get_template('indra/statements_view.html')
 
+reader_order = ['reach', 'sparser', 'medscan', 'trips', 'eidos']
```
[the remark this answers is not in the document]
kkaris: Added some docstrings for the globals!

kkaris:
All the failing tests are related to `arty.scai.fraunhofer.de` being down (I pinged it from the terminal just now)

cthoyt:
I think all of the Fraunhofer resources for BEL have been decommissioned, we should probably go through and fix/update those

PR #1303, opened 2021-06-22
Add Gnbr processor

STIRLIN6:
Add gnbr processor.


setup.py
```diff
@@ -13,7 +13,8 @@ def main():
                     'requests>=2.11', 'lxml', 'ipython', 'future',
                     'networkx>=2', 'pandas', 'ndex2==2.0.1', 'jinja2',
                     'protmapper>=0.0.21', 'obonet', 'sympy==1.3',
-                    'tqdm', 'pybiopax>=0.0.5']
+                    'tqdm', 'pybiopax>=0.0.5', 'copy', 'itertools', 're',
```
cthoyt, line 16: these 4 additions are actually builtin and don't need to be listed here

STIRLIN6: Thanks! Corrected ☑️

PR #1308, opened 2021-07-06
Unify versioning of bioontology

cthoyt:
This PR adds a bioontology command for getting the version in the CLI. Then it tries to update the usage in the GHA config to use it instead of hard-coding the variables

*not sure if the way I set the variable will work so we'll see in the run

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,15 +27,12 @@ jobs:
     - name: Install dependencies
       run: |
         echo $GITHUB_EVENT_NAME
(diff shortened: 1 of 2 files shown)


bgyori:
Nice!

cthoyt:
This does not work since it assumes INDRA is installed. @bgyori can we do the installation of INDRA before downloading the bioontology?

cthoyt:
@bgyori py36 passes and 39 only fails due to an unrelated pandas issue

bgyori (approved):

PR #1309, opened 2021-07-07
Html assembly updates

kkaris:
This PR makes two updates to the `HtmlAssembler`:
- A belief badge is added as an option to the output. The badge is added if the argument `show_belief` is set to `True` when the html model is generated with `HtmlAssembler.make_model()` or `HtmlAssembler.save_model()`.
- The source name for each evidence is now a link to the source, if `source_url` is set in `Evidence.annotations`.

`indra/tests/test_html_assembler.py::test_assembler` is extended to check for the updates.

### DIFF
--- a/indra/assemblers/html/assembler.py
+++ b/indra/assemblers/html/assembler.py
@@ -638,7 +638,8 @@ def _format_evidence_text(stmt, curation_dict=None, correct_tags=None):
                         'original_json': ev.to_json(),
                         'num_curations': num_curations,
                         'num_correct': num_correct,
-                        'num_incorrect': num_incorrect
+                        'num_incorrect': num_incorrect,
+                        'source_url': ev.annotations.get('source_url', '')
                         })
 
     return ev_list
--- a/indra/assemblers/html/assembler.py
+++ b/indra/assemblers/html/assembler.py
@@ -739,14 +739,18 @@ def id_url(ag):
 
 def src_url(ev: Evidence) -> str:
     """Given an Evidence object, provide the URL for the source"""
-    # Get source url from evidence or from SOURCE_INFO as backup.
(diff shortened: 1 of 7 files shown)


indra/assemblers/html/assembler.py
```diff
@@ -637,7 +638,8 @@ def get_role(ag_ix):
                         'original_json': ev.to_json(),
                         'num_curations': num_curations,
                         'num_correct': num_correct,
-                        'num_incorrect': num_incorrect
+                        'num_incorrect': num_incorrect,
+                        'source_url': ev.annotations.get('source_url', '')
```
cthoyt, line 642: why not just splat in all `ev.annotations`?

cthoyt: oh... there could potentially be all sorts of stuff there

kkaris: I guess we could, although it is already accessible under `'original_json': ev.to_json()`

bgyori (commented):
This looks great! I think we could add one more feature: when generating source links, we could expand this `ev.annotations.get('source_url', '')` to a separate function with a bit more logic. Namely, if there is no `source_url` available but the `source_api` of the evidence is a database source (you can use `indra.sources.SOURCE_INFO` to check the `type` of the source), we could use the `link` provided in `SOURCE_INFO` as a fallback (this will typically link to the database's landing page which is still better than nothing).

kkaris:
I suppose the previous errors were only temporary

bgyori:
I've been using this branch and realized one more small thing: it would be good to limit the number of decimal places in the belief badge since in some cases we currently get some very long (like 15 decimal places) numbers. How about limiting to 4 decimal places?

PR #1312, opened 2021-07-08
Add LSPCI to the bio ontology

bgyori:
This PR adds a custom resource file for LSP compounds. The approach here is to only include LSPCIs that actually produce statements from TAS relevant for INDRA, and have a name. These are then added to the BioOntology with a name and `isa` relationships to their members that come from various namespaces.


indra/resources/update_resources.py
```diff
@@ -774,6 +774,40 @@ def update_biomappings():
...
+        if 'LSPCI'not in stmt.subj.db_refs:
+            continue
+        for k, v in stmt.subj.db_refs.items():
+            if k in {'TEXT', 'LSPCI'}:
+                continue
+            lspci_members[stmt.subj.db_refs.get('LSPCI')].add((k, v))
+
+    # We then process the names table in a way that we always prioritize the
+    # first row for each LSPCI since the table is pre-sorted by priority
+    df = pandas.read_csv('lsp_compound_names.csv')
```
cthoyt, line 793:
```suggestion
    df = pandas.read_csv('lsp_compound_names.csv', dtype={'lspci_id': str})
```

this means you don't have to keep casting

indra/tests/test_ontology.py
```diff
@@ -335,4 +335,15 @@ def test_standardize_hgnc_fplx_mesh_bug():
...
+
+def test_lspci():
+    assert bio_ontology.get_name('LSPCI', '18') == 'Pentane-1,5-Diamine'
+    members = bio_ontology.get_children('LSPCI', 18)
+    # These are some of the members, not all
+    expected_members = {('CAS', '462-94-2'),
+                        ('CHEBI', 'CHEBI:18127'),
+                        ('CHEMBL', 'CHEMBL119296'),
+                        ('PUBCHEM', '273')}
+    assert expected_members < set(members)
```
cthoyt, line 349: add trailing newline to reduce diff

indra/tests/test_ontology.py
```diff
@@ -335,4 +335,15 @@ def test_standardize_hgnc_fplx_mesh_bug():
...
+
+
+def test_lspci():
+    assert bio_ontology.get_name('LSPCI', '18') == 'Pentane-1,5-Diamine'
+    members = bio_ontology.get_children('LSPCI', 18)
+    # These are some of the members, not all
+    expected_members = {('CAS', '462-94-2'),
+                        ('CHEBI', 'CHEBI:18127'),
+                        ('CHEMBL', 'CHEMBL119296'),
+                        ('PUBCHEM', '273')}
```
cthoyt, line 348: what about the HMS-lincs ids and the ecommons ones? are they not in TAS?

indra/resources/update_resources.py
```diff
@@ -774,6 +774,40 @@ def update_biomappings():
...
 
 
+def update_lspci():
+    # We first create a dict of LSPCIs and their members but only for ones
+    # that actually have TAS statements corresponding to them
+    from indra.sources import tas
+    tp = tas.process_from_web(affinity_class_limit=10)
+    lspci_members = defaultdict(set)
+    for stmt in tp.statements:
+        if 'LSPCI'not in stmt.subj.db_refs:
```
cthoyt, line 784: check spacing

cthoyt:
Would it be good to gzip the JSON or make JSON lines?

indra/tests/test_ontology.py
bgyori: HMS-LINCS IDs do sometimes show up for TAS compounds (not for this one though), I'm not sure about emolecules, that's probably not relevant from an assembly perspective.

cthoyt: Agreed

bgyori:
> Would it be good to gzip the JSON or make JSON lines?

Yeah, I was thinking about another format too. I have tried to avoid gzipping files since releases are compressed anyway and for version control it's nice to have a text-based format even for resource files. I turned the data into a TSV which made it a lot smaller.

PR #1313, opened 2021-07-21
Context and annotations added to PyBEL graph

bgyori:
This PR updates the PyBEL assembler to add context and annotations from evidences.

### DIFF
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -335,15 +335,12 @@ class PybelAssembler(object):
                             zip(pybel_lists, (stmt.obj_from, stmt.obj_to)):
             for agent in agent_list:
                 node = _get_agent_grounding(agent)
-                # TODO check for missing grounding?
-                pybel_list.append(node)
-        try:
-            rxn_node_data = reaction(
-                reactants=pybel_lists[0],
-                products=pybel_lists[1],
-            )
-        except:
-            return
+                if node is not None:
+                    pybel_list.append(node)
+        rxn_node_data = reaction(
+            reactants=pybel_lists[0],
+            products=pybel_lists[1],
+        )
 
         self.model.add_node_from_data(rxn_node_data)
         obj_edge = None  # TODO: Any edge information possible here?
--- a/indra/assemblers/pybel/assembler.py
+++ b/indra/assemblers/pybel/assembler.py
@@ -337,11 +337,14 @@ class PybelAssembler(object):
                 node = _get_agent_grounding(agent)
                 # TODO check for missing grounding?
                 pybel_list.append(node)
+        try:
+            rxn_node_data = reaction(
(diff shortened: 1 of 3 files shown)


indra/assemblers/pybel/assembler.py
```diff
@@ -337,11 +337,14 @@ def _assemble_conversion(self, stmt):
                 node = _get_agent_grounding(agent)
                 # TODO check for missing grounding?
                 pybel_list.append(node)
+        try:
+            rxn_node_data = reaction(
+                reactants=pybel_lists[0],
+                products=pybel_lists[1],
+            )
+        except:
+            return
```
bgyori, line 346: I'm not sure in what situation an error can happen here, it might make sense to figure this out and handle it explicitly rather than have a generic try/except here. @ddomingof do you happen to remember why this had to be added?

indra/assemblers/pybel/assembler.py
```diff
@@ -653,6 +658,39 @@ def _get_evidence(evidence):
...
+
+        # Add the annotations
+        for key, value in context_annotations:
+            annotations[key] = {value: True}
+
+    if evidence.annotations:
+        ignore_keys = {"agents", "raw_grounding", "bel", "prior_uuids",
+                       "provenance", "found_by", "text_refs", "obj_polarity",
+                       "obj_adjectives", "ANNOTATOR", "obj_adjectives",
+                       "SEQUENCE", "subj_context", "obj_context", ""}
```
bgyori, line 681: I'm not sure that adding any annotation we come across other than ones listed here is a good idea. We might consider flipping the logic here and allowing the parameterization of which annotations keys to include explicitly while ignoring all others.

ddomingof: Yes you are right. Controlling what comes in makes sense.

indra/assemblers/pybel/assembler.py
ddomingof: Yes, there errors with some weird nodes that were not grounded and were part of a reaction. There were not many but it was required to catch the error. We can remove this line as this is not necessarily something that should be handled here

ddomingof: Unfortunately, I cant push to your branch but yes, I would just remove this,

ddomingof (approved):

indra/assemblers/pybel/assembler.py
bgyori: I'm pretty sure I figured out what the corner case was and handled it directly and removed the try/except.

indra/assemblers/pybel/assembler.py
bgyori: I changed the implementation to add an argument to the PybelAssembler called `annotations_to_include` that allows controlling externally which annotation entries to include in the PyBEL graph.

PR #1320, opened 2021-08-03
Add module for fixing common statement invalidities

bgyori:
This PR adds a new module to fix some common invalidities in INDRA Statements. It's important to note that these have all been addressed in #1172 at the level of source processors. However, many invalidities "live on" in various places where old statements are persisted so it's useful to have these in-memory statement fixing functions in place as well.

### DIFF
--- a/indra/tools/fix_invalidities.py
+++ b/indra/tools/fix_invalidities.py
@@ -1,6 +1,10 @@
+__all__ = ['fix_invalidities', 'fix_invalidities_db_refs',
+           'fix_invalidities_agent', 'fix_invalidities_context',
+           'fix_invalidities_stmt', 'fix_invalidities_evidence']
+
 import re
 import copy
-from typing import List
+from typing import List, Mapping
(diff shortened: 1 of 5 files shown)


indra/tools/fix_invalidities.py
```diff
@@ -0,0 +1,155 @@
...
+                db_refs['UP'] = v.split(',')[0]
+            if v.startswith('SL-'):
+                db_refs['UPLOC'] = db_refs.pop('UP')
+        elif k == 'UAZ':
+            db_refs.pop('UAZ')
+            if v.startswith('CVCL'):
+                db_refs['CVCL'] = v
+        elif k == 'TAXONOMY' and v == '-1':
+            db_refs.pop('TAXONOMY', None)
+        elif k == 'LINCS' and re.match(r'\d+-\d+', v):
```
cthoyt, line 124: sorry for the late comment but it might be good to pre-compile these regular expressions so it doesn't have to compile it inline every time

PR #1323, opened 2021-08-10
Assembling IndraNet with custom preassembly

dianakolusheva:
This PR extends the IndranetAssembler class by adding an alternative method to assemble networks utilizing custom preassembly functionality. It also adds two custom preassembly functions to assemble signed and unsigned graphs. Graphs assembled with data frame method and with preassembly method are identical except for the edge beliefs. Preassembly method allows for more robust belief calculation using any belief scorer because we can work with statement objects directly.

### DIFF
--- a/indra/assemblers/indranet/assembler.py
+++ b/indra/assemblers/indranet/assembler.py
@@ -288,8 +288,7 @@ class IndraNetAssembler():
                 logger.debug('Skipping a statement of a type %s.'
                              % type(stmt).__name__)
                 continue
-            agents = stmt.agent_list()
-            not_none_agents = [a for a in agents if a is not None]
+            not_none_agents = stmt.real_agent_list()
 
             # Exclude statements with less than 2 agents
             if len(not_none_agents) < 2:
@@ -417,8 +416,8 @@ class IndraNetAssembler():
             IndraNet graph object.
         """
         # Filter out statements with one agent or with None subject
-        stmts = [stmt for stmt in self.statements if len(
-            [ag for ag in stmt.agent_list() if ag is not None]) > 1]
+        stmts = [stmt for stmt in self.statements
+                 if len(stmt.real_agent_list()) > 1]
         if exclude_stmts:
             exclude_types = tuple(
                 get_statement_by_name(st_type) for st_type in exclude_stmts)
@@ -452,7 +451,7 @@ class IndraNetAssembler():
             graph_stmts = [stmt for stmt in stmts if stmt not in complex_stmts
                            and stmt not in conv_stmts]
             for stmt in complex_stmts:
-                agents = [ag for ag in stmt.agent_list() if ag is not None]
+                agents = stmt.real_agent_list()
                 if len(agents) > complex_members:
                     continue
                 for a, b in permutations(agents, 2):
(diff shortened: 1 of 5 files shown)


indra/assemblers/indranet/assembler.py
```diff
@@ -48,10 +53,93 @@ def add_statements(self, stmts):
...
-                   extra_columns=None):
+                   belief_flattening=None, belief_scorer=None,
+                   weight_flattening=None, extra_columns=None):
+        """Assemble an IndraNet graph object.
+
+        Parameters
+        ----------
+        method : str
+            Method for assembling an IndraNet graph. Accepted values: `df` and
+            `preassembly`.
```
bgyori, line 66: Could you add a couple more sentences here to explain the difference here for posterity?

indra/assemblers/indranet/assembler.py
```diff
@@ -281,9 +369,173 @@ def make_df(self, exclude_stmts=None, complex_members=3,
...
+            ...     return G
+
+        Returns
+        -------
+        model : IndraNet
+            IndraNet graph object.
+        """
+        # Filter out statements with one agent or with None subject
+        stmts = [stmt for stmt in self.statements if len(
+            [ag for ag in stmt.agent_list() if ag is not None]) > 1]
```
bgyori, line 421: You can use `stmt.real_agent_list()` instead of `stmt.agent_list() if ag is not None`

indra/assemblers/indranet/assembler.py
```diff
@@ -281,9 +369,173 @@ def make_df(self, exclude_stmts=None, complex_members=3,
...
+                    graph_stmts.append(
+                        DecreaseAmount(stmt.subj, obj, stmt.evidence))
+                for obj in stmt.obj_to:
+                    graph_stmts.append(
+                        IncreaseAmount(stmt.subj, obj, stmt.evidence))
+            if graph_type == 'unsigned':
+                graph_stmts = ac.run_preassembly(
+                    graph_stmts, return_toplevel=False,
+                    belief_scorer=belief_scorer,
+                    matches_fun=agent_name_stmt_matches)
```
bgyori, line 471: I wonder if we should also take refinements into account here, without also supplying a custom `refinement_fun` along with a custom `matches_fun`, issues could come up. Alternatively, if we want to cleanly ignore refinements here, we should pass the `run_refinement=False` argument here.

indra/preassembler/custom_preassembly.py
```diff
@@ -37,3 +41,22 @@ def agent_name_stmt_type_matches(stmt):
     agents = [agent_name_matches(a) for a in stmt.agent_list()]
     key = str((stmt.__class__.__name__, agents))
     return key
+
+
+@register_pipeline
+def agent_name_stmt_matches(stmt):
+    """Return the normalized agent names."""
+    agents = [agent_name_matches(a) for a in stmt.agent_list()]
```
bgyori, line 49: Actually, `agent_name_matches`, as implemented above does some non-standard string processing on the name which I implemented for a specific custom assembly. I would not use it here directly, instead, we could either consider two agents to match if the actual, unmodified `agent.name` matches, or we could do matching based on grounding with name as a fallback.

bgyori:
Nice!

PR #1326, opened 2021-09-19
Add some types to the BioOntology and extend resources

bgyori:
This PR adds types for many of the entries in the BioOntology. The focus is on annotating entries in namespaces that are being normalized to in standard assembly workflows. Two resources and their clients are also extended:
- The MeSH supplementary resource file now contains the primary terms that a supplementary concept is mapped to. This is useful for many use cases, here it is used to find an appropriate type for supplementary concepts.
- The HGNC resource file and client now exposes locus types for human genes.

### DIFF
--- a/indra/databases/hgnc_client.py
+++ b/indra/databases/hgnc_client.py
@@ -2,6 +2,7 @@ import os
 import re
 import logging
 import requests
+from typing import Union
 import xml.etree.ElementTree as ET
 from functools import lru_cache
 
@@ -268,6 +269,25 @@ def get_hgnc_entry(hgnc_id):
     return xml_tree
 
 
+def get_gene_type(hgnc_id: str) -> Union[str, None]:
+    """Return the locus type of the genve with the given HGNC ID.
+
+    See more under Locus type at
+    https://www.genenames.org/help/symbol-report/#!/#tocAnchor-1-2
+
+    Parameters
+    ----------
+    hgnc_id :
+        The HGNC ID of the gene to get the locus type of.
+
+    Returns
+    -------
+    :
+        The locus type of the given gene.
+    """
+    return gene_type.get(hgnc_id)
+
+
 def is_kinase(gene_name):
     """Return True if the given gene name is a kinase.
 
@@ -334,6 +354,7 @@ def _read_hgnc_maps():
     ensembl_ids = {}
     ensembl_ids_reverse = {}
     hgnc_withdrawn_new_ids = {}
+    gene_types = {}
     # Skip the header
     next(csv_rows)
     for row in csv_rows:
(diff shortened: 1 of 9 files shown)


indra/ontology/bio/ontology.py
```diff
@@ -157,7 +163,8 @@ def add_famplex_nodes(self):
                 os.path.join('famplex', 'entities.csv')), delimiter=','):
             entity = row[0]
             nodes.append((self.label('FPLX', entity),
-                          {'name': entity}))
+                          {'name': entity,
+                           'type': 'protein_family_complex'}))
```
cthoyt, line 167: I guess it's not so simple but it would be nice to get a mixture of strict protein families, strict complexes, and mixtures (i.e., famplexes) from the famplex resource

indra/tests/test_mesh.py
```diff
@@ -115,3 +115,17 @@ def test_mesh_mapping():
...
         ('CHEBI', 'CHEBI:4672')
+
+
+def test_supplementary_to_primary():
+    from indra.statements.validate import assert_valid_id
+    for supp, primaries in mesh_client.mesh_supp_to_primary.items():
+        for primary in primaries:
+            assert_valid_id('MESH', primary)
+    mesh_client.get_primary_mappings('C009879') == ['D011140']
+    mesh_client.get_primary_mappings('C114158') == ['D005063', 'D009942']
```
cthoyt, line 126: I was under the impression that there were things in the supplement that didn't have a corresponding primary record. Is this not true?

indra/databases/hgnc_client.py
```diff
@@ -268,6 +269,25 @@ def get_hgnc_entry(hgnc_id):
     return xml_tree
 
 
+def get_gene_type(hgnc_id: str) -> Union[str, None]:
```
cthoyt, line 272: `Union[str, None]` is the same as `Optional[str]`

bgyori: Right, I was thinking, Optional sounds more intelligible for inputs parameters and Union for return values so went with this.

indra/ontology/bio/ontology.py
bgyori: That would be possible, I didn't do it here just because that further stratification wasn't needed for statistics but we could implement that. I was actually thinking of potentially adding another layer called subtype for things like this.

indra/tests/test_mesh.py
bgyori: I actually didn't find any examples of that, all the supplementary concepts are mapped to one or more primary term.

indra/ontology/bio/ontology.py
steppi: > I guess it's not so simple but it would be nice to get a mixture of strict protein families, strict complexes, and mixtures (i.e., famplexes) from the famplex resource

I played around with implementing something like this when I made famplex into a python package last year. I just tried the simple thing where if all of the relationships beneath a term were `isa` then it's a family, if all of these edges are `partof`, then it's a complex, and if there's a mixture of edge types then it's a famplex. I'd written a function to find the type dynamically, but ended up not including it.

indra/tests/test_mesh.py
cthoyt: Fascinating! I will have to look into this and definitely improve the pyobo mesh importer

indra/databases/hgnc_client.py
cthoyt: Now that you mention it, I actually really like this

indra/tests/test_mesh.py
steppi: Yeah, that's actually something they enforce when they make supplementary concepts. Sometimes the mappings are to something much higher up in the ontology, but there's always at least one mapping.

```
Supplementary Records
Supplementary Records, also called Supplementary Chemical Records(SCRs), are used to index chemicals, drugs, and other concepts such as rare diseases for MEDLINE and are searchable by Substance Name [NM] in PubMed. Unlike Descriptors, SCRs are not organised in a tree hierarchy. Instead each SCR is linked to one or more Descriptors by the Heading Mapped To (HM) field in the SCR. They also include a Indexing Information (II) field field that is used to refer to other descriptors that are from related topics. SCRs are created daily and distributed nightly Monday-Thursday. There are currently over 230,000 SCR records with over 505,000 SCR terms. Like all MeSH records, SCRs are searchable in the MeSH Browser. Four classes of SCRs exist. 
```
https://www.nlm.nih.gov/mesh/intro_record_types.html

cthoyt: Oh that makes more sense

PR #1351, opened 2021-11-10
Expand all with query parameter

kkaris:
This PR adds the possibility to expand all statements after the page has loaded by adding `expand_all=true` as a query parameter.

Test by running any test in `test_html_assembler.py` that produces an html page and adding `expand_all=true` to the parameters when viewing it in the browser.


bgyori (approved):
This works great, thanks!

PR #1355, opened 2021-11-12
Function to retrieve PubMedIDs for given PubChem CID 

samuelbunga:
This PR consists of a new function that takes PubChem CID as an input and returns back a list of PMIDs corresponding to it.
```
from indra.databases.pubchem_client import get_pmids 
pmids = get_pmids('22442') # aspirin
```

### DIFF
--- a/indra/databases/pubchem_client.py
+++ b/indra/databases/pubchem_client.py
@@ -78,18 +78,21 @@ def get_preferred_compound_ids(pubchem_cid):
 
 
 def get_pmids(pubchem_cid: str) -> List[str]:
-    """Return xref PMIDs for a given PubChem CID.
+    """Return depositor provided PMIDs for a given PubChem CID.
+
+    Note that this information can also be obtained via PubMed
+    at https://www.ncbi.nlm.nih.gov/sites/entrez?LinkName=pccompound_pubmed&db=pccompound&cmd=Link&from_uid=<pubchem_cid>.
 
     Parameters
     ----------
     pubchem_cid :
-        The PubChem CID whose xref PubMedID's will be returned.
+        The PubChem CID whose PMIDs will be returned.
 
     Returns
     -------
-    list of str
-        PubMedIDs corresponding to the given PubChem CID. If none present,
-        an empty list is returned.
+    :
+        PMIDs corresponding to the given PubChem CID. If none present,
+        or the query fails, an empty list is returned.
     """
     url = '%s/compound/cid/%s/xrefs/PubMedID/JSON' % \
           (pubchem_url, pubchem_cid)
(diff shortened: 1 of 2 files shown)


indra/databases/pubchem_client.py
```diff
@@ -74,3 +75,28 @@ def get_preferred_compound_ids(pubchem_cid):
...
     pref_ids = sorted([str(pid) for pid in pref_ids])
     return pref_ids
+
+
+def get_pmids(pubchem_cid: str) -> List[str]:
+    """Return xref PMIDs for a given PubChem CID.
+
+    Parameters
+    ----------
+    pubchem_cid : str
```
cthoyt, line 85: type annotation not necessary here

indra/databases/pubchem_client.py
```diff
@@ -74,3 +75,28 @@ def get_preferred_compound_ids(pubchem_cid):
...
+        The PubChem CID whose xref PubMedID's will be returned.
+
+    Returns
+    -------
+    list
```
cthoyt, line 90: Documentation of type is not complete - it's actually better to just have it as actual type hints
```suggestion
    :
```

indra/tests/test_pubchem_client.py
```diff
@@ -4,3 +4,9 @@
 def test_get_inchi_key():
     ik = pubchem_client.get_inchi_key('5280613')
     assert ik == 'JLVSPVFPBBFMBE-HXSWCURESA-O', ik
+
+
+def test_get_pmids():
```
cthoyt, line 9: add a test for an obviously wrong PMID and also for an okay looking one that's invalid

indra/tests/test_pubchem_client.py
```diff
@@ -4,3 +4,9 @@
...
+    pmids = pubchem_client.get_pmids('2244')
+    example_pmid = '19036898'
+    assert pmids[0] == example_pmid
```
bgyori, line 12: I wouldn't do it this way since the order of PMIDs could be non-deterministic or change over time. How about
```
assert example_pmid in pmids
```

indra/databases/pubchem_client.py
samuelbunga: I just fixed it. It will now return a list of 'str'

indra/databases/pubchem_client.py
```diff
@@ -74,3 +75,28 @@ def get_preferred_compound_ids(pubchem_cid):
...
+    """Return xref PMIDs for a given PubChem CID.
+
+    Parameters
+    ----------
+    pubchem_cid :
+        The PubChem CID whose xref PubMedID's will be returned.
+
+    Returns
+    -------
+    list of str
```
cthoyt, line 90:
```suggestion
    :
```

don't forget this ^ it's important to make sure the doc strings don't contain these so they don't get out of sync and so they're properly programatically parsed. The `:` is the placeholder

bgyori: I agree, I changed this

bgyori (approved):

PR #1361, opened 2021-12-06
Use latest MINERVA project

dianakolusheva:
This PR updates the MINERVA client to use the latest available project instead of a fixed default version.


bgyori (approved):
Looks good, thanks!

PR #1372, opened 2022-03-09
Add base assembler class

cthoyt:
Adding a base assembler class allows for other assemblers to inherit some shared functionality (or at least unify the interface a bit).

For example, I find myself doing this quite a bit:

```
from indra.assemblers import SomethingAssembler

assembler = SomethingAssembler(stmts)
model = assembler.make_model()
```

And with the base class, it's possible to just do

```
from indra.assemblers import SomethingAssembler

model = SomethingAssembler.model_from_stmts(stmts)
```

### DIFF
--- a/indra/assemblers/utils.py
+++ b/indra/assemblers/utils.py
@@ -1,0 +1,22 @@
+"""Utilities for INDRA assemblers."""
+
+from typing import Generic, TypeVar
+from abc import ABC, abstractmethod
+
+__all__ = [
+  "Assembler",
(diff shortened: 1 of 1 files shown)


bgyori:
Not sure. The point is that the constructor and make_model can take different arguments and have different semantics. Sure, a shortcut can be useful in some cases but not critical.

cthoyt:
Alright, I'll close this (it can also be accomplished with a helper function)

PR #1373, opened 2022-03-15
Add EC client

cthoyt:
This PR does the following:

1. Implements PyOBO-based ontology loader
2. Adds EC resource
3. Adds an EC client and tests
4. Adds EC content to Bio-ontology (updates mock ontology and adds tests)


bgyori (approved):
I fixed a number of problems with the PR, looks like it works now.

PR #1378, opened 2022-05-05
Fix handling of xrefs from OBO

bgyori:
This PR fixes and issue in integrating xrefs from OBO-derived resources and increases the number of cross-references in the bioontology.

### DIFF
--- a/indra/ontology/bio/ontology.py
+++ b/indra/ontology/bio/ontology.py
@@ -6,6 +6,7 @@ from ..ontology_graph import IndraOntology
 from indra.util import read_unicode_csv
 from indra.statements import modtype_conditions
 from indra.resources import get_resource_path
+from indra.databases.identifiers import ensure_prefix_if_needed
 from indra.statements.validate import assert_valid_db_refs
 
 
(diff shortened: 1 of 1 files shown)


bgyori:
Integrating xrefs between OBO-sourced IDs actually turns out to be problematic for several reasons:
1. Several xrefs point to obsolete or non-existent entries in other ontologies
2. Several xrefs are technically valid but simply incorrect, e.g., 'GO:GO:0140446' -> 'CHEBI:CHEBI:147341'
3. Some OBOs like MONDO put replaced-by relations as xrefs to entries in the ontology itself, these are currently picked up without further qualification as if they were normal xrefs.
4. There are non-trivial relationships with mappings from e.g., Biomappings that should be reconciled.

1 and 3 are relatively easy to address. I'm worried about 2, one potential solution being to restrict which namespaces we integrate mappings between to exclude e.g., GO-CHEBI.

cthoyt:
@bgyori can we revisit this? I think it will solve the issue I showed last friday on cogex that the MONDO term for asthma wasn't connected to the rest of the asthma terms with an xref relation

PR #1379, opened 2022-05-05
Add xrefs endpoint using IndraOntology.get_mappings()

kkaris:
This PR adds a cross-reference endpoint to the ontology web app, wrapping `indra.ontology.IndraOntology.get_mappings()`.

### DIFF
--- a/indra/ontology/app/app.py
+++ b/indra/ontology/app/app.py
@@ -59,6 +59,15 @@ def get_component_label():
         **{k: v for k, v in request.json.items() if k in kwargs}))
 
 
+@app.route('/get_xrefs', methods=['GET'])
+def get_xrefs():
+    ont = request.json.get('ontology')
+    ontology = ontologies.get(ont)
+    kwargs = ('ns', 'id')
+    return jsonify(ontology.get_mappings(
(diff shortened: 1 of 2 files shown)


indra/ontology/app/app.py
```diff
@@ -59,6 +59,15 @@ def get_component_label():
         **{k: v for k, v in request.json.items() if k in kwargs}))
 
 
+@app.route('/get_xrefs', methods=['GET'])
+def get_xrefs():
+    ont = request.json.get('ontology')
+    ontology = ontologies.get(ont)
+    kwargs = ('ns', 'id')
+    return jsonify(ontology.get_mappings(
+        **{k: v for k, v in request.json.items() if k in kwargs}))
```
bgyori, line 68: This function takes a single `ns` and `id` argument so the use of the dict iteration seems odd, why not just use
```
ontology.get_mappings(request.json['ns'], request.json['id'])
```

bgyori: Huh, looks like I started this pattern in prior endpoints, and I can't recall why it would have made sense to do this, other than for optional arguments...

bgyori (approved):

PR #1386, opened 2022-07-18
Eliminate isa/partof ontology cycles

bgyori:
This PR adds some exclusion criteria to eliminate any directed cycles in the isa/partof subgraph of the BioOntology graph which can induce cycles during Statement refinement finding.


bgyori:
The test failures are due to external issues in the Biolookup and TRIPS services.

PR #1396, opened 2022-10-26
Add optional residue/position feature to CountScorer

johnbachman:
Adds a feature to the CountsScorer to indicate whether a statement has (not-None) residue and position fields. When used to train a Belief model on Modification statements that have been sitemapped, this allows the model to exploit the information that the extracted residue and position corresponds to the target substrate (i.e., increase the likelihood that those elements of the statement are correct).


bgyori (approved):
Looks great, thanks!

PR #1397, opened 2022-10-28
Single quotes don't work for Windows

kwalcock:
They are OK for PowerShell, but don't work on the old command prompt.


bgyori:
Thank you @kwalcock!

PR #1406, opened 2023-03-25
Allow enhanced author parsing

pagreene:
This is a small PR that adds the option to get full details given about an author in PubMed XMLs. The default behavior remains the same, however there is a new keyword argument in the `get_metadata_from_xml_tree` function called `detailed_authors` that can be set to True to get this behavior.


bgyori (approved):
Thanks @pagreene, this is a good addition! I added this option to one of the higher level functions and extended an existing test to check the content with/without the new flag.

PR #1416, opened 2023-06-29
Journal info

kkaris:
This PR extends the pubmed client to be able to extract publication as well as issue information from pubmed medline XML data. Two new namespaces are also added for validation.

## todo
- [x] Update protmapper and re-run tests when https://github.com/gyorilab/protmapper/pull/45 is merged

### DIFF
--- a/indra/literature/pubmed_client.py
+++ b/indra/literature/pubmed_client.py
@@ -323,9 +323,9 @@ def _get_issue_info(journal: ET.Element):
         issue_year = None
 
     return {
-        "issue_volume": issue_volume,
-        "issue_issue": issue_issue,
-        "issue_year": issue_year
+        "volume": issue_volume,
+        "issue": issue_issue,
+        "year": issue_year
     }
 
 
@@ -365,9 +365,9 @@ def get_issn_info(
                 "other": [str, ...]
             },
             "issue_dict": {
-                "issue_volume": str,
+                "volume": str,
                 "issue": str,
-                "issue_year": int
+                "year": int
             }
         }
     """
@@ -466,9 +466,9 @@ def _get_journal_info(medline_citation, get_issns_from_nlm: bool):
         'issn_list': issn_list,
         'issn_l': issn_linking,
         'journal_nlm_id': nlm_id,
-        'issue': issue_info['issue_issue'],
-        'issue_volume': issue_info['issue_volume'],
-        'issue_year': issue_info['issue_year'],
+        'issue': issue_info['issue'],
+        'volume': issue_info['volume'],
+        'year': issue_info['year'],
     }
 
 
@@ -631,7 +631,7 @@ def get_metadata_from_xml_tree(tree, get_issns_from_nlm=False,
         Dictionary indexed by PMID. Each value is a dict containing the
         following fields: 'doi', 'title', 'authors', 'journal_title',
         'journal_abbrev', 'journal_nlm_id', 'issn_list', 'page',
-        'issue_volume', 'issue', 'issue_pub_date'.
+        'volume', 'issue', 'issue_pub_date'.
     """
     # Iterate over the articles and build the results dict
     results = {}
--- a/indra/literature/pubmed_client.py
+++ b/indra/literature/pubmed_client.py
@@ -792,7 +792,8 @@ def get_issns_for_journal(nlm_id):
     tree = send_request(pubmed_fetch, params)
     if tree is None:
         return None
-    issn_list = [(e.attrib, e.text) for e in tree.findall('.//ISSN')]
+    issn_list = [(e.attrib.get("IssnType", "other"), e.text)
+                 for e in tree.findall('.//ISSN')]
     issn_linking = tree.find('.//ISSNLinking')
     issn_l = issn_linking.text if issn_linking else None
     issn_dict = {
--- a/indra/literature/pubmed_client.py
+++ b/indra/literature/pubmed_client.py
@@ -329,6 +329,83 @@ def _get_issue_info(journal: ET.Element):
     }
 
 
+def get_issn_info(medline_citation: ET.Element,
+                  get_issns_from_nlm: bool = False):
+    """Given a medline citation, get the issn info from the article
+
+    Parameters
+    ----------
+    medline_citation : xml.etree.ElementTree.Element
+        The MedlineCitation element of the PubMed XML tree.
+    get_issns_from_nlm : bool
+        If True, get the ISSN from the NLM catalog as well.
+
+    Returns
+    -------
+    dict
+        A dictionary journal, issue, and ISSN info. The structure is as
+        follows:
+        {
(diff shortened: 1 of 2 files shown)


indra/literature/pubmed_client.py
```diff
@@ -307,12 +307,130 @@ def _find_elem_text(root, xpath_string):
...
+    if issue_pub_date is not None:
+        # Get issue year
+        issue_year = _find_elem_text(issue_pub_date, 'Year')
+        issue_year = int(issue_year) if issue_year else None
+
+    else:
+        issue_year = None
+
+    return {
+        "issue_volume": issue_volume,
```
cthoyt, line 326: the prefix of `issue_` is redundant since this dict is inside another dict with key `issue_dict` -> just make "volume", "issue", "year"

indra/literature/pubmed_client.py
```diff
@@ -307,12 +307,130 @@ def _find_elem_text(root, xpath_string):
...
+    # Get the ISSN from the article record
+    issn_dict = {}
+    issn_element = journal.find("ISSN")
+    if issn_element is not None:
+        issn_type = issn_element.attrib["IssnType"]
+        issn = issn_element.text
+        issn_dict["issn"] = issn
+        if issn_type.lower() in ["electronic", "print"]:
+            type_prefix = "p_" if issn_type.lower() == "print" else "e_"
+            issn_dict[type_prefix + "issn"] = issn
```
cthoyt, line 394: why not just create a data structure that captures this in a structured way, rather than adding cryptic keys? Is it possible there are multiple print/electronic, or even other types? I definitely saw this scenario on wikidata. Instead, you could make a list like
```
[
  { "issn": XXX, "type": "print" },
  { "issn": YYY, "type": "electronic" },
  { "issn": ZZZ, "type": "other" },
]
``

File: indra/literature/pubmed_client.py [hunk 3]
```diff
@@ -662,14 +795,26 @@ def get_issns_for_journal(nlm_id):
...
+    issn_linking = tree.find('.//ISSNLinking')
+    issn_l = issn_linking.text if issn_linking else None
+    issn_dict = {}
+    for issn_type, issn in issn_list:
+        if issn_type.lower() in ('print', 'electronic'):
+            type_prefix = "p_" if issn_type.lower() == 'print' else "e_"
+            issn_dict[type_prefix + "issn"] = type_prefix + issn
+        elif issn != issn_l:
+            if 'other' in issn_dict:
+                issn_dict['other'].append(issn)
```
cthoyt, line 809: use a defaultdict to reduce complexity

indra/literature/pubmed_client.py
kkaris: Per journal you typically only have the ISSN (whether it be 'print', 'electronic' or other categories) and the ISSN-L (none of them are guaranteed to be present though, but the NLM ID is as far as I've seen), so I don't think we need to do a list, but I agree that the structure can be simplified.

kkaris:
All the failing tests seem to be related to the uniprot client and not to the code in this PR

kkaris:
At least some of the access to uniprot seem to have moved to `https://rest.uniprot.org/uniprotkb/` from `https://legacy.uniprot.org/uniprot/` - 30 of 31 tests pass when updating `uniprot_url` at https://github.com/gyorilab/protmapper/blob/master/protmapper/uniprot_client.py#L18

https://www.uniprot.org/help/api_retrieve_entries

kkaris:
I see now that there is an open PR for this in protmapper: https://github.com/gyorilab/protmapper/pull/45

indra/literature/pubmed_client.py
```diff
@@ -4,7 +4,7 @@
 import logging
 import requests
 from time import sleep
-from typing import List
+from typing import List, Literal
```
cthoyt, line 7: If you want to import `Literal` this way, we need to make explicit that the minimum version is now python 3.8. Otherwise you can do `from typing_extensions import Literal` which backports all of this to earlier versions

cthoyt: Personally, I think it's fine to bump to minimum 3.8 since 3.7 is past end of life

indra/literature/pubmed_client.py
```diff
@@ -307,37 +307,157 @@ def _find_elem_text(root, xpath_string):
...
+    # Issue info
+    issue_info = _get_issue_info(journal)
+
     # Add the ISSN from the article record
-    issn_list = []
+    issn_set = set()
     issn = _find_elem_text(journal, 'ISSN')
     if issn:
-        issn_list.append(issn)
+        issn_set.add(issn)
```
cthoyt, line 432: do you need to keep track of what the publication type for each ISSN is? maybe this should be a dict?

indra/literature/pubmed_client.py
bgyori: I'm not thrilled about a breaking change for the sake of a type annotation.

kkaris:
Hmm, I reran the tests because the previous failed tests seemed to be related to temporary connection issues (they passed locally), but now the tests stop at the build with:
```
Collecting kappy==4.1.2 (from indra==1.22.0)
  Downloading kappy-4.1.2-cp38-cp38-manylinux2010_x86_64.whl (12.3 MB)
     ━━━━━━━━━━                               3.1/12.3 MB 1.9 MB/s eta 0:00:05
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    kappy==4.1.2 from https://files.pythonhosted.org/packages/ac/a4/306c4aa4aa178df434a56be20225b3429db20e901c38939047436d0fe2f6/kappy-4.1.2-cp38-cp38-manylinux2010_x86_64.whl (from indra==1.22.0):
        Expected sha256 b8de8bcd857823215fcc3c184f68de5f1d80dc3b3352107749b8156f85d32ff4
             Got        08ec5ddf5296036acf34a806f5e47957a61774715842ba9c824ee4897e84f21b
```

indra/literature/pubmed_client.py
kkaris: I can remove it

kkaris: Resolved in https://github.com/sorgerlab/indra/pull/1416/commits/7877aa10d19c5be755a966cdb7f5c616b8904501

indra/literature/pubmed_client.py
kkaris: That ISSN type is available if using `get_issn_info` further up in the file.

The function where this line is from (`_get_journal_info`) is nested in other functions where there is an expectation of `issn_list` being present as a list of ids, so changing code in this function would potentially include updating downstream usage as well so I'd argue it's outside the scope of this PR.

PR #1424, opened 2023-10-26
Implement retrieving more than 10k PMIDs and all metadata

bgyori:
This PR adds a simple wrapper around PubMed's edirect CLI (https://www.ncbi.nlm.nih.gov/books/NBK179288/) to retreive PMIDs such that we can easily get all PMIDs for queries that return more than 10k results (this turns out to be very convoluted to solve with the REST API even if not entirely impossible).

It also adds a wrapper around a function for getting metadata to allow for a single function call above the limit of 200 PMIDs per metadata request call.

### DIFF
--- a/indra/literature/pubmed_client.py
+++ b/indra/literature/pubmed_client.py
@@ -1,6 +1,8 @@
 """
 Search and get metadata for articles in Pubmed.
 """
+import time
+import tqdm
 import logging
 import random
 import subprocess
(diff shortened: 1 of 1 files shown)


indra/literature/pubmed_client.py
```diff
@@ -845,3 +850,35 @@ def get_substance_annotations(pubmed_id: str) -> List[str]:
...
+        A term for which the PubMed search should be performed.
+
+    Returns
+    -------
+    list[str]
+        A list of PMIDs for the given search term.
+    """
+    cmd = f'esearch -db pubmed -query "{search_term}" | efetch -format uid'
+    res = subprocess.getoutput(cmd)
+    # Output is divided by new lines
```
cthoyt, line 879: I'd add in the following code to give an error if the class path isn't set up properly, since otherwise it returns the following list: `['bin/sh: esearch: command not found', '/bin/sh: efetch: command not found']`

```python
    if "esearch: command not found" in res:
        raise RuntimeError("esearch is not properly on the filepath")
```

PR #1426, opened 2023-11-20
Add more robust retry to the pubmed client's send_request

kkaris:
This PR adds a more robust retry architecture for the `send_request` function in the pubmed client by adding a recursive retry with a randomly increased sleep time between retries.

The setup has been stress-tested with four simultaneous test runs of `tests/test_pubmed_client.py `.


kkaris:
The failed test is from `test_db_rest`, I'll check that out as well.

kkaris:
Re-ran the tests so that they would run simultaneously

bgyori (approved):

PR #1433, opened 2024-01-12
Remove mesh cycle from bio-ontology

kkaris:
This PR adds functionality to remove edges from the bio ontology and does that for one edge in the MeSH 2024 data. The edge removed is `MESH:D015835 -[isa]-> MESH:D013285` which prevents a circular hierarchy structure since `MESH:D013285 -[isa]-> MESH:D015835` is already a relation. See https://meshb.nlm.nih.gov/record/ui?ui=D015835 under the MeSH Tree Structure tab then check the branch for `Nervous System Diseases [C10]` where the hierarchical path `MESH:D015835 -> MESH:D013285 -> MESH:D015835` is present, creating an `is_a/partof` cycle.

The bio ontology version is bumped to 1.32 with this PR.

### DIFF
--- a/indra/ontology/bio/ontology.py
+++ b/indra/ontology/bio/ontology.py
@@ -26,7 +26,7 @@ class BioOntology(IndraOntology):
     # should be incremented to "force" rebuilding the ontology to be consistent
     # with the underlying resource files.
     name = 'bio'
-    version = '1.31'
+    version = '1.32'
     ontology_namespaces = [
         'go', 'efo', 'hp', 'doid', 'chebi', 'ido', 'mondo', 'eccode',
     ]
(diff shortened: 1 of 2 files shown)


kkaris:
All the failing tests are related to 404 errors from the cbio client

indra/ontology/bio/ontology.py
```diff
@@ -634,6 +644,21 @@ def assert_valid_node(self, label):
...
+    def remove_edges(self, edges_to_remove):
+        initial_edge_count = len(self.edges)
+        for source, target, e_type in edges_to_remove:
+            if (source, target) in self.edges:
+                # Check that the edge type matches
+                if self.edges[source, target]['type'] != e_type:
+                    continue
+
+                # Remove the edge
+                self.remove_edge(source, target)
```
bgyori, line 656: It is unlikely to play a big role here but removing edges one by one tends to be slower than using `remove_edges_from` with a list argument. We can leave it like this for simplicity though.

bgyori (approved):

PR #1446, opened 2024-06-05
Replace inspect.getargspec with inspect.getfullargspec

### LINKED ISSUE #1445: Replace `inspect.getargspec` with `inspect.getfullargspec` (or possibly `inspect.signature`)
`inspect.getargspec` is deprecated since 3.0(!), was removed in 3.11 and is replaced by `inspect.getfullargspec`. `inspect.signature` is also mentioned as the python 3 preferred way to deal with introspection, but `inspect.getfullargspec` seems to be safe from removal (for now) so easiest is to just replace it.

This affects the pipeline module: https://github.com/sorgerlab/indra/blob/master/indra/pipeline/pipeline.py#L323

See the Python 3.10 docs for the last version mentioning of `inspect.getargspec`:
https://docs.python.org/3.10/library/inspect.html#inspect.getargspec

kkaris:
This PR resolves #1445.

### DIFF
--- a/indra/pipeline/pipeline.py
+++ b/indra/pipeline/pipeline.py
@@ -320,7 +320,7 @@ class AssemblyPipeline():
             new_kwargs['statements'] = statements
         if kwargs:
             for k, v in kwargs.items():
-                if k not in new_kwargs and k in inspect.getargspec(func).args:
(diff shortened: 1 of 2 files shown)


kkaris:
At least the inspect module issue is already fixed in #1447

kkaris:
Superseeded by #1452 and #1447

PR #1452, opened 2024-07-02
Update Flask, Flask Restx, unpin Gilda

### LINKED ISSUE #1448: Flask>=3 requires changes to flask-jwt-extended usage
Upgrading to `Flask>=3,<4` comes with a second order consequence: currently we use `flask-jwt-extended<4`, which requires `Flask<2.0,>=1.0`. However, upgrading `flask-jwt-extended` to be compatible with `Flask>=3,<4` comes with an api change documented here: https://flask-jwt-extended.readthedocs.io/en/stable/v4_upgrade_guide.html#api-changes

In short, we need to replace all instances of the decorator `@jwt_optional` with `@jwt_required(optional=True)` wherever we update to `Flask>=3,<4`.

Related: https://github.com/gyorilab/gilda/issues/137

I'll try to list all places in downstream repositories in sorger/gyorilab where we would need to make the change:

- [x] [emmaa/emmaa_service/api.py](https://github.com/gyorilab/emmaa/blob/master/emmaa_service/api.py) - 9 instances
- [x] [indra_cogex/apps/curator/curator_blueprint.py](https://github.com/gyorilab/indra_cogex/blob/main/src/indra_cogex/apps/curator/curator_blueprint.py) - 16 instances
- [x] [indra_cogex/apps/data_display/__init__.py](https://github.com/gyorilab/indra_cogex/blob/main/src/indra_cogex/apps/data_display/__init__.py) - 3 instances
- [x] [ui_util/indralab_auth_tools/indralab_auth_tools/auth.py](https://github.com/gyorilab/ui_util/blob/master/indralab_auth_tools/indralab_auth_tools/auth.py) - 1 instance
- [x] [indra_db/indra_db_service/config.py](https://github.com/gyorilab/indra_db/blob/master/indra_db_service/config.py#L42) - 1 instance

## To test the upgrade

- Use e.g. CoGEx with the updates provided in https://github.com/gyorilab/indra_cogex/pull/172 as a test case
- Build the CoGEx docker _locally_ and ensure the docker builds from the branch for CoGEx and for `indralab_auth_tools` to provide the full updates. Make sure the build is tagged with something else than `latest` to not overwrite the current running service.
- Run the docker _locally_ and check that login and curation capabilities work as expected (although actually logging in probably won't work as login typically requires `https`) as well as looking for any other issues that come from updates.
- Repeat the build and deploy steps for the current service and check again that everything works (especially the actual login functionality).
- Repeat the whole process for emmaa (which runs on a separate server) and indra_db (which uses lambda + api gateway)

## Testing

- [x] Test with CoGEx - see https://github.com/gyorilab/indra_cogex/pull/172
- [x] Test with EMMAA - see https://github.com/gyorilab/emmaa/pull/280
- [x] Test with new indra db deployment - https://github.com/gyorilab/indra_db/pull/222

kkaris:
This PR updates the extras requirements in setup.py to resolve #1448. 

Resolves #1448.

**Todo**

- [x] ~Revisit the cython and pyjnius upgrade: It was necessary to get Python 3.10 working in https://github.com/gyorilab/indra_cogex/pull/172, but since downgrading to python 3.8 (via use of ubuntu 20 as base image) this might not be necessary any longer. Leaving the cython and pyjnius versions untouched would be the more conservative approach (_don't fix what ain't broken..._)~ _Fixed in https://github.com/sorgerlab/indra/pull/1452/commits/90eba2c756704355a1408fb50a6141e975ebdcf7_

_Update 7/8 2024_
Code is tested with https://github.com/gyorilab/indra_cogex/pull/172 and https://github.com/gyorilab/ui_util/pull/20, but not yet with https://github.com/gyorilab/emmaa/pull/280 or a new deployment of db.indra.bio on lambda+api gateway.

_Update 7/11 2024_
Code has been tested and is used in production for EMMAA using Python 3.9.

_Update 7/18 2024_
Code has been tested and is used in the dev deployment for INDRA DB.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,8 +36,7 @@ def main():
                       'plot': ['matplotlib'],
                       'isi': ['nltk<3.6', 'unidecode'],
                       'api': ['flask>=3.0,4.0', 'flask_restx<0.4', 'flask_cors',
-                              'docstring-parser', 'gunicorn',
-                              'markupsafe<2.1.0'],
+                              'docstring-parser', 'gunicorn'],
                       # scikit-learn 1.5.0 breaks DisambManager.run_adeft_disambiguation
                       # see: https://github.com/gyorilab/adeft/issues/80
                       'sklearn_belief': ['scikit-learn<1.5.0'],
--- a/setup.py
+++ b/setup.py
@@ -35,8 +35,11 @@ def main():
                       'graph': ['pygraphviz'],
                       'plot': ['matplotlib'],
                       'isi': ['nltk<3.6', 'unidecode'],
-                      'api': ['flask>=3.0,<4.0', 'flask_restx<0.4', 'flask_cors',
-                              'docstring-parser', 'gunicorn'],
+                      'api': ['flask>=3.0,<4.0',
+                              'flask_restx>=1.3.0',
+                              'flask_cors',
+                              'docstring-parser',
+                              'gunicorn'],
                       # scikit-learn 1.5.0 breaks DisambManager.run_adeft_disambiguation
                       # see: https://github.com/gyorilab/adeft/issues/80
                       'sklearn_belief': ['scikit-learn<1.5.0'],
--- a/setup.py
+++ b/setup.py
@@ -18,8 +18,8 @@ def main():
     extras_require = {
                       # Inputs and outputs
                       'trips_offline': ['pykqml'],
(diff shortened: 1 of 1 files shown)


kkaris:
The only two tests failing are for the `indra_db_rest` (timeout errors)

PR #1476, opened 2025-07-29
Pubmed client updates

kkaris:
This PR updates the pubmed client as well as adds tests for the new functions introduced and functions that were previously uncovered or badly covered.

### New functions
- `get_full_xml_by_pmids`: this functions uses edirect's `efetch` utility to bulk download an XML containing multiple `<PubmedArticle>` tags.
- `get_nct_ids_for_pmids`: Gets the NCT IDs for a list of PubMed IDs (using edirect's efetch).
- `get_nct_ids_for_pmid`: Get the NCT IDs for a given PubMed ID (using the regular rest API).
- `get_nct_ids_from_full_xml`: Gets the NCT IDs for a given PubMed ID from a full XML containing multiple `<PubmedArticle>` XML tags.
- `get_nct_ids_from_article_xml`: Extracts NCT IDs from a `<PubmedArticle>` XML tag.

### Updated functions
- `get_article_xml` is refactored in fully backwards compatible way so that part of the code is its own function.
- `_get_article_info` has been updated to also get publication types.
- `get_metadata_from_xml_tree` is refactored in a backwards compatible way so that the code extracting data from the `<PubmedArticle>` tag is isolated.

### Other updates
- The minerva client's `resource_url` and api's `base_url` were changed to http from https, as the https URLs did not resolve any longer, resulting in the tests for the `sources.minerva` module to fail.
- Skip acsn module tests that rely on loading data from web, as https://acsn.curie.fr/ is down.

### DIFF
--- a/indra/tests/test_acsn.py
+++ b/indra/tests/test_acsn.py
@@ -1,17 +1,36 @@
 import pandas
 import requests
+from unittest import skip
+from functools import lru_cache
 from indra.sources.acsn import api
 from indra.ontology.standardize import get_standard_agent
 from indra.sources.acsn.processor import get_stmt_type, AcsnProcessor
 
 
-relations_df = pandas.read_csv(api.ACSN_RELATIONS_URL, sep='\t')
-gmt_file = requests.get(api.ACSN_CORRESPONDENCE_URL).text.split('\n')
-correspondence_dict = api._transform_gmt(gmt_file)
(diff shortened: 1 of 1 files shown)


kkaris:
The tests in `test_acsn.py` are failing because https://acsn.curie.fr/ doesn't seem to exist any more.

bgyori (approved):

PR #1479, opened 2025-09-10
Search page template update and depmap analysis fix

haohangyan:
This PR adds a second right bar to show the summarization of the page and the clickable example samples. Changed the layout and the way of how 'main', 'right bar', 'second right bar' and 'footer' interact

This PR also includes some changes regarding the depmap analysis where we pass in the source count dictionary as we calculate the matrix rows from statements


indra/assemblers/indranet/assembler.py
```diff
@@ -592,6 +592,13 @@ def statement_to_rows(stmt, exclude_stmts=None, complex_members=3,
             pos = None
         # Create a simple flat list of just the values instead
         # of a dict with keys
+        statemet_hash = stmt.get_hash(refresh=True)
```
bgyori, line 595: Typo: statemet -> statement

kkaris:
When I have an extra wide window (can also be triggered by zooming out), the summary side bar disappears. Fix that and then this PR should be good to go in combination with https://github.com/gyorilab/indralab-vue/pull/45.

Issue #1480, opened 2025-10-15
Update chebi client's 'entry from web' client


kkaris:
The URL used in `get_chebi_entry_from_web` in the `chebi_client`, `http://www.ebi.ac.uk/webservices/chebi/2.0/test/` seems to be down (deduced from running `test_chebi_name_from_web` in `test_chebi_client.py`).

If this is permanent, a potential replacement is [the ChEBI REST API's](https://www.ebi.ac.uk/chebi/backend/api/docs/) endpoint for getting compound information (at `/chebi/backend/api/public/compound/{chebi_id}/`).

PR #1481, opened 2025-10-15
Expand sqlite ontology

kkaris:
This PR allows setting the importable `bio_ontology` to the SqliteOntology, in addition to the already existing BioOntology and VirtualOntology, by setting `INDRA_ONTOLOGY_URL` to `'sqlite'`.

A bug in the SqliteOntology is also fixed where because the `initialize` method was not defined, the parent class method would run and error.

A test in test_pmc_client.py that tests for invalid PMC IDs was fixed. The previous PMC ID passed in now actually points to an existing record. A new invalid PMC ID was picked.


kkaris:
The two tests consistently failing are related to what I posted about in #1480, with the chebi client's 'entry from web' api being broken.

PR #1486, opened 2025-11-24
Fix timeout test

kkaris:
This PR updates the tests for `indra.sources.indra_db_rest`:
- new timeouts for tests to accommodate for when the database is not warmed up and takes longer to respond
- Replace a hash in `test_get_statements_by_hash` for a statement which no longer exists


kkaris:
Unrelated tests that only showed up for the PR failed. I'll look into those tomorrow.

PR #1491, opened 2026-03-10
Handle PubmedData being None in pubmed client

kkaris:
This PR handles a corner case when extracted `PubmedData` values are `None`.

Other updates:
- Move text to docstring for `get_full_xml_by_pmids` from code comments
- Clearer error from bad parsing in `get_full_xml_by_pmids`
- Add missing return section in docstring for `get_metadata_from_pubmed_article`


kkaris:
Code added via #1494

PR #1507, opened 2026-06-23
Pubmed xml download updates

kkaris:
This PR updates the pubmed client with the following:
- The XML downloading process has been updated to allow more flexibility so that downstream usage can adopt the pubmed client's download process instead of implementing its own
- After testing parallel download, it was determined that a speedup of _at least_ 2-4x was possible with a modest amount of 4 workers, which greatly reduces the time it takes to download the full XML dump from scratch. Because of this, the possibility to run downloads in parallel is added, with a default of serial download (i.e. `max_workers=1`).

### DIFF
--- a/indra/literature/__init__.py
+++ b/indra/literature/__init__.py
@@ -5,7 +5,6 @@ from indra.literature import pubmed_client
 from indra.literature import pmc_client
 from indra.literature import crossref_client
 from indra.literature import elsevier_client
-from functools import lru_cache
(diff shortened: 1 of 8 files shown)


kkaris:
I'll look into the failing tests

kkaris:
All the failing tests are related to the pmc_client getting 429 errors, which seems to be related to the updated/new PMC API. Fixing that might be out of scope for this PR.

_Update Jun 25:_
See #1509 for proposed fixes to this.

